PR #11, opened 2018-06-13
Hover MarkedString

pedrorgirardi:
Hi @snoe,

I should have asked you before creating this PR, but while taking a look at Clojure LSP I made this small change to use `MarkedString` for the `Hover`. 

I'm impressed with the Clojure LSP capabilities, I don't think any other tool has such good support for ClojureScript except Cursive. Amazing job!

*It's great that you have all those specs, it really helps.

### DIFF
--- a/src/clojure_lsp/handlers.clj
+++ b/src/clojure_lsp/handlers.clj
@@ -295,12 +295,12 @@
         cursor-sym (:sym (find-reference-under-cursor line column local-env))]
     (log/warn "definition" doc-id line column cursor-sym)
     (first
-      (for [[env-doc-id {:keys [usages]}] file-envs
-            {:keys [sym tags] :as usage} usages
-            :when (= sym cursor-sym)
-            :when (and (or (= doc-id env-doc-id) (:public tags))
-                       (:declare tags))]
-        {:uri env-doc-id :range (->range usage)}))))
+     (for [[env-doc-id {:keys [usages]}] file-envs
+           {:keys [sym tags] :as usage} usages
+           :when (= sym cursor-sym)
+           :when (and (or (= doc-id env-doc-id) (:public tags))
+                      (:declare tags))]
+       {:uri env-doc-id :range (->range usage)}))))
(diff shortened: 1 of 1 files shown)


pedrorgirardi:
...how do you generate the binary? `lein uberjar`? If I install the standalone JAR as `clojure-lsp` I get this:

screen shot 2018-06-13 at 8 12 30 am

pedrorgirardi:
Thank you for taking the time to comment on the changes and pointing me out the way to go. 

I will try to make the changes tomorrow morning.

pedrorgirardi:
I don't understand this:
```clojure
(#'handlers/refactor (path->uri doc-id)
                       (inc (int line))
                       (inc (int col))
                       (.getCommand params)
                       args)

snoe:
At the moment, the `#'vars` aren't working as I expect which is unfortunate. So I'll take this as a chance to [rubberduck](https://en.wikipedia.org/wiki/Rubber_duck_debugging).

Usually, when connecting to a live repl, if you re-eval a `(defn refactor [...])` any long lived process using that function will not pick up your new changes. In this case, `(.startListening launcher)`.

The idea behind using a var is that it should always point to the last eval'd version and not hold a reference to that initial version.

If this was working, I'd be able to re-eval `(defn refactor [,,,])` and the next time I invoked that endpoint I'd get the changes.  

While developing I'm using the rrun plugin, but when I eval `@db/db` I get an empty map which suggests that I'm not in the same session as the running server. So maybe the vars are working, but the server is running in it's own process. I should look at invoking `clojure.tools.nrepl.server` directly in main instead of this old rrun plugin that must not be doing what I expected.

pedrorgirardi:
I just pushed some changes. Please let me know if that's what you were talking about :)

*I used the VS Code extension and it looks like the formatter changed a little bit the code.

pedrorgirardi:
Commits squashed.

pedrorgirardi:
Thank you!

PR #15, opened 2018-12-10
Boot support

patrickpichler:
I hope the way i implemented it is OK.

I'm relatively new to Clojure and would be glad for each comment on how to improve the code.

### DIFF
--- a/src/clojure_lsp/handlers.clj
+++ b/src/clojure_lsp/handlers.clj
@@ -154,7 +154,7 @@
           (string/split sep)))
     (catch Exception e
       (log/warn "Error while looking up classpath info in" root-path (.getMessage e))
-      '[])))
+      [])))
 
 (defn lein-lookup-classpath [root-path]
   (lookup-classpath #(shell/sh "lein" "classpath" :dir root-path) root-path))
--- a/src/clojure_lsp/handlers.clj
+++ b/src/clojure_lsp/handlers.clj
@@ -136,57 +136,45 @@
     (async/pipeline-blocking 5 output-chan xf (async/to-chan dirs) true (fn [e] (log/warn e "hello")))
     (async/<!! (async/into {} output-chan))))
 
-(defprotocol Project
-  "Protocol defining common methods for projectc"
-  (calculate-hash [project] "calulate a hash for the project definition")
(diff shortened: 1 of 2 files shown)


patrickpichler:
@snoe can i get any feedback on this pull request?

patrickpichler:
Thanks for your feedback, I will try to encoorparate that :+1:

patrickpichler:
@snoe I now removed the record and replaced it with a map with two keys.

src/clojure_lsp/handlers.clj
```diff
@@ -136,29 +136,58 @@
...
+  (lookup-classpath #(shell/sh "boot" "show" "--fake-classpath" :dir root-path) root-path))
+
+(defn build-lein-project [root-path]
+  (let [file-hash (hash-project-file (io/file root-path "project.clj"))
+        classpath (lein-lookup-classpath root-path)]
+    {:project-hash file-hash :classpath classpath}))
+
+(defn build-boot-project [root-path]
+  (let [file-hash (hash-project-file (io/file root-path "build.boot"))
+        classpath (lein-lookup-classpath root-path)]
```
snoe: Should be `boot-lookup-classpath`

src/clojure_lsp/handlers.clj
```diff
@@ -136,29 +136,58 @@
...
-      (-> (shell/sh "lein" "classpath" :dir root-path)
+    (let [sep (re-pattern (System/getProperty "path.separator"))]
+      (-> (command)
           (:out)
           (string/trim-newline)
           (string/split sep)))
     (catch Exception e
-      (log/warn "Could not run lein in" project-root (.getMessage e)))))
+      (log/warn "Error while looking up classpath info in" root-path (.getMessage e))
+      '[])))
```
snoe: Vector doesn't need quoting

src/clojure_lsp/handlers.clj
patrickpichler: Thanks, that was an obvious one. I think the issue was that my tests solely consisted of seeing if I get some autocompletion for a boot project (which isn't enought, as this is done by nrepl I think).

patrickpichler:
@snoe I changed the requested code parts :+1:

PR #18, opened 2019-01-19
Fix a part of LanguageClient-neovim example configure

rinx:
I think it should be "add-missing-libspec" command in your example code of configuring LanguageClient-neovim example.

Thanks.


snoe:
Thanks!

PR #22, opened 2019-03-26
update emacs instructions to inhibit lsp indentation

dpsutton:
lsp _can_ indent but the clojure mode does not yet support this. leave
this to clojure-mode which is much better at the moment.


snoe:
Thanks for figuring this out.

Issue #1, opened 2018-04-16
Submit to Homebrew


snoe:
It would be helpful to get a formula in homebrew

I guess this is the starting point: https://docs.brew.sh/Formula-Cookbook

markx:
Homebrew has a "test" step to verify the installation. How do you think this can be tested?

snoe:
Well, the binary responds to JSON RPC on stdin so if the have a framework around that, sending a supported message on stdin should suffice https://microsoft.github.io/language-server-protocol/specification

markx:
Since the binary blocks on reading stdin and doesn't exit, I think it will have to run in a process, and then we start another process to check the output and stop the binary. This sounds a little too complex to me.

It'll be trivial to test if it can take a flag , like `--help` or `--version`, and output some fixed string. I think this is the case for many interactive programs on homebrew.

snoe:
That makes sense to me.

markx:
I did a little search, and found that most, if not all, lang servers are distributed via the language's own package manager, instead of homebrew. In this way it is automatically available on all platform it supports. Is it possible to do this via leiningen, or something else?

snoe:
Good question.

There's a few exceptions in the list here https://langserver.org#implementations-server I think the main difference is that so many languages now have blessed package managers that can install binaries. 

Clojure now has `clj` maybe we can specify the github repo on the cli and a main to run...

With lein, we'd have to make this into a lein plugin and tell people to add it to their profiles, or do what cider does and inject the plugin from cli... it's doable but will probably get hairy and it excludes people not using lein.

We could probably wrap this in a package.json and distribute with npm which, the world being as it is, would probably be available for most people. 

Some language tools (like lein and npm) do the `curl | sh` approach is an approach that gets people running but has security concerns.

snoe:
I want to enumerate the the problems I'm trying to solve with homebrew:

1) A standard, stable way of launching the server for clients, many of whom will have their own releases. So the command should not change, should not require hard coded paths, and be as close to cross platform as possible. This means we need some executable on the path that can launch the server. (e.g. `clojure-lsp`, `java -m clojure.lsp.main`, etc).

2) Installable - ideally it's one step to get clojure-lsp onto their system and ready to run. No `chmod` step, or downloading and moving to a $PATH.

3) Upgradeable - there should be a straightforward way to get the latest version

[...]

PR #26, opened 2019-04-07
print version when run with args

markx:
Prepare for #1.


snoe:
Thanks, opened this https://github.com/snoe/clojure-lsp/issues/29 so we don't always see `0.1.0-SNAPSHOT`

PR #32, opened 2019-04-13
Apply defaults when initializationOptions are null

ProjectFrank:
Previously when a client sent null for the initializationOptions param,
the default client-settings would not be used and client-settings itself
would be nil. This change makes it so that the default client-settings
are used when initializationOptions is null.


snoe:
Thanks!

PR #39, opened 2019-04-23
Indent introduce-let body

ProjectFrank:
Make it so that the body of the let is indented two spaces instead of one.

Before:

```clojure
(+ 1 1)

(let [two (+ 1 1)]
 two)
```

After:
```clojure
(+ 1 1)

(let [two (+ 1 1)]
  two)
```


ProjectFrank:
It would be nicer to run the result of an edit like this one through cljfmt. I didn't do that because that's not how things are being done elsewhere, but please let me know if that change would be preferred.

snoe:
Running refactors through cljfmt would be preferred but this is a nice change too.

Issue #27, opened 2019-04-08
Better jump to definition in jarfiles.


snoe:
Currently when trying to get the definition of something defined in a jar file we return a pseudo uri for zipfiles that vim understands but no one else does.

`zipfile:///Users/case/.m2/repository/org/clojure/clojure/1.9.0/clojure-1.9.0.jar::clojure/core.clj`

Let's figure out what servers do for java and see if we can use that instead. (They might use a `jdt` protocol uri). Most clients I've seen have special handling for something in the java lsp server which I think is related to this problem.

https://github.com/neoclide/coc.nvim/issues/652

snoe:
Should be fixed with 
24ebfa6f591e1f8bb80aac191c6adb635497bf77  and
7bc1fb3d1b4a5535362420fc67bc4a90f5731d01 which has an example of a client handling the IANA `jar` scheme.

benedekfazekas:
trying to make this work with emacs.

I put a `.lsp/config.edn` in the project like:
```
{"dependency-scheme" "jar"}
```

still getting back an uri starting with `zipfile:///...`

what am i doing wrong?

snoe:
ah, rm the `.lsp/sqlite.1.db` in your project

benedekfazekas:
yup, the right uri comes back now. cheers

benedekfazekas:
still on the emacs side of things.

I figured I need an `uri-handler` to make jumping into source work for dependencies (jarred up obviously) I put together something rather hacky and it works :).

if I ever clean that up this should go into https://github.com/emacs-lsp/lsp-mode probably in a separate lsp-clojure.el file. Does this sound right @snoe ?

this starting to feel amazing btw :) thanks for the whole thing!!

snoe:
Yeah, that sounds about right, it'd be great to get it distributed with lsp-mode

PR #40, opened 2019-04-26
Update emacs section in the README

benedekfazekas:
to reflect the recently added (emacs-lsp/lsp-mode#791) built-in support for `clojure-lsp` in `emacs-lsp` as discussed in #27. Mention the need for the right `dependency-scheme` config, and also a defcustom in `lsp-mode` to override the command to start the `clojure-lsp` server.

---

So far I am loving this, thanks again @snoe for your work. I find that this is really valuable when I am looking around in projects on my local disk -- jumping around in deps etc, and find references really shine. For development I still prefer to go REPL driven -- when I tried to just go with `clojure-lsp` when writing code I found myself doing TDD to get feedback and I cringed ;). But again even with cider running the easy find-references/rename is still very valuable.

Refactorings wise I just realised the most of the refactorings are already available in elisp only clojure mode bar `add-missing-libspec`. I might look into make that work in emacs next. @yyoncho also mentioned that `company-lsp` is not working too well with `clojure-lsp` this may need some looking into too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,20 +185,40 @@ Proof of concept in the client-vscode directory in this repo.
 I tried making a client but my hello world attempt didn't seem to work. If someone wants to take this on, I'd be willing to package it here too.
 
 ### emacs
-Using https://github.com/emacs-lsp the following works for registering clojure-lsp:
+[lsp-mode](https://github.com/emacs-lsp/lsp-mode) has built in support for `clojure-lsp` since `lsp-mode-20190416.1936`. With `use-package` add the following to your emacs config:
 
 ```
-(require 'lsp-mode)
-(lsp-register-client
- (make-lsp-client :new-connection (lsp-stdio-connection '("bash" "-c" "clojure-lsp"))
-                  :major-modes '(clojure-mode clojurec-mode clojurescript-mode)
-                  :server-id 'clojure-lsp))
-(add-to-list 'lsp-language-id-configuration '(clojure-mode . "clojure-mode"))
-(setq lsp-enable-indentation nil)
-(add-hook 'clojure-mode-hook #'lsp)
-(add-hook 'clojurec-mode-hook #'lsp)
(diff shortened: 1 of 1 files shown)


snoe:
Thanks @benedekfazekas I'm taking a huge amount of inspiration from refactor-nrepl. I agree, though the repl is still very much needed for development but I like to think the static approach for code manipulation can be better in many ways (and less good in some).

In the end though, when people ask "what editor is best with clojure" I hope we can answer "it doesn't matter just install clojure-lsp".

I know completion could use some work, but I'm not really sure where to go with it. It gives pretty bad results with vim's omnicomplete, decent with deoplete, and pretty great with coc's completion. I would certainly appreciate some insight into why it behaves well in some but not others.

benedekfazekas:
maybe a little known fact but there are refactorings available in clojure-mode already, naturally for both clojure and clojurescript as they are implemented in elisp. See docs here: https://github.com/clojure-emacs/clojure-mode#refactoring-support

yyoncho:
@ProjectFrank are you willing to contribute https://github.com/ProjectFrank/lsp-clojure/blob/master/lsp-clojure.el#L8 to lsp-mode? 

@snoe is it possible to pass dependency-scheme in initialization options? I tried 

```
  "initializationOptions": {
    "dependency-scheme": [
      "jar"
    ]
  }
```

but it seems like it does not work?

ProjectFrank:
@yyoncho sure thing

benedekfazekas:
nice one, can't wait to try `add-missing-libspec`

snoe:
@yyoncho `"dependency-scheme": "jar"` should work (no vector), but you might have to drop the `<projectroot>/.lsp/sqlite.1.db` file as well

yyoncho:
@snoe thanks, it works. I have tried it that way to but it didn't work I guess due to `.lsp/sqlite.1.db`.

PR #59, opened 2019-08-18
[vscode] Satisfy vsce CLI requirements to be able to package the extension

roman01la:
When attempting to package the vscode extension with vsce CLI, it requires to remove `vscode` from `dependencies` and edit default README file. Are you going to publish the extension to the store btw?

### DIFF
--- a/client-vscode/README.md
+++ b/client-vscode/README.md
@@ -1,65 +1,11 @@
-# client-vscode README
-
-This is the README for your extension "client-vscode". After writing up a brief description, we recommend including the following sections.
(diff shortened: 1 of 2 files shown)


snoe:
I'm not opposed to putting it into the store but I'd actually prefer if someone were to "fork" the extension into a standalone repo and take ownership of it.

Mainly because I don't use it and it really is a poc in my mind.

roman01la:
Ok, I can do that. I'll submit a PR updating readme with a link to the repo with extension

PR #60, opened 2019-08-18
Include all cljs specific core syms

roman01la:
This PR adds cljs.core specific symbols


snoe:
Thanks! For curiosity's sake, which symbols were you using?

roman01la:
@snoe I'm not sure I understand your question

PR #65, opened 2019-08-22
Provide more concrete messages for different types of unused declaration

roman01la:
This PR adds concrete messages for unused-ns and unused-param diagnostics


snoe:
Thanks!

PR #69, opened 2019-09-28
Use severity 2 instead of 1 for unused references and declarations

tolgraven:
As an aside, does unused declaration actually make sense for non-private defns?


snoe:
Thanks!

I guess it depends if you're writing a library vs an app. I'd be happy to accept a pr that puts the non-private unused decls behind a clientsettings flag.

PR #75, opened 2019-10-13
Add installer script

rnadler:
Figuring out if I needed to update and doing the manual installation became tedious, so I wrote a quick bash installer. This would probably be useful for others.

Having `clojure-lsp` correctly report its version would be nice. Using the executable file size isn't great, but it works.  

Please modify/improve/fix as needed.

Cheers!

### DIFF
--- a/install_clojure_lsp
+++ b/install_clojure_lsp
@@ -4,16 +4,23 @@
 # Keep a local copy of old versions
 # Requires curl
 
-if [ "$(id -u)" != "0" ]; then
-  echo "This script must be run as root" 1>&2
-  exit 1
-fi
-
 # Customize 
 REPO=snoe/clojure-lsp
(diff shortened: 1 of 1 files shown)


install_clojure_lsp
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+
+# Install the latest version of an executable file from a Github release
+# Keep a local copy of old versions
+# Requires curl
+
+if [ "$(id -u)" != "0" ]; then
+  echo "This script must be run as root" 1>&2
+  exit 1
```
snoe: I would prefer not having a script that must be run as root, if the `DST_DIR` could be specified and/or defaulted to `$HOME/bin` I think the rest of the script could be run without root. 

What are your thoughts? Is that too paranoid or weird for install scripts?

PR #80, opened 2019-11-04
Fixes instructions for Emacs users

agzam:
Otherwise it throws:
```
Warning (lsp-mode): Unable to calculate the languageId for current buffer. Take a look at lsp-language-id-configuration.
```


snoe:
Thank you

PR #91, opened 2019-12-16
Update lsp4j dependency to 0.8.1

patrickpichler:
In order to implement features such as `documentSymbol` the lsp4j dependency version needs to be increased.


snoe:
Thanks!

PR #95, opened 2019-12-28
Add a diagnostic for calling a function with the wrong number of arguments

### LINKED ISSUE #62: `Wrong number of arguments` diagnostic
This is an essential diagnostic to have I think 🙏

Luminiscental:
Adds a diagnostic to check against supported arities of a function when it is called in references, fixes #62.

### DIFF
--- a/src/clojure_lsp/crawler.clj
+++ b/src/clojure_lsp/crawler.clj
@@ -86,7 +86,7 @@
           :when (when-let [relevant-function (last relevant-functions)]
                   (:sym relevant-function)
                   (try
-                    (not-any? #(supports-argc % argc) (:signatures relevant-function))
+                    (not-any? #(supports-argc % argc) (get-in relevant-function [:signatures :sexprs]))
                     (catch Exception e
                       false)))]
       {:range (shared/->range call-site)
--- a/src/clojure_lsp/crawler.clj
+++ b/src/clojure_lsp/crawler.clj
@@ -67,6 +67,12 @@
        :message (str "Unknown forward declaration: " (:str usage))
        :severity 1})))
 
+(def ignore-arity
+  #{'clojure.core/defn     ; regex-like arglists
+    'clojure.core/defmacro ; regex-like arglists
+    'clojure.core/format   ; return type annotation breaks clojure.core arglists
+    })
+
 (defn ^:private supports-argc [signature argc]
   (let [min-argc (count (take-while #(not= '& %) signature))
         has-rest (not= min-argc (count signature))]
@@ -84,7 +90,14 @@
                 relevant-functions (filter #(= (:sym %) function-sym) function-references)
                 relevant-function (last relevant-functions)
                 overloads (get-in relevant-function [:signatures :sexprs])]
-          :when (and overloads (not-any? #(supports-argc % argc) overloads))]
+          :when (and
+                  overloads
+                  (not (ignore-arity function-sym))
+                  (try
+                    (not-any? #(supports-argc % argc) overloads)
+                    (catch Exception e
+                      (log/warn "Couldn't interpret signature for" function-sym ":" (.getMessage e))
+                      false)))]
       {:range (shared/->range call-site)
        :code :wrong-arity
        :message (let [plural (not= argc 1)]
--- a/src/clojure_lsp/crawler.clj
+++ b/src/clojure_lsp/crawler.clj
@@ -69,23 +69,19 @@
 
 (defn ^:private diagnose-wrong-arity [uri usages]
   (let [all-envs (assoc (:file-envs @db/db) uri usages)
-        declarations (->> usages
-                          (filter #(and (contains? (:tags %) :declare)
-                                        (= (:kind %) :function))))
-        declared-references (remove (comp #(contains? % :alias) :tags) declarations)
-        references (->> all-envs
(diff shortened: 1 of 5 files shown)


Luminiscental:
Adding a note that this doesn't handle threading macros properly, I will fix this later.

src/clojure_lsp/crawler.clj
```diff
@@ -67,6 +67,32 @@
...
+                relevant-functions (filter #(->> % (:sym) (= function-sym)) function-references)]
+          :when (when-let [relevant-function (last relevant-functions)]
+                  (not-any? #(supports-argc % argc) (:signatures relevant-function)))]
+      {:range (shared/->range call-site)
+       :code :wrong-arity
+       :message (let [plural (not= argc 1)]
+                  (format "No overload %s for %d argument%s"
+                          (:str call-site) argc (if plural "s" "")))
+       :severity 1})))
+
```
snoe: I loaded up your branch and I was curious why clojure.core fields weren't being checked for arity.

It's because `usages` and `all-envs` are mixed up, you want the `call-sites` from `usages` (the current doc) and `function-references` from `all-envs`.  Then I found that many of the `clojure.core` signatures are returning as `(quote ([a] [a b]))`, or have type hints like `'(^clojure.lang.ISeq [coll])` which explode in `supports-argc`. So I added a `catch` to ignore functions it can't handle. 

Here's where I am but there's still some functions erroneously being flagged. For instance, in this project `zsub/subzip` takes 1 arg but it's diagnosed as wrong arity.



```
(defn ^:private diagnose-wrong-arity [uri usages]
  (let [all-envs (assoc (:file-envs @db/db) uri usages)
        function-references (filter :signatures (mapcat val all-envs))
        call-sites (->> usages
                        (filter :argc))]
    (for [call-site call-sites
          :let [argc (:argc call-site)
                function-sym (:sym call-site)
                relevant-functions (filter #(->> % (:sym) (= function-sym)) function-references)]
          :when (when-let [relevant-function (last relevant-functions)]
                  (:sym relevant-function)
                  (try
                    (not-any? #(supports-argc % argc) (:signatures relevant-function))
                    (catch Exception e
                      false)))]
      {:range (shared/->range call-site)
       :code :wrong-arity
       :message (let [plural (not= argc 1)]
                  (format "No overload %s for %d argument%s"
                          (:str call-site) argc (if plural "s" "")))
       :severity 1})))
```

Luminiscental: Thanks for the review!

I've applied this usages/all-envs fix, and also I reworked the :signatures storage to remember both the arglists and the eval'd signature, although this broke one of the parser tests somehow.

I will try and look at fixing the broken test tomorrow.

Luminiscental: Okay most of the issues with annotations and quoted arglists are fixed now, however some macros like `defn` have `:arglists` written almost like a regex, which `supports-argc` just interprets literally for now, which is broken.

I'm not sure if the best approach is to have an ignore list for functions like these or to auto-detect if an `:arglists` looks like a regex or something else

snoe: An ignore list would be fine - I don't think clojure.core itself is super consistent about those annotations. 

Another possibility would be a map that allowed overriding the signatures or argcs for a function.

snoe: > I wasn’t able to reproduce the `subzip` misdiagnosis

It's possible that was a repl state issue, it was in parser.clj. But I'll give it another look once these changes are ready.

Luminiscental: Okay I've added a short ignore list. Checking over parser.clj I saw an issue with nested threading but that should be fixed now, and I've merged from master so things should be ready now.

Luminiscental:
One small note: the updated `is-params` in `parser.clj` fixes an issue where return type annotations threw off the parser when looking for function signatures, so e.g. `clojure.core/format` shows signature `String/format` in the main branch whereas with this change it shows `^String [fmt & args]` as expected.

test/clojure_lsp/handlers_test.clj
```diff
@@ -55,6 +55,81 @@
...
+                "No overload foo for 4 arguments"]
+               (map :message usages)))))
+    (testing "for threading macros"
+      (reset! db/db {:file-envs {"file://a.clj" (parser/find-usages "(defn foo ([x] x) ([x y z] (z x y)))
+                                                                     (defn bar [] :bar)
+                                                                     (defn baz [arg & rest] (apply arg rest))
+                                                                     (->> :test
+                                                                       (foo)
+                                                                       (foo 1)
+                                                                       (bar))
```
snoe: There's an issue with thread macros when you don't wrap single arity functions in parens. If you change this line to just `bar` you'll notice the diagnostic disappears. If you look at the `find-usages` result  you'll see the reference is missing.

I suspect parser test code like `(defn x [y] y) (-> 1 y)` will show you a difference compared `(defn x [y] y) (-> 1 (y))`.

Luminiscental: Okay I think this is fixed now, I added a parser test to check `:argc` values on usages as well as adding a case of this in the handler test.

snoe:
Looks great! It looks like cljs files also need the ignore, I'll make that change right away.

Thanks again, this is a really nice feature.

PR #96, opened 2020-01-03
Add basic document-symbol support

### LINKED ISSUE #90: Implement `documentSymbol` lsp function
It would be a nice to have to list all symbols of a document for easy traversal.

patrickpichler:
This adds basic support for the documentSymbol feature (#90)

Tasks:
- [x] show namespaces
- [x] show functions
- [x] show definitions
- [x] return in hierachical order


patrickpichler:
@snoe I would appreciate if you would look over the specs usage and maybe give me some feedback on it. This was the first time I had to use spec.

snoe:
@patrickpichler This looks great, thanks. 
The namespacing of the enum spec is how I've handled that issue in the past, and something you just have to deal with when using `:req-un`

PR #97, opened 2020-01-04
Add emacs doc on adding lein to path

Looveh:
If the path to project managers aren't added to `PATH` the crawler is unable to launch them as sub processes when launched from emacs.


snoe:
Thank you!

PR #98, opened 2020-01-11
Add workspace-symbol support

patrickpichler:
The lsp now supports the workspace symbol request. This allows to query
the whole workspace for a certain symbol and jump to it.

### DIFF
--- a/src/clojure_lsp/handlers.clj
+++ b/src/clojure_lsp/handlers.clj
@@ -353,6 +353,28 @@
         sym (:sym cursor)]
     (into [] (comp (filter #(= (:sym %) sym))
                    (map file-env-entry->document-highlight)) local-env)))
+(defn file-env-entry->workspace-symbol [uri [e kind]]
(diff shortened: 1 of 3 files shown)


patrickpichler:
The implementation might be a bit slow and naive. It might make sense to keep an index of all symbols cached somewhere.

snoe:
I'm not too concerned with the speed, as long as clients aren't constantly hitting it. 

Do other implementations query with `starts-with`? I found it a bit surprising since it requires writing out the whole namespace. Is it too slow in your opinion if it's uses a substring match? (maybe if it returned nothing until 2 or 3 chars?)

patrickpichler:
Yeah, the starts-with annoyes me now as well after using it for some time. I implemented it that way, because [eclipse JDT lsp](https://github.com/eclipse/eclipse.jdt.ls) uses it as well.

snoe:
I love it! Thanks a ton.

PR #103, opened 2020-01-16
Add completion item resolve support

Cnly:
This PR adds support for the [completion item resolve request](https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#completionItem_resolve) so clojure-lsp could show some more detailed docs for the items when the user moves along the completion menu (normally displayed in a floating window next to the completion menu).

Currently this is done by filling the `data` field of each of the completion items with its symbol name when the completion menu is generated. Then when the user moves to an item in the menu, the LSP client sends over the original completion item with the `data` field, and clojure-lsp will return the documentation for that symbol.

PS: I'm sill familiarising myself with Clojure so I'd appreciate it if you would point out for me things not done right in the patch. Or if you're not happy with the quality, etc., feel free to close this. :)

### DIFF
--- a/src/clojure_lsp/handlers.clj
+++ b/src/clojure_lsp/handlers.clj
@@ -69,6 +69,24 @@
                               [(:uri text-document) edits])
                             changes))}))
 
+(defn- generate-docs [content-format usage]
+  (let [{:keys [sym signatures doc tags]} usage
+        signatures (some->> signatures
+                            (:strings)
+                            (string/join "\n"))
+        tags (string/join " " tags)]
+    (case content-format
+      "markdown" {:kind "markdown"
+                  :value (cond-> (str "```\n" sym "\n```\n")
+                           signatures (str "```\n" signatures "\n```\n")
+                           (seq doc) (str doc "\n")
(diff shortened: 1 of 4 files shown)


src/clojure_lsp/handlers.clj
```diff
@@ -187,24 +188,54 @@
...
+                             "markdown" {:kind "markdown"
+                                         :value (cond-> (str "```\n" sym "\n```\n")
+                                                  signatures (str "```\n" signatures "\n```\n")
+                                                  (seq doc) (str doc "\n")
+                                                  (seq tags) (str "\n----\n" "lsp: " tags))}
+                             ;; Default to plaintext
+                             (cond-> (str sym "\n")
+                               signatures (str signatures "\n")
+                               (seq doc) (str doc "\n")
+                               (seq tags) (str "\n----\n" "lsp: " tags)))})))
```
snoe: Can this extract a function from `hover` so the code is shared?

Cnly: Sure, it's done.

snoe:
This is really great, thanks for this!

Code looks great too.

Issue #104, opened 2020-01-16
Arity diagnostic does not work on schema.core/defn


ericdallo:
The arity check diagnotic feature is awesome, but it looks like broke `schema.core/defn` arity check :(
Is too hard to add support for that?

![image](https://user-images.githubusercontent.com/7820865/72526953-5ac5cc80-3846-11ea-8f42-3869935766a2.png)

snoe:
fixed by ef37e7e11154d688c96c2a7ad27c9d4a5b584c03

arity checks are ignored with an `:ignore-arity? true` on the macro def.
Happy to take a pr to somehow produce a signature sexpr that can be checked.

PR #106, opened 2020-01-17
Fix arity checking for schema.core/defn

Luminiscental:
Adds handling for `:-` type annotations in signatures, so things like in #104 don't have to be ignored.

As well as testing the arity checking here, I tried to keep the `:ignore-arity?` functionality tested with a client macro def in the test but I'm not sure if it should be tested differently, very welcome to suggestions.

### DIFF
--- a/src/clojure_lsp/parser.clj
+++ b/src/clojure_lsp/parser.clj
@@ -520,7 +520,7 @@
   (cond
     (= 'quote (first arglists))
     (let [sexprs (eval arglists)]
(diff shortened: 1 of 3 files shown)


ericdallo:
Thanks for adding the support!

ericdallo:
Tested manually and it works :+1:

Luminiscental:
Yup that seems smart, I think I've implemented what you described now

snoe:
Perfect, thank you.

PR #107, opened 2020-01-18
Add support for shadow-cljs projects

dpsutton:
As far as I can tell, everything seems to work once a classpath is
established. There are three ways to do so with shadow depending on
tooling:

1. shadow-cljs is installed globally and can be called
2. it is run with npm as `npx shadow-cljs ...`
3. it is run with yarn as `yarn shadow-cljs ...`

So just test the system for which are present and go with that. Yarn
and npx have the ability to have all of their output silenced with
--quiet and --silent so all we get back are classpaths.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ Note: You may also consider configuring project specs via the (optional) `.lsp/c
                    :classpath-cmd ["clj" "-Spath"]}]}
    ```
 Each project-spec will add to the list of dependencies for lsp to crawl:
-  - `project-file` is the required filename used by your build tool (project.clj, build.boot, deps.edn, package.json, etc)
+  - `project-path` is the required filename used by your build tool (project.clj, build.boot, deps.edn, package.json, etc)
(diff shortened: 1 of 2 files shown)


snoe:
I don't think `exectuable-exists?` will work in windows.

This might be a case where we build in the most common or global command, but then ask users to specify their `:project-specs` in `.lsp/config.edn` (with a section in README).

dpsutton:
> I don't think exectuable-exists? will work in windows.
Correct. I thought I saw some usage of other shell stuff that would prevent Windows as well so i thought i was embracing the current support.

Not sure which is the most likely to work. Depends if the user has the project configured with npm or yarn. Perhaps as a first step require shadow-cljs be installed globally which is probably a pretty safe bet at first.

dpsutton:
Updated to rely on the globally installed version. Commit message has the way to configure if need to run under `npm` or `yarn`. Didn't see a great spot in the readme for it but didn't spend a ton of time thinking about it.

snoe:
Thank you!

PR #109, opened 2020-01-21
Fix multi arity macro def

### LINKED ISSUE #108: Multi-arity functions with schema.core/defn
The last problem (I hope) I found with `schema.core/defn` is that it `macro-def` does not support `multi-arity` functions like:

image

Is this a big problem? or is there some way to change the macro-def adding this support?

Thanks!

Luminiscental:
Simple stop-gap fix which makes `schema.core/defn` use `:function-params-and-bodies` and allows `:signature-style :typed` for `:function-params-and-bodies` elements.

Fixes #108


snoe:
Looks great, thank you.

PR #110, opened 2020-01-26
Fix docstring indentation weirdness on hover

Luminiscental:
When docstrings are wrapped onto multiple lines, an indentation is picked up on lines after the first line when they are aligned in the source, which leads to hover output like:

alt text

To fix this I added a `format-docstring` function that detects leading whitespace in the lines after the first, and removes it. With these changes we get:

alt text

It should only ever trim whitespace off the start of lines, so the worst case for very pathological inputs is the indentation looking bad again which I think is a fine trade-off for having the majority of cases look better.

### DIFF
--- a/src/clojure_lsp/handlers.clj
+++ b/src/clojure_lsp/handlers.clj
@@ -70,9 +70,11 @@
                             changes))}))
 
 (defn- drop-whitespace [n s]
-  (let [fully-trimmed (string/triml s)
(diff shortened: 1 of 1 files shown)


Luminiscental:
With a more complicated case further indentation is preserved as desired:

alt text

snoe:
Thank you!

PR #117, opened 2020-03-13
Remove `c` as triggercharacters

caioaao:
Trigger characters are characters that trigger the autocompletion but are not part of the result. With that the result of completing anything starting with `c` had an extra `c` appended to it


snoe:
Thanks!

PR #125, opened 2020-04-06
Improve emacs config in readme

ericdallo:


README.md
```diff
@@ -274,7 +274,9 @@ I tried making a client but my hello world attempt didn't seem to work. If someo
 ```
 (use-package lsp-mode
   :ensure t
-  :commands lsp
+  :hook ((clojure-mode . lsp)
+         (clojurec-mode . lsp)
+         (clojurescript-mode . lsp))
   :config
   ;; add paths to your local installation of project mgmt tools, like lein
   (setenv "PATH" (concat
```
ericdallo, line 282: I don't think this is required for running `clojure-lsp`, is there any other motivation @snoe?

dpsutton: @ericdallo I added this a while ago. My thinking was you wanted lsp to start up in clojure buffers. Also, there's no notion of projects to lsp, so if you wanted to start it in a project, you could do so manually. But then each new clojure file in the project would not have lsp enabled.

I'm happy to be proved wrong and corrected though. Is there a better way?

ericdallo: I see, I think this should change from OS to OS, as my OS (NixOS) does not have bins on `/usr/local/bin` folder, but I agree this is an example of a configuration, so I'm OK with that.

PR #130, opened 2020-05-08
Remove unused aliases on clean-ns

ericdallo:
On clean-ns now we find the unused aliases then on formatting remove it the alias is unused.

Finally, we don't need to start CIDER to clean-ns :tada:


snoe:
This is fantastic thank you!

PR #132, opened 2020-05-10
Bump leiningen to 2.9.3

chenrui333:
relates to https://github.com/Homebrew/homebrew-core/pull/54447

homebrew side is always shipping with the latest `leiningen` build, so good to keep the circle-ci dependency up to date as well.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/clojure:lein-2.7.1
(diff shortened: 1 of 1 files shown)


snoe:
Thank you!

chenrui333:
my pleasure!

snoe:
@chenrui333 fyi, that docker image didn't exist circleci seems to only have a 2.8.0 image.

I switched to clojure/lein-2.9.1 in 92c8190 which is the latest there hopefully that's close enough for homebrew?

PR #144, opened 2020-06-18
Add code actions handler

ericdallo:
Implement LSP code actions support adding 2 code actions:

* Clean namespace
* Add missing namespace - showing only when there is a `unknow-ns` on diagnostics and the server can really add that missing namespace.

We can add more code actions latter :)

### DIFF
--- a/src/clojure_lsp/refactor/transform.clj
+++ b/src/clojure_lsp/refactor/transform.clj
@@ -344,7 +344,7 @@
                                    (or
                                      (set/subset? #{:public :ns} (:tags usage))
                                      (get-in usage [:tags :alias]))))
-                         (mapv (fn [{:keys [sym tags] alias-str :str alias-ns :ns :as usage}]
+                         (mapv (fn [{:keys [sym _tags] alias-str :str alias-ns :ns}]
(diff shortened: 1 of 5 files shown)


ericdallo:
c/c/ @snoe

src/clojure_lsp/refactor/transform.clj
```diff
@@ -359,7 +359,7 @@
         qualified-ns-to-add (cond
                               (= 1 (count posibilities))
                               (-> posibilities first :alias-ns))]
-    (add-known-libspec zloc ns-to-add qualified-ns-to-add)))
+    (add-known-libspec zloc ns-to-add alias->info)))
```
snoe, line 362: This is causing all of the alias usages to be splatted into the require...

ericdallo: Oh, sorry, I accidentally changed this line while creating tests...

ericdallo: Fixed it @snoe!

snoe:
Perfect! Thank you.

PR #147, opened 2020-06-20
Improve hover markdown and arity

ericdallo:
* Use markdown Clojure for color syntax.
* Add arity on the right side of function instead of on another markdown line

Before: 
image

After:
Screenshot from 2020-06-20 17-02-10


snoe:
Thank you!

PR #150, opened 2020-06-25
Lread rewrite cljc test

snoe:
Waiting on this here until rewrite-cljc and cljfmt are up on clojars. Might not need to wait on cljfmt it would just mean rewrite-clj would also be included...


ericdallo:
Closing this as this is quite outdated with master and we'll probably use rewrite-clj v1 branch when released

PR #153, opened 2020-06-28
Fix clean-ns bug

ericdallo:
Fixes https://github.com/snoe/clojure-lsp/issues/152

The problem was that when we remove the node and is the first require, the up node is the require form, not a previous require.

### DIFF
--- a/src/clojure_lsp/refactor/transform.clj
+++ b/src/clojure_lsp/refactor/transform.clj
@@ -271,7 +271,10 @@
   (if (z/vector? node)
     (let [alias-node (-> node z/down z/leftmost)]
       (if (contains? unused-aliases (z/sexpr alias-node))
-        (-> node z/remove z/up)
+        (let [removed-node (-> node z/remove)]
+          (if (z/list? removed-node)
+            (z/down removed-node)
+            (z/up removed-node)))
         node))
     node))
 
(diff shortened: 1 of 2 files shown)


snoe:
@ericdallo This seems to still NPE when there are no other requires:

```clojure
(def another-ns-to-clean
  (str "(ns foo.bar\n"
       " (:require\n"
       "   [foo  :as f] ))\n"
       "(defn func []\n"
       "  (b/some))"))
```

```
ERROR in (clean-ns-test) (zip.clj:67)
Uncaught exception, not in assertion.
expected: nil
  actual: java.lang.NullPointerException: null
 at clojure.zip$node.invokeStatic (zip.clj:67)
    clojure.zip$node.invoke (zip.clj:64)
    rewrite_clj.custom_zipper.core$node.invokeStatic (core.clj:59)
    rewrite_clj.custom_zipper.core$node.invoke (core.clj:59)
    clojure_lsp.refactor.transform$clean_ns.invokeStatic (transform.clj:299)
    clojure_lsp.refactor.transform$clean_ns.invoke (transform.clj:281)
    clojure_lsp.refactor.transform_test$fn__21543$fn__21574.invoke (transform_test.clj:183)
```

ericdallo:
@snoe Fixed, now this happens when only one unused require:
```clojure
(ns foo.bar 
  (:require [foo :as f]))
```
becomes
```clojure
(ns foo.bar)
```

snoe:
Thank you!

Issue #157, opened 2020-07-20
Can't get macro defs to work for simple let binding


socksy:
Hi, in my `.lsp/config.edn` I've put the following:

```clojure
{"macro-defs" {reagent.core/with-let ["bindings" "bound-elements"]}}
```
to try and get reagent's `with-let` to be handled like a normal let. However, while this removes the "undefined" warnings, it seems to completely ignore everything inside the binding, causing lots of unused warnings and no longer recognising any mistakes inside that binding. Am I misunderstanding what `macro-defs` is supposed to do, or have I made a syntax error? (I was trying to follow the documentation in the README)

ericdallo:
@socksy for me it works with:
```clojure
{"macro-defs" {reagent.core/with-let [:bindings :bound-elements]}}
```

![image](https://user-images.githubusercontent.com/7820865/87984368-417d3180-cab0-11ea-85a7-be220615db2a.png)

With unused warning on `b`:
![image](https://user-images.githubusercontent.com/7820865/87984566-8608cd00-cab0-11ea-8d0c-f75aab8a0032.png)

socksy:
@ericdallo Thanks! That works for me too. Perhaps the readme should be updated to reflect that?

ericdallo:
I'll open a PR for that later, thanks!

PR #158, opened 2020-07-20
Update documentation for macro-defs to use keywords

socksy:
I had problems with getting the examples in the docs working (see #157), so I thought I'd update the docs to what worked for me


ericdallo:
Thank you!

PR #161, opened 2020-08-02
formatting: return empty array instead of null when no changes

daaku:
It seems like the expectation is that edits is `TextEdit[]`, that is a non-nil. This change makes it return an empty array if there are no formatting changes.

### DIFF
--- a/src/clojure_lsp/handlers.clj
+++ b/src/clojure_lsp/handlers.clj
@@ -513,7 +513,8 @@
         new-text (cljfmt/reformat-string
                    text
                    (get-in @db/db [:settings "cljfmt"]))]
-    (when-not (= new-text text)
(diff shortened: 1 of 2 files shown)


ericdallo:
Thank you for the contribution @daaku!
In the [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_formatting), returning `null` looks to be acceptable by the spec, but I agree empty is better in semantic.

daaku:
Does look you're right! I was playing around with nvim-lsp and ran into this quickly. It does seem like nvim-lsp needs better `null` handling overall.

ericdallo:
I'll merge this as it seems ok for me as follow the spec and helps nvim-lsp users :)

PR #162, opened 2020-08-02
Replace CircleCI with Github Actions

ericdallo:
* Add a separeted test CI, with this we can require PRs to pass tests before merge.
* migrate CircleCI release to use a release CI.


ericdallo:
@snoe could you take a look?

.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+    name: Tag new version
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Bump version and push tag
+        run: "bash release"
+
+  release:
+    name: Create Release
+    needs: tag
```
snoe, line 19: I think this will run with any tag, can it check that it only runs on `release-*`?

ericdallo: Actually @snoe this workflow will run at every push on master which I think we should release a new tag when this happens don't you think?

When this workflow runs, it will run the `release` script tagging on github, and this `needs` tells to github action to only run this `release` step, after the `tag` step before.

snoe: @ericdallo hrm, oh I missed the `run` above. I guess we can do this, personally I like more control over releases but let's try it and see how it works out.

PR #164, opened 2020-08-05
Integrate clj-kondo for diagnostics

### LINKED ISSUE #66: Reuse existing parsing/linting libraries
So I was spending an evening implementing invalid arity diagnostic and wondering if it would make sense to use existing libraries for linting. My original concern was about double parsing, once by lsp and once by linting library, but clj-kondo https://github.com/borkdude/clj-kondo seem to provide analysis data which could be reused in lsp. The library seems to have lots of linting rules. @snoe what do you think?

Cnly:
This PR uses clj-kondo to enhance the diagnostics; some linters overlapping clojure-lsp's functionalities have been disabled.

Currently, clj-kondo is run for the whole project upon initialisation, using `.clj-kondo/cache` as its cache directory. Then when source files are changed, updated content is fed to clj-kondo.

Closes #66.


ericdallo:
Well done @Cnly! I'll take a look and test it soon

src/clojure_lsp/crawler.clj
```diff
@@ -167,12 +168,56 @@
...
+    :message (str (string/upper-case (str (first message))) (string/join (rest message)))
+    :severity (case level
+                :error 1
+                :warning 2
+                :info 3)
+    :source "clj-kondo"}))
+
+(def kondo-base-args {:cache true
+                      :cache-dir ".clj-kondo/cache"
+                      :config {:linters (into {} (map #(hash-map % {:level :off})
```
ericdallo, line 192: why this level `:off`?

src/clojure_lsp/crawler.clj
```diff
@@ -167,12 +168,56 @@
...
+                                                        :unused-bindings
+                                                        :unresolved-symbol
+                                                        :unresolved-namespace
+                                                        :unused-namespace}))}})
+
+(defn- run-kondo!
+  ([paths]
+   (kondo/run! (merge kondo-base-args {:lint paths})))
+  ([text lang]
+   (with-in-str text (kondo/run! (merge kondo-base-args {:lint ["-"]
```
ericdallo, line 203: what this `["-"]` mean? that we won't lint anything?

src/clojure_lsp/crawler.clj
```diff
@@ -167,12 +168,56 @@
...
+         (map (partial kondo-finding->diagnostic lines)))))
+
+(defn find-diagnostics [project-aliases uri text usages]
+  (let [kondo-chan (async/go (kondo-find-diagnostics uri text))
+        unknown (diagnose-unknown project-aliases usages)
         unused (diagnose-unused uri usages)
         unknown-forwards (diagnose-unknown-forward-declarations usages)
         wrong-arity (diagnose-wrong-arity uri usages)
-        result (concat unknown unused unknown-forwards wrong-arity)]
+        result (concat unknown unused unknown-forwards (async/<!! kondo-chan) wrong-arity)]
```
ericdallo, line 220: I think the idea is to in the future replace the clojure-lsp lint with clj-kondo only, so we can start with a feature flag that toggle kondo lint or clojure-lsp lint, what do you think?
Something like [:use-clj-kondo?](https://github.com/snoe/clojure-lsp/blob/master/src/clojure_lsp/refactor/transform.clj#L298) on the settings would be a good approach

src/clojure_lsp/crawler.clj
Cnly: A bunch of linters that overlap with clojure-lsp's existing functionalities are disabled here.

src/clojure_lsp/crawler.clj
Cnly: We tell clj-kondo to lint stdin by passing that argument. See: https://github.com/borkdude/clj-kondo.lsp/blob/66a54966d8e5b26643e9b39241ba26c1215f5f9d/server/src/clj_kondo/lsp_server/impl/server.clj#L120-L124

src/clojure_lsp/crawler.clj
Cnly: > This is concating clj-kondo lints with clojure-lsp lints right?

Yep.

> I think the idea is to in the future replace the clojure-lsp lint with clj-kondo only, so we can start with a feature flag that toggle kondo lint or clojure-lsp lint, what do you think?

I also have the idea that we can switch over to clj-kondo, but I kind of made this PR as a starting point and gradually we can replace clojure-lsp's linting functionalities with those currently disabled kondo linters, then finally we can consider replacing the lsp's parser with kondo's analysis output. What are your thoughts this route?

ericdallo: I see, so this PR adds kondo to lint things that clojure-lsp doesn't yet, right?
After this PR, we can start replacing each diagnose like `unknown`, `unused`, `wrong-arity` and others with clj-kondo, right?
I'm Ok with that, I think this way we don't need the feature flag that much

Cnly: Yes, that's exactly what I meant :) Hopefully we can collect some feedback as the switch progresses.

ericdallo: Perfect!
Do you know some lint that returns from this new integration?
I could only reproduce at my machine the linters from clojure-lsp

Cnly: Try this bizarre form:

```clojure
(let []
  (do
    {1}))
```

kondo produces three messages on my machine. Also here's a handy list of all linters in kondo: https://github.com/borkdude/clj-kondo/blob/master/src/clj_kondo/impl/config.clj#L13

ericdallo: Nice, it really works nice, in the future maybe append to the message the linter used for that diagnostic like clojure-lsp does (to help the user find which linter is that if he pretends to disable it)

PR #165, opened 2020-08-06
Migrate most linters to use clj-kondo

ericdallo:
With this PR we now use all default linters from clj-kondo letting `clojure-lsp` only take care of `unused references` warnings, and `unknown forward declaration` (not supported by kondo).

Besides the tests, I tested each linter manually to check if we would break some current linter.

### DIFF
--- a/src/clojure_lsp/crawler.clj
+++ b/src/clojure_lsp/crawler.clj
@@ -58,13 +58,13 @@
                        #{:ns} :unused-ns
                        #{:public} :unused-public
                        :unused)]
-          :when (or (not= :unused-ns code)
-                    (not (string/index-of uri "test/")))]
+          :when (and (not= code :unused-param)
+                     (or (not= :unused-ns code)
+                         (not (string/index-of uri "test/"))))]
       {:range (shared/->range usage)
        :code code
        :message (case code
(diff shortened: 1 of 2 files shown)


ericdallo:
@Cnly Do you know if we can in a next PR make `clj-kondo` lint the `unused references` (unused public function and unused namespace)? I think this is the advantage of we scan and cache the whole project when connecting, right?

Cnly:
For unused namespaces I think clj-kondo supports them. But why do we actually want to give a warning about unused functions if they are public?

ericdallo:
@Cnly  I  think it's a nice feature that a LSP could provide, most LSPs provide that, is pretty common I remove a code and see that other function can be removed too, and this is something that we get from for free when achieving references.
If clj-kondo does not support that we can keep using that way, but it would be nice if we replace the most `clojure-lsp` linter warnings with clj-kondo

Cnly:
That makes sense. I just asked on Clojurians and the answer is no :( Maybe we can do this later when we make use of clj-kondo's [analysis data](https://cljdoc.org/d/clj-kondo/clj-kondo/2020.07.29/doc/analysis-data).

ericdallo:
Yeah I think this analysis data is the way to achieve the references feature using kondo

PR #166, opened 2020-08-06
Make config key for clj-kondo consistent with others

Cnly:
Since other config keys are all strings.

### DIFF
--- a/src/clojure_lsp/crawler.clj
+++ b/src/clojure_lsp/crawler.clj
@@ -209,7 +209,7 @@
 
 (defn- kondo-find-diagnostics [uri text]
   (let [file-type (shared/uri->file-type uri)
-        user-config (get-in @db/db [:settings :clj-kondo])
+        user-config (get-in @db/db [:settings "clj-kondo"])
         {:keys [findings]} (run-kondo! text file-type user-config)
         lines (string/split-lines text)]
     (->> findings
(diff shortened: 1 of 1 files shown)


ericdallo:
Do you know why the others are strings and not keywords?
clj-kondo and other integrations like cljfmt use keywords, it would be nice to safe change those to accept keywords too otherwise we will have something like this confusing pattern:
```clojure
{"clj-kondo": {:linters{...}}
 "cljfmt" {:indents {...}}}
```

Cnly:
> Do you know why the others are strings and not keywords?

Well, I don't really have an idea about that. It's possible that they've just been some design decisions @snoe made, but it's already too late to change them without breaking current user configs?

ericdallo:
Yeah, we can see what @snoe think about it, but as a public library, we can give the support for both options and "Deprecate" the string option, and after some time remove/major release remove the string support, WDYT?

ericdallo:
I'll merge this to keep the string pattern, but we should see @snoe opinion later

snoe:
I think the reasoning had to do with clients accepting initializationOptions as json. It probably could have been worked around in other ways, but I don't think we coerce those from strings to keywords.

ericdallo:
I see, I think we can add some function to get the settings that already handles keywords or strings, this would keep the compatibility and add the support for keywords, WDYT?

PR #167, opened 2020-08-06
Provide missing exists?, this-as for Clojurescript

FieryCod:
@snoe It seems that `exists?` symbol is missing for clojurescript which triggers bunch of errors on flycheck.
It's my first time I read the clojure-lsp source code, but it seems that one line change should resolve the issue.


ericdallo:
Thank you!

PR #169, opened 2020-08-15
Add command for retrieve server-info

ericdallo:
* adds a new command: `server-info` which returns the port, project-root and config.

### DIFF
--- a/src/clojure_lsp/crawler.clj
+++ b/src/clojure_lsp/crawler.clj
@@ -296,7 +296,7 @@
                   (str (get-in (run-kondo! source-paths kondo-user-config) [:summary :duration]) "ms"))
         (async/<!! crawler-output-chan)))))
 
-(defn find-project-settings [project-root]
+(defn find-raw-project-settings [project-root]
   (let [config-path (Paths/get ".lsp" (into-array ["config.edn"]))]
     (loop [dir (uri->path project-root)]
       (let [full-config-path (.resolve dir config-path)
@@ -304,10 +304,14 @@
             parent-dir (.getParent dir)]
         (cond
           (.exists file)
-          (edn/read-string {:readers {'re re-pattern}} (slurp file))
+          (slurp file)
 
           parent-dir
           (recur parent-dir)
 
           :else
-          {})))))
+          "{}")))))
+
+(defn find-project-settings [project-root]
+  (->> (find-raw-project-settings project-root)
+       (edn/read-string {:readers {'re re-pattern}})))
--- a/src/clojure_lsp/handlers.clj
+++ b/src/clojure_lsp/handlers.clj
@@ -16,9 +16,9 @@
     [rewrite-clj.node :as n]
     [rewrite-clj.zip :as z])
   (:import
-    [java.net URI URL JarURLConnection]
-    [java.util.jar JarFile]
-    [java.nio.file Paths]))
+   [java.net URL JarURLConnection]
+   [org.eclipse.lsp4j
+    ApplyWorkspaceEditParams]))
 
 (defn check-bounds [line column {:keys [row end-row col end-col] :as _usage}]
   (cond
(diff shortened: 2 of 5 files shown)


src/clojure_lsp/crawler.clj
```diff
@@ -296,18 +296,22 @@
...
-(defn find-project-settings [project-root]
+(defn find-raw-project-settings [project-root]
   (let [config-path (Paths/get ".lsp" (into-array ["config.edn"]))]
     (loop [dir (uri->path project-root)]
       (let [full-config-path (.resolve dir config-path)
             file (.toFile full-config-path)
             parent-dir (.getParent dir)]
         (cond
           (.exists file)
-          (edn/read-string {:readers {'re re-pattern}} (slurp file))
```
ericdallo, line 307: it was needed to declare this function that just return the file string otherwise, the returned config edn to the user is not human readable.

src/clojure_lsp/handlers.clj
```diff
@@ -489,6 +483,33 @@
...
+                        (ApplyWorkspaceEditParams.
+                          (interop/conform-or-log ::interop/workspace-edit result)))))))
+
+(defn ^:private execute-server-info []
+  (let [db @db/db
+        info {:project-root (:project-root db)
+              :project-settings (crawler/find-raw-project-settings (:project-root db))
+              :client-settings (:client-settings db)
+              :port (:port db)}]
+    (interop/conform-or-log ::interop/server-info info)))
```
snoe, line 503: The interop specs are used for conforming into lsp4j objects and less about "type" checking.

I think before conforming, this should result in a clojure map like 
`{:message-type :info :message <string(markdown?) of server info>}` 
and then
`(conform-or-log ::interop/show-message result)` would build a ShowMessage instance in lsp4j https://microsoft.github.io/language-server-protocol/specifications/specification-current/#window_showMessage

src/clojure_lsp/main.clj
```diff
@@ -315,24 +307,18 @@
...
-                                                       (inc (int col))
-                                                       command
-                                                       args)]
-                  (.get (.applyEdit (:client @db/db)
-                                    (ApplyWorkspaceEditParams.
-                                      (interop/conform-or-log ::interop/workspace-edit result)))))
-                (catch Exception e
-                  (log/error e)))))))
-    (CompletableFuture/completedFuture 0))
+   (go :executeCommand
```
snoe, line 312: My reading of ExecuteCommand is that it doesn't expect results back, but rather triggers a server->client message elsewhere. So I don't think this should be switched to supplyAsync.

Rather the server-info branch would probably call `(.showMessage (:client @db/db) conformed result)`

ericdallo: Actually `executeCommand` can return `any` response according to spec, but for this case, I think using a `showMessage` would suit better for what we need.

src/clojure_lsp/handlers.clj
```diff
@@ -489,6 +483,33 @@
...
       (log/warn "Could not find a form at this location. row" line "col" column "file" doc-id))))
 
+(defn ^:private execute-refactor [command args]
+  (let [[doc-id line col & args] (map interop/json->clj args)]
+    (when-let [result (refactor doc-id
+                                (inc (int line))
+                                (inc (int col))
+                                command
+                                args)]
+      (.get (.applyEdit (:client @db/db)
```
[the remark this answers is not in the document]
ericdallo: Yeah, I thought about that, but as the refactor command can return something different from the executeCommand, we would need to do the ` cond` logic here, is that what you suggest?

ericdallo:
Implemented support for this on emacs lsp-mode: https://github.com/emacs-lsp/lsp-mode/commit/4145a70ce1d4bfb2463606ba34c5965080b080d9

PR #172, opened 2020-08-20
remove unused refers when clean-ns

ericdallo:
fix-unused-refers

_this PR only remove the ns if all refers are unused, I'll implement the functionality for one or more unused later._

### DIFF
--- a/src/clojure_lsp/crawler.clj
+++ b/src/clojure_lsp/crawler.clj
@@ -68,18 +68,20 @@
                   (str "Unused declaration: " (:str usage)))
        :severity 2})))
 
-(defn ^:private process-unused-aliases
-  [usages declared-aliases]
+(defn ^:private process-unused
+  [usages declarations]
   (let [ensure-sym (fn [s] (when-not (string? s) s))]
     (->> usages
-         (remove (comp #(contains? % :declare) :tags))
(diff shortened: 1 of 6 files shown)


ericdallo:
@snoe I didn't see any problem adding the `:declare` and `:alias` to the `:refer` tags, let me know what you think, it's kind of confusing for me what they mean truly.

snoe:
I'm not sure that's right, some background:

There should only ever be one declaration (usage tagged `:declare`) for a thing in the project, so a referred symbol would be `:declare` inside the required ns.

An `:alias` tag is strictly for aliased namespaces and comes into play as the namespace portion of symbols and keywords. Marking the `refer` as `alias` would probably introduce some strange completion issues with the refer symbol being in the namespace slot.

snoe:
So removing to remove refers you would probably need to ask crawler for unused-refers (or combine it with `unused-aliases`) and use it in [here](https://github.com/snoe/clojure-lsp/blob/master/src/clojure_lsp/refactor/transform.clj#L281)

ericdallo:
I see, thanks for the great explanation @snoe, I'll try to find the refers together with the alias or combine like you said

ericdallo:
@snoe I hope now it's ok

snoe:
This is great, thanks!

PR #173, opened 2020-08-20
Remove unused code

ericdallo:
* Remove comment blocks
* Remove unused imports
* Rename some unused args adding a _ prefix
* Remove unused functions

### DIFF
--- a/.lsp/config.edn
+++ b/.lsp/config.edn
@@ -3,4 +3,5 @@
                      and [[:inner 1]]
                      or [[:inner 1]]
                      are [[:inner 0]]}}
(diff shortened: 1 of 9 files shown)


snoe:
A few things, I think we should add logging ns to ignored warnings in a kondo config/and for clean-ns `{:unused-namespace {:exclude [clojure.tools.logging]}`
`zspy` and `zass` are debugging functions that should not be removed.
`back-to-mark` and `inside?` should not be removed as they have been useful utilities while hacking on refactorings.

snoe:
I think also the comment block in `parser` should be kept, I use it as a bit of a repl history. Maybe if I understood the motivation behind dropping comment blocks though, I'd change my mind.

ericdallo:
No problem, I'll change those

PR #175, opened 2020-08-27
Add support for keywords on settings

ericdallo:
With this, we can support keywords on our settings file without removing the string keys support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ Other clients might provide a higher level interface to `workspace/executeComman
 
 It is possible to pass some options to clojure-lsp through clients' `InitializationOptions`. Options are a map with keys:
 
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -138,7 +138,7 @@ Other clients might provide a higher level interface to `workspace/executeComman
 
 It is possible to pass some options to clojure-lsp through clients' `InitializationOptions`. Options are a map with keys:
 
-`source-paths` value is a vector of project-local directories to look for clj/cljc/cljs files. Default is `["src","test"]`.
+`source-paths` value is a vector of project-local directories to look for clj/cljc/cljs files. Default is `#{"src", "test"}`.
```
daemianmack, line 141: Says `vector` but default is a set.

ericdallo: Thanks!

PR #176, opened 2020-08-27
WIP - kondo on references

ericdallo:
Things to fix on this branch
[x] Replace all diagnostics with kondo
[ ] Replace all `file-envs` with `uri->analysis` usages

Things that need to be resolved in kondo in order to drop our parser.clj:

[ ] analysis for start/end text location - so analysis entry can be determined based on cursor position
[ ] analysis for param & binding (defn/let) definitions and usages AND valid range for them 
[ ] [maybe] analysis for keywords


snoe:
Hi @borkdude I spent some time with the kondo analyzer and I'm wondering if you are ok adding these things to kondo's analyzer, or would you consider them outside of the scope of your project? You've also been helpful on telling me when I'm off base in the past, so that would be helpful too.

The points in the top comment (which I added) basically all pointing to the problem of finding information about the [thing under the cursor](
https://github.com/snoe/clojure-lsp/pull/176/commits/d8c7ff9bb4b3947e44678b5c673d0b7f46df9765#diff-f8159cdbc0e017113c5c7e2a28b7ca93R40). A reverse lookup from text position to entry (which obviously doesn't have to function in this crude fashion).

For example, with the cursor at the `|`: `(defn foo [bar] [::qaz/qux "string|"])` 
- The code in the diff above will return the `defn` entry because its :col is 2 and `foo` has a :col of 1 (and the other tokens are non-existent). 
- Even if `bar` and `::qaz/qux` were in analysis and col pointed to the first char of each name, it would be nice if I could say that the cursor is not on anything interesting in this scenario.

Then for bindings and params that define "scopes", being able to `rename` an `x` in `(let [x 1] (let [x x] x))` and find the proper definition and usages depending on which `x` you are on would also be important for us.

borkdude:
@snoe For detecting in which defn one is:

Currently clj-kondo emits row and col for var-definitions, but we could easily also emit end-row and end-col since we already have that information. Using that extra position you can then infer which defn the user was visiting.

borkdude:
About locals: clj-kondo knows where a local is introduced and where the local is used. We could add that to the analysis info, something like: `{:locals [{:name x :row 12 :col 16 :id 123}], :local-usages [{:name x :row 17 :col 17 :id 123}]}`? Other suggestions about the exact format welcome. I'd like to have issues in clj-kondo for this as well.

ericdallo:
About the usages, I think that could work, WDYT @snoe?

snoe:
> For detecting in which defn one is in
Hrm I don't think this is just solved by end-col.
In the example `(defn foo [bar]...)` foo has a :col of 1 and defn has a :col of 2 but the span of the foo symbol is :col 7-9. We would need to be be able to unambiguously look at a position and get the analysis entry from that, whether from definitions or usages.

borkdude:
@snoe I'm sorry, I don't understand your last reply.

borkdude:
I might have misunderstood your examples then. I'm not sure if the suggested feature can be easily by clj-kondo, although it can still offer the local information like suggested. Instead of the `:id` I suggested earlier we can also just join the location information:

```
{:locals [{:name x :row 12 :col 16 :id 123}], :local-usages [{:name x :row 17 :col 17 :bound-row 12 :bound-col 16}]}
```

ericdallo:
Closing as we start implementing it here: https://github.com/clojure-lsp/clojure-lsp/pull/261

PR #178, opened 2020-08-27
Update kondo and only run when project changes

snoe:
In java was running into a GC overhead/out of memory on large classpaths
https://github.com/borkdude/clj-kondo/pull/964

Kondo also was running on every init instead of checking `cp-cache`
Now should only happen if your dependencies change.


.clj-kondo/config.edn
```diff
@@ -0,0 +1 @@
+{}
```
ericdallo, line 1: Do we need this?

PR #183, opened 2020-09-09
Add support for renaming files

ericdallo:
With that after we call `textDocument/rename` on a namespace symbol, we will send a `RenameFile` to client too


snoe (approved):
This is awesome, thanks!

PR #197, opened 2020-12-03
Add doc-id, row, and col as arguments to code-lens-references command in codeLens/resolve response

### LINKED ISSUE #196: Add arguments to command in codeLens/resolve response
I noticed the `codeLens/resolve` response includes a range and a command, but the command does not include arguments. I'm not sure how other clients handle this, but in VS Code, when I click a code lens in the editor, it calls the `code-lens-references` command (which I had to implement). 

Inside this command handler, I do not have the range information from the server response, so I don't have the location of the symbol to find references for. I think if the command includes the range in the arguments too then this can more easily be passed to the command. I'm new to this LSP stuff, so maybe I am misunderstanding something.

I'm using the vscode-languageclient, and I don't see a way to manually read the codeLens/response message to get the range info, but even if there is, that seems like the wrong way to go about this, if what I assume above is correct.

```json
{
  "range": {
    "start": {
      "line": 71,
      "character": 6
    },
    "end": {
      "line": 71,
      "character": 10
    }
  },
  "command": {
    "title": "1 references",
    "command": "code-lens-references"
  }
}
```

bpringe:
I made this change so that peeking code references by clicking the code lens would work in VS Code. This works well with the vscode-languageclient, and since it's additive I don't think it would cause any issues with other clients.

Updated related tests.

Fixes #196

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@
 /node_modules
 /test.clj
 /.clj-kondo/.cache
+/.clj-kondo/cache
+/.calva/output-window
--- a/src/clojure_lsp/handlers.clj
+++ b/src/clojure_lsp/handlers.clj
@@ -661,4 +661,4 @@
                           count
                           (str " references"))
              :command "code-lens-references"
-             :arguments {:range range}}})
+             :arguments [doc-id row col]}})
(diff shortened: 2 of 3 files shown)


bpringe:
Pic for reference
image

ericdallo (approved):

PR #201, opened 2020-12-07
Fix mixed JSON and EDN syntax in README

devn:


ericdallo (approved):
Thanks!
I thought I changed everything to edn but It seems I forgot some yet :sweat_smile:

PR #207, opened 2020-12-18
Add initialization options that allow disabling formatting providers

### LINKED ISSUE #206: Add initialization options to disable formatting providers
Calva provides its own formatter that is a customized version of cljfmt. I think the plan is to eventually merge any changes we need into main cljfmt and use it from clojure-lsp, but for now having lsp provide formatting can cause some confusion (VS Code allows a user to choose which formatter to use, but since it sees them as both coming from Calva, they are named the same in the UI picker).

Until we can use cljfmt from lsp, being able to disable formatters from lsp is desirable.

bpringe:
Fixes #206


ericdallo (approved):
LGTM, thank you!

Issue #208, opened 2020-12-18
textDocument/documentSymbol response contains duplicates for cljc files


bpringe:
Using clojure-lsp with Calva in VS Code, for some reason the client sends duplicate `codeLens/resolve` requests, so it receives duplicate responses. Also, the responses contain 1 more count of references than actually exist.

![image](https://user-images.githubusercontent.com/12722744/102655627-76f17e00-4127-11eb-90f8-30397141a137.png)

If I click one of those references to peek references, it shows only 1 reference in the peek window (which is correct). If I remove that reference to `hello`, the codeLens update to show 0 references (also correct). If any number of references above 0 is returned by lsp, the number is off by one, positively.

bpringe:
The client sending duplicate requests seems more of an issue of VS Code, so I'll have to investigate this more. Could duplicate requests cause lsp to send an inaccurate references count?

ericdallo:
This is really weird, it works pretty well using Emacs (client is [lsp-mode](https://emacs-lsp.github.io/lsp-mode/))
![image](https://user-images.githubusercontent.com/7820865/102658153-c484df80-4155-11eb-96f5-f9727900d0aa.png)

bpringe:
I just realized that clojure-lsp is actually returning duplicates in the `textDocument/codeLens` response.

`textDocument/codeLens` request from client:

```json
{
  "textDocument": {
    "uri": "file:///home/brandon/development/simplefi/src/simplefi/data_transfer_specs.cljc"
  }
}
```

`textDocument/codeLens` response:

```json
[
  {
    "range": {
      "start": {
        "line": 0,
        "character": 4
      },
      "end": {
        "line": 0,
        "character": 32
      }
    },
    "data": [
      "file:///home/brandon/development/simplefi/src/simplefi/data_transfer_specs.cljc",
      1,
      5
    ]
  },
  {
    "range": {
      "start": {
        "line": 8,
        "character": 5
      },
      "end": {
        "line": 8,
        "character": 19
      }
    },
    "data": [
      "file:///home/brandon/development/simplefi/src/simplefi/data_transfer_specs.cljc",
      9,
      6
    ]
  },
  {
    "range": {
      "start": {
        "line": 35,
        "character": 5
      },
      "end": {
        "line": 35,
        "character": 29
      }
    },
    "data": [
      "file:///home/brandon/development/simplefi/src/simplefi/data_transfer_specs.cljc",
      36,
      6
    ]
  },
  {
[...]

PR #211, opened 2020-12-20
Fix duplicated usages/references for cljc files

ericdallo:
#fix 208

This PR changes how we save usages on db. We now save `file-type` as a set, so instead of 2 usages with `clj` and another with `cljs` file-types, we now persist only one with `#{clj cljs}` as file-type.
This is only applicable for non `norename` tags as we just want to do that for user namespace/alias usages

### DIFF
--- a/src/clojure_lsp/crawler.clj
+++ b/src/clojure_lsp/crawler.clj
@@ -203,7 +203,7 @@
                       [uri (safe-find-references uri text false true)])))
              (remove (comp nil? second)))
         output-chan (async/chan 5)]
-    (async/pipeline-blocking 5 output-chan xf (async/to-chan jars) true (fn [e] (log/warn e "hello")))
+    (async/pipeline-blocking 5 output-chan xf (async/to-chan! jars) true (fn [e] (log/warn e "hello")))
     (async/<!! (async/into {} output-chan))))
 
 (defn crawl-source-dirs [dirs]
@@ -219,7 +219,7 @@
                                   (safe-find-references uri (slurp uri) false false))))
             (remove (comp nil? second)))
         output-chan (async/chan)]
-    (async/pipeline-blocking 5 output-chan xf (async/to-chan dirs) true (fn [e] (log/warn e "hello")))
+    (async/pipeline-blocking 5 output-chan xf (async/to-chan! dirs) true (fn [e] (log/warn e "hello")))
     (async/<!! (async/into {} output-chan))))
 
 (defn lookup-classpath [root-path command-args env]
--- a/src/clojure_lsp/db.clj
+++ b/src/clojure_lsp/db.clj
@@ -11,7 +11,7 @@
 (defonce diagnostics-chan (async/chan 1))
 (defonce edits-chan (async/chan 1))
 
-(def version 2)
+(def version 3)
 
 (defn make-spec [project-root]
   (let [lsp-db (io/file (str project-root) ".lsp" "sqlite.1.db")]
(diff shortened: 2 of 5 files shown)


ericdallo:
Confirmed it fixed #208

bpringe:
I'm not sure what in these changes might cause this, but I noticed when the lsp db doesn't exist in a project, it's taking 90+ seconds to return the initialize response. Before these changes, it took only about 40k seconds. I tested at least a couple times with a jar from before these changes and a jar with these changes.

Even with the changes though, if the db exists already, it only takes about 6 seconds.

Any idea how these changes might have more than doubled the initialize response time when no lsp db exists?

ericdallo:
Hum, I don't think this PR could cause that, we just iterate a single time to each usage when persisting which seems to not be a problem.
The initial delay is already a problem since we started to use clj-kondo.
I recommend testing it checking the initialize time displayed on `/tmp/lsp.out` and removing `/your-project/.lsp` folder to reanalyze without any cache

ericdallo:
You can also wrap it with a  `log/debug` with the time spent in that function for debugging purposes

bpringe:
> I recommend testing it checking the initialize time displayed on /tmp/lsp.out and removing /your-project/.lsp folder to

I did this and used the trace logging which says how long each request/response takes. Before these changes, with the .lsp folder removed it took ~30 seconds to respond, and with these changes with the .lsp folder removed, it takes about ~90 seconds to respond.

I agree these changes don't seem like they should cause an issue, just reporting what I've seen. Not sure if anything should be done about it.

bpringe:
Just to note, this was with the same project, of course.

ericdallo:
I'll try to compare the older version and the current to double-check if changed anything, thanks!

ericdallo:
Hey, after a lot of tests I could prove that this PR indeed introduced a regression regarding performance :(
But I could improve it to make it almost as fast as before:

Before the cljc duplicated usages fix:
```clojure
Evaluation count : 6 in 6 samples of 1 calls.
             Execution time mean : 847.996168 ms
    Execution time std-deviation : 42.774901 ms
   Execution time lower quantile : 801.961730 ms ( 2.5%)
   Execution time upper quantile : 913.601543 ms (97.5%)
                   Overhead used : 1.947168 ns
```

[code before](https://github.com/clojure-lsp/clojure-lsp/blob/master/src/clojure_lsp/parser.clj#L1131-L1157) the performance fix.
```clojure
Evaluation count : 6 in 6 samples of 1 calls.
             Execution time mean : 3.575516 sec
    Execution time std-deviation : 224.909583 ms
   Execution time lower quantile : 3.304419 sec ( 2.5%)
   Execution time upper quantile : 3.841209 sec (97.5%)
                   Overhead used : 1.947168 ns
```

[With the performance fix](https://github.com/clojure-lsp/clojure-lsp/pull/243)
```clojure
             Execution time mean : 877.753781 ms
    Execution time std-deviation : 12.686530 ms
   Execution time lower quantile : 857.340680 ms ( 2.5%)
   Execution time upper quantile : 890.014269 ms (97.5%)
                   Overhead used : 1.947168 ns
```

This function is used every time we need to crawl a code in a file to parse and understand the symbols, since we could have a lot of file for multiple jars/source-dirs, this should improve the performance as it was before :)

note: I did the performance tests using [criterium](https://github.com/hugoduncan/criterium) benchmark tool.

bpringe:
Awesome find!

PR #214, opened 2020-12-24
Add Calva reference to readme.

bpringe:
Feel free to modify the wording here, if you approve of adding this.


bpringe:
Maybe the proof of concept line could be removed? But Calva uses java to run the jar, so that PoC is different and maybe should be left as reference. :man_shrugging:

ericdallo:
Thank you!

PR #226, opened 2020-12-30
Fix jar file URIs to be compatibile with Windows and Unix

### LINKED ISSUE #223: Refactorings and jar definition lookups do not work on Windows
I am not sure this is not related to Calva, but it seems we do things right on the client end when I look at it.

Anyway, for this example, on Windows,, definitions for `remove`, `apply`, `mapv` etcetera do not work. Definitions for `util/add7` and for `v` works.

```clojure
(ns pez.lab.seven-segments
  (:require [clojure.string :as string]
            [pez.util :as util]))

(comment
  (util/add7 8))

(defn transpose [v]
  (->> v
       (remove nil?)
       (apply mapv vector)))
...
```

Also refactorings like trying to unwind the thread on `v` doesn't work. It gives a warning in the log:

```
WARN clojure-lsp.handlers: Could not find a form at this location. ...what looks like the right location to me...
```

This seem somewhat covered by https://github.com/clojure-lsp/clojure-lsp/issues/25

bpringe:
Also bump db version so changes work with clients without removing db.

Fixes #223

This change seems to work on both Windows and Linux for jar lookups. Refactorings, however, still do not work on Windows. I'm investigating that now.

### DIFF
--- a/src/clojure_lsp/crawler.clj
+++ b/src/clojure_lsp/crawler.clj
@@ -189,7 +189,7 @@
                               (remove #(.isDirectory %))
                               (map (fn [entry]
                                      [(if (= "jar" dependency-scheme)
-                                        (str "jar:file://" jar-file "!/" (.getName entry))
+                                        (str "jar:file:///" jar-file "!/" (.getName entry))
                                         (str "zipfile://" jar-file "::" (.getName entry)))
                                       entry
                                       jar]))))))
--- a/src/clojure_lsp/crawler.clj
+++ b/src/clojure_lsp/crawler.clj
@@ -61,7 +61,7 @@
                               (enumeration-seq)
                               (remove #(.isDirectory %))
                               (map (fn [entry]
-                                     [(if (= "jar" dependency-scheme)
+                                     [(if jar-dependency-scheme?
                                         (str "jar:file:///" jar-file "!/" (.getName entry))
                                         (str "zipfile://" jar-file "::" (.getName entry)))
                                       entry
--- a/src/clojure_lsp/db.clj
+++ b/src/clojure_lsp/db.clj
@@ -11,7 +11,7 @@
 (defonce diagnostics-chan (async/chan 1))
 (defonce edits-chan (async/chan 1))
 
-(def version 4)
+(def version 5)
 
 (defn make-spec [project-root]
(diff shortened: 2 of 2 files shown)


bpringe:
Btw, I noticed that if we changed `crawl-jars` to concat `file:/` (one slash) it then did not work on Linux but worked on Windows. Two slashes works on Linux but now Windows. Three+ slashes seems to work on both.

bpringe:
On linux when I call thread-first when my cursor next to the last form in the comment in the following code:

```clojure
(ns windows
  (:require [hiccup.core :as h]))

(comment
  (h/html [:div "hello"])
  
  (inc (first [1 2])))
```

It works and I see in the logs:

```
INFO  clojure-lsp.main: Executing command thread-first with args #object[java.util.ArrayList 0x1ab66a5c ["file:///home/brandon/development/clojure-test/src/windows.clj", 6, 2]]
```

On Windows, however, with the same code, it does not work, and I see this warning in the logs:

image

`could not find a form at this location. row 7 col 3.` - the args sent though were row 6 col 2, and these work on Linux with the same code on the same lines. Just posting some findings here as I go.

bpringe:
@ericdallo I increased the db version so that the changes are picked up on projects that have been crawled by lsp before. I assume that's partly what the version is for?

src/clojure_lsp/crawler.clj
```diff
@@ -61,8 +61,8 @@
                               (enumeration-seq)
                               (remove #(.isDirectory %))
                               (map (fn [entry]
-                                     [(if jar-dependency-scheme?
-                                        (str "jar:file://" jar-file "!/" (.getName entry))
+                                     [(if (= "jar" dependency-scheme)
```
ericdallo, line 64: Why we need this change?

bpringe: Oops, must have merged poorly. I meant to accept the new `jar-dependency-scheme?`. I'll fix.

ericdallo (commented):
LGTM, let me test it manually on linux

ericdallo (approved):
Tested it on my NixOS and it works!
Thank you for the huge help on this @bpringe

bpringe:
Awesome! You're welcome, and thank you.

PR #230, opened 2020-12-31
Add flag to crawl jars async on startup

ericdallo:
This PR make 2 important changes:

* Most important: add a `async-jars-on-startup?` flag which will allow start server without scan jars, this will alow user interact with everything from the project but going to jars/getting info from symbols of jars. After the async jar crawl is done, it will send a LSP info message to user saying that the jars were scanned successfully, with that, user can use jar features normally. i.e: `clojure-lsp` has **8** source-dirs and **120** jars, the jar scan is almost 4 times slower than src scan. 

* Use `clojure.core.async` on project crawling (source, file-envs and jar-crawling), crawling in parallel instead of waiting for source, file-envs and jar-crawling in this sync order.
This just starts async the 3 crawlers, but we still need to wait for all 3 to finish to respond to client. (unless above flag is one, then we just wait for source and file-envs crawls).

### DIFF
--- a/src/clojure_lsp/crawler.clj
+++ b/src/clojure_lsp/crawler.clj
@@ -1,21 +1,21 @@
 (ns clojure-lsp.crawler
   (:require
-   [clojure-lsp.db :as db]
-   [clojure-lsp.feature.diagnostics :as f.diagnostic]
-   [clojure-lsp.feature.references :as f.references]
-   [clojure-lsp.shared :as shared]
-   [clojure-lsp.window :as window]
-   [clojure.core.async :as async]
-   [clojure.edn :as edn]
-   [clojure.java.io :as io]
-   [clojure.java.shell :as shell]
-   [clojure.set :as set]
-   [clojure.string :as string]
-   [clojure.tools.logging :as log]
-   [digest :as digest])
+    [clojure-lsp.db :as db]
+    [clojure-lsp.feature.diagnostics :as f.diagnostic]
+    [clojure-lsp.feature.references :as f.references]
+    [clojure-lsp.shared :as shared]
+    [clojure-lsp.window :as window]
+    [clojure.core.async :as async]
+    [clojure.edn :as edn]
+    [clojure.java.io :as io]
+    [clojure.java.shell :as shell]
+    [clojure.set :as set]
+    [clojure.string :as string]
+    [clojure.tools.logging :as log]
+    [digest :as digest])
   (:import
-   [java.util.jar JarFile]
-   [java.nio.file Paths]))
+    [java.util.jar JarFile]
+    [java.nio.file Paths]))
 
 (defn ^:private file->uri [file]
   (str (.toUri (.toPath file))))
@@ -47,11 +47,12 @@
         declarations (f.diagnostic/usages->declarations usages)
         excludes (-> (get-in @db/db [:settings :linters :unused-namespace :exclude] #{}) set)
         declared-refers (->> declarations
-                              (filter (comp #(contains? % :refer) :tags))
-                              (remove (comp excludes :ns)))]
+                             (filter (comp #(contains? % :refer) :tags))
+                             (remove (comp excludes :ns)))]
     (process-unused usages declared-refers)))
 
 (defn crawl-jars [jars dependency-scheme]
+  (log/info "Crawing" (count jars) "jars...")
   (let [jar-dependency-scheme? (= "jar" dependency-scheme)
         xf (comp
              (mapcat (fn [jar-file]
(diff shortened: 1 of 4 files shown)


snoe:
Hrm be careful with this one. Make sure you're not running out of thread pool with 3x threads each pipelining. 

Also I'm not sure if Kondo will change this. It's quite possible that proper analysis requires jars to be known before analyzing other files. 

Jars are slow but with the cache it should be a one time cost. Make sure we're not getting ourselves into a corner exposing this option to the user.

ericdallo:
Thanks for taking a look @snoe!
Currently, we use kondo only for analysis of the current file open file, right? so I don't think crawling async jars could affect that.
I made some tests and it seems to me to works really well, do you see any other issue this may cause?

bpringe:
@ericdallo Let me know when/if you want me to try this. I was looking at the changes then noticed you pushed again :smile: . I may not get to it until a day or two from now.

ericdallo:
It's done, it was missing a fix that I forgot to push.
Np

ericdallo:
BTW @bpringe its easier to review this with `Hide whitespace changes` enabled ;)

ericdallo:
@bpringe Sorry, I did one last change... After @snoe heads up, I tested this branch with a huge clojure project and realized that we can have OutOfMemory when we scan a lot of jars and clj-kondo crawl the project in other thread.
I fixed only crawling the jars after clj-kondo scan, which is enough to start the LSP server and user use all project-related features, and use jar features after the async as well. 

This way we have a safer approach which may improve project start and speed and may fit for most users, we can think on changing the default of this flag to true in the future.

src/clojure_lsp/crawler.clj
```diff
@@ -46,11 +47,12 @@
...
         excludes (-> (get-in @db/db [:settings :linters :unused-namespace :exclude] #{}) set)
         declared-refers (->> declarations
-                              (filter (comp #(contains? % :refer) :tags))
-                              (remove (comp excludes :ns)))]
+                             (filter (comp #(contains? % :refer) :tags))
+                             (remove (comp excludes :ns)))]
     (process-unused usages declared-refers)))
 
 (defn crawl-jars [jars dependency-scheme]
+  (log/info "Crawing" (count jars) "jars...")
```
snoe, line 55: Typo

src/clojure_lsp/crawler.clj
```diff
@@ -142,47 +152,66 @@
...
+                                   (crawl-source-dirs (:directory classpath-entries-by-type))))
+        jar-envs-chan (async/go
+                        (cond
+                          use-cp-cache (:jar-envs loaded)
+                          async-scan-jars-on-startup? {}
+                          :else (crawl-jars jars dependency-scheme)))
+        jar-envs (async/<!! jar-envs-chan)]
+    (db/save-deps root-path project-hash classpath jar-envs)
+    (start-clj-kondo-scan! use-cp-cache classpath)
+    (when start-clj-kondo-scan!
```
snoe, line 193: this looks wrong

snoe:
I also tested startup on my large project, on master it takes under 3 mins to crawl everything, on this branch it takes 4 min 30 and even though is is "ready" at about 3 mins too it is unusable because load is over 10 at that point. I suspect disk io is the bottleneck and more threads don't always help (this is a 2020 16gb mbp). This unfortunately doesn't seem like a win to me.

@ericdallo 
> Currently, we use kondo only for analysis of the current file open file, right? so I don't think crawling async jars could affect that.
I was referring to when we make the switch away from our parser to kondo's analysis - I believe (but could be mistaken) that jar dependencies may need to be crawled first, the cache definitely made a difference in linting. Java is very particular about dependency order and crawling in the wrong order could cause mistakes.

In the code as it exists today, it looks like db is used in `:refer all` to add symbols from that namespace to the current context, otherwise they would get tagged as unknown. If the jar's get crawled after, then find-references would miss any of those. So there's an ordering issue and a reason jars were crawled first. Maybe it's rarely used these day, but it would be great to get rid of `clojure_core.clj` and I think that'll only happen with strict ordering like this.

Another potential issue, I've had projects that shadow packages from other jars. So if you have a package like `korma.core` in your project to get around something in the jar that provides korma.core, I suspect after the jars are crawled go to definition will go to the jar instead of the project file.

So:
That very first startup is painful, but afterwards and day to day, I personally don't notice it. 

One alternative:
If people are changing deps a lot and if we wanted to be smarter about speeding it up, we don't need to depend on the project hash, but could keep a hash or mtime + path of each jar (or file in each jar). 

It's also really important to not change the `db/version` unless absolutely necessary, which maybe is what precipitated this work?

ericdallo:
I see, thank you for the tests @snoe, it makes sense indeed.
Do you think when we change to clj-kondo on everything, would clojure-lps starts faster?
Should we close this PR so? I really would like to have something that removes these huge start time, or tries to move non-critical things to later for a better user UX

snoe:
Yeah, I think. kondo will help a lot, I'm not sure how much overhead analysis takes but on that project kondo only takes 17s.
I think it also might have finer grained caching, like i described so that could help too. 

graal too would be a huge win. 

I guess the question is what's critical? Then what tradeoffs are we making to get the critical stuff working without a full picture.

ericdallo:
Agree, kondo takes 200s on my huge project and it only analyzes the classpath, not jars ATM.
Ok, I'll close this PR, thanks for the insights!

bpringe:
> It's also really important to not change the db/version unless absolutely necessary, which maybe is what precipitated this work?

Just want to note that I changed it recently after fixing an issue with providing jar content on Windows, and this was because lsp was not re-analyzing the project with the new code, unless I bumped the version (so the fix wasn't working, essentially, without deleting the sqlite db). Just noting in case this was a concern/question. Thinking I did the right thing there.

PR #235, opened 2021-01-05
Improve "add missing namespace" code action

ericdallo:
This PR improves the add missing namespace code action:

- Add the namespace which will be required so `Add missing namespace` -> `Add 'clojure.test' missing namespace`
- It suggests known common alias namespaces like `set` -> `clojure.set`, `walk` -> `clojure.walk`, `string` -> `clojure.string` etc
- It suggests known common refers like `deftest`, `testing`, `is`, `are`, `fact`, `facts` etc.

clojure-code-action-add-missing-ns

### DIFF
--- a/src/clojure_lsp/feature/code_actions.clj
+++ b/src/clojure_lsp/feature/code_actions.clj
@@ -2,7 +2,8 @@
   (:require
    [clojure-lsp.feature.definition :as f.definition]
    [clojure-lsp.feature.refactor :as f.refactor]
-   [clojure-lsp.refactor.transform :as r.transform]))
+   [clojure-lsp.refactor.transform :as r.transform]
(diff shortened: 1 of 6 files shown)


src/clojure_lsp/feature/code_actions.clj
```diff
@@ -13,20 +14,24 @@
         line (dec row)
         character (dec col)
         has-unknow-ns? (some #(= (compare "unresolved-namespace" (some-> % .getCode .get)) 0) diagnostics)
```
Cnly, line 16:
```suggestion
        has-unknown-ns? (some #(= (compare "unresolved-namespace" (some-> % .getCode .get)) 0) diagnostics)
```
Tiny typo :)

Issue #237, opened 2021-01-06
Handle classpath-cmd on Windows without config change


bpringe:
In order to make clojure-lsp work on Windows for a typical project, the `:classpath-cmd` in `.lsp/config.edn` (`:project-specs`) needs to be set to `["powershell.exe" <command+args for clj or lein>]`.

It seems that using `lein.bat` as the command is not necessary, so long as the first command is `powershell.exe`. Therefore, I think if clojure-lsp would change its default classpath command to add `powershell.exe` to the beginning if the OS is Windows, this would remove the requirement for people to add a custom classpath cmd in their config (for typical projects).

I think this would improve the out-of-the-box experience. The alternative (for us) would be Calva modifying the `.lsp/config.edn`, but I don't think this is a good route.

ericdallo:
It makes total sense to improve the out-of-the-box experience, but AFAIK we don't know which OS user is using.
I found [these default settings](https://github.com/clojure-lsp/clojure-lsp/blob/0efbc42acdf92e4d9de3980f87c31fe71ab91af7/src/clojure_lsp/crawler.clj#L123-L131), I think adding to it maybe fix the issue?

bpringe:
We can tell which OS the user is using via `(System/getProperty "os.name")`.

Thanks for pointing me to those settings. Another thing I noticed is that with shadow-cljs it assumes they have `shadow-cljs` installed, but this is true also with the other commands.

In addition to adding `powershell.exe` as the first command if the OS is Windows, I think in any case, clojure-lsp should send a message to the client informing when there is an error looking up classpath info (as it logs in clojure-lsp.out). This way, at least when their is an error, they will know immediately.

ericdallo:
Ok, I think we can check the OS that way and add it if it is Windows, I just not sure how it'd looks like the config for Windows:
```clojure
  {:project-path "?"
    :classpath-cmd ["powershell.exe" "?"]}
```

> In addition to adding powershell.exe as the first command if the OS is Windows, I think in any case, clojure-lsp should send a message to the client informing when there is an error looking up classpath info (as it logs in clojure-lsp.out). This way, at least when their is an error, they will know immediately.

We can send a message using `windows/message` LSP method with `error` type, then the client shows to user automatically :)

bpringe:
>We could send a message using windows/message LSP method with error type, then the client shows to user automatically :)

Yes, that's what I was meaning. :smiley:

ericdallo:
[...]

PR #238, opened 2021-01-07
WIP - Handle classpath lookup on Windows without config change

bpringe:
This is a work in progress, and also includes showing an error message via `window/showMessage` when the classpath lookup fails, the the user of the client is immediately made aware something went wrong.

### DIFF
--- a/src/clojure_lsp/crawler.clj
+++ b/src/clojure_lsp/crawler.clj
@@ -3,6 +3,7 @@
    [clojure-lsp.db :as db]
    [clojure-lsp.feature.diagnostics :as f.diagnostic]
    [clojure-lsp.feature.references :as f.references]
+   [clojure-lsp.interop :as interop]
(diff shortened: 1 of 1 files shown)


ericdallo:
Hey @bpringe really thanks for the PR, but I'm finishing the fix for #237 and adding the window message 😅 
I didn't know you would take a look on this, sorry

ericdallo:
I will push the fix in some minutes!

ericdallo:
Done here: https://github.com/clojure-lsp/clojure-lsp/commit/1d8a0fb065cd44be14d3d169892df9aaed614459

bpringe:
No problem! I got a little more practice with developing using the repl connected to Calva. Had fun calling `.showMessage` from the repl in clojure-lsp and seeing it pop up in Calva :smile:

ericdallo:
Nice! yeah, developing with the connected REPL is very handy!

PR #241, opened 2021-01-07
Fix classpath-cmd modification for Windows

### LINKED ISSUE #237: Handle classpath-cmd on Windows without config change

bpringe:
Fixes #237 (slight change to previous fix)

### DIFF
--- a/src/clojure_lsp/crawler.clj
+++ b/src/clojure_lsp/crawler.clj
@@ -125,7 +125,7 @@
 (defn ^:private classpath-cmd->windows-safe-classpath-cmd
   [classpath]
   (if (shared/windows-os?)
-    (concat ["powershell.exe"] classpath)
+    (into ["powershell.exe"] classpath)
     classpath))
 
 (def ^:private default-project-specs


bpringe:
There was a use of `into` elsewhere on what was previously a list instead of a vector, because of the `concat` that was changed here to `into`, so the shell args were not in the correct order.

bpringe:
Test on Windows, btw.

ericdallo (approved):
Thank you for the help!

PR #248, opened 2021-01-10
Use MkDocs to Generate Website Assets

dcfrankel:
I was bored and went ahead and created a simple mkdoc configuration for generating a static website for the project and its documentation. My thought is that we could create a github page for the project to server the generated assets (so the assumption is the url is something like `https://clojure-lsp.github.io`). I also took the liberty of cleaning up some of the formatting in the documentation (although I tried not to change any of the actual information).

This is a very basic setup and I'm happy to make changes and improvements based on any feedback, just let me know! I'll probably update the theme and some other UI elements. I can also work on investigating ways for us to auto build the assets as well if there is interest.

I appreciate the work being done on the LSP so thought I'd try to contribute in some way 😄 .


ericdallo (changes requested):
Hey @dcfrankel, thank you very much for this!
It's on my backlog adding a github pages for that indeed, but I'd prefer if we follow the way I implemented it for [Emacs LSP](https://emacs-lsp.github.io/lsp-mode/), would you mind?

- we don't need to remove the README since it's very used for users that search only via github.
- IMO `mkdocs-material` has a UI better than default mkdocs :p 
- It'd be nice to have a Makefile to build the docs local like [here](https://github.com/emacs-lsp/lsp-mode/blob/master/Makefile#L95-L99)
- It'd be nice to have a Github action to deploy the pages at `gh-pages` branch like [here](https://github.com/emacs-lsp/lsp-mode/blob/master/.github/workflows/docs.yml)

WDYT? Would you mind make these changes? LMK if need any help too

dcfrankel:
Yep that sounds good and seems simple enough. I did have some issues getting mkdocs to recognize the README.md if it wasn't moved into the docs directory. Do you happen to know how to get that to work?

ericdallo:
Yes! I had that exact same problem, I did a workaround but that work for months now, I just copy the README to the docs folder during the build :)
You can check it [here](https://github.com/emacs-lsp/lsp-mode/blob/master/Makefile#L96) and [here](https://github.com/emacs-lsp/lsp-mode/blob/master/.github/workflows/docs.yml#L33)

dcfrankel:
Ok, I've added all the additions you requested @ericdallo. I took a slightly different approach in the github action (I wasn't sure what the docker image you were using was doing), and simply used the python commands to generate the files. This also applies for the Makefile.

ericdallo:
In `lsp-mode` we use a custom image: https://github.com/emacs-lsp/docs-image/blob/master/Dockerfile, because mkdocs-material doesn't have all the necessary plugins, but I think it's ok we start without that.

dcfrankel:
> In `lsp-mode` we use a custom image: https://github.com/emacs-lsp/docs-image/blob/master/Dockerfile, because mkdocs-material doesn't have all the necessary plugins, but I think it's ok we start without that.

Ah I see, that makes sense. My solution will work fine for the github actions, but you're correct that individuals will need to install python, mkdocs, and mkdocs-material to generate the assets locally.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip 
+          pip install mkdocs-material
+
+      - name: Generate docs 
+        run: |
+          cp -rf README.md docs
+          mkdocs build --clean
```
ericdallo, line 28: We could use directly the mkdocs-material docker image and avoid installing python manually:
```suggestion
          docker run --rm -v ${PWD}:/docs squidfunk/mkdocs-material:6.2.4 -- build
```
WDYT?

ericdallo (commented):
Looks really good @dcfrankel, I think using the docker image this is ready to go, WDYT?

Makefile
```diff
@@ -0,0 +1,6 @@
+local-webpage :
+	cp -rf README.md docs
+	mkdocs build --clean
```
ericdallo, line 3: Yes, if we use the docker image from mkdocs-material, we could replace this for:
```suggestion
	docker run --rm -it -p 8000:8000 -v ${PWD}:/docs squidfunk/mkdocs-material:6.2.4
```

.github/workflows/docs.yml
ericdallo: Since the docker image already exists and we are not using any hype plugin, I think this would fit perfect

Makefile
dcfrankel: Yep, makes sense

dcfrankel:
> Looks really good @dcfrankel, I think using the docker image this is ready to go, WDYT?

Ok, great! I've added your suggested commits.

dcfrankel:
I tested the makefile locally and it seems to generate the static site assets fine. I think this is good to go unless you have any additional changes you'd like to see.

Makefile
```diff
@@ -1,6 +1,6 @@
 local-webpage :
 	cp -rf README.md docs
-	docker run --rm -it -p 8000:8000 -v ${PWD}:/docs squidfunk/mkdocs-material:6.2.4
+	docker run --rm -v ${PWD}:/docs squidfunk/mkdocs-material:6.2.4 -- build
```
ericdallo, line 3: Actually generating local the site instead of binding to a port is not easy to test local changes like check if the docs are good or how the UI behave, but I can change it later

ericdallo (approved):
Amazing work @dcfrankel, thank you very much for the contribution!

PR #251, opened 2021-01-11
Prevent orphaned server processes

bpringe:
Add parent process probe if `processId` is sent in initialize request. Exit server gracefully if parent process is not alive.

The process liveness check has been tested on Linux and Windows.

### DIFF
--- a/src/clojure_lsp/main.clj
+++ b/src/clojure_lsp/main.clj
@@ -431,19 +431,14 @@
         (end
           (apply #'handlers/extension method args)))))
 
-(defn ^:private process-alive?
-  [pid]
-  (let [{:keys [exit]} (shell/sh "kill" "-0" (str pid))]
-    (= exit 0)))
-
 (defn ^:private start-parent-process-liveness-probe!
   [ppid server]
   (future (run!
            (fn [_]
              (Thread/sleep 5000)
-             (log/info "Checking parent process" ppid "liveness")
-             (if (process-alive? ppid)
-               (log/info "Parent process" ppid "is running")
+             (log/debug "Checking parent process" ppid "liveness")
+             (if (shared/process-alive? ppid)
+               (log/debug "Parent process" ppid "is running")
                (do
                  (log/info "Parent process" ppid "is not running - exiting server")
                  (.exit server))))
--- a/src/clojure_lsp/main.clj
+++ b/src/clojure_lsp/main.clj
@@ -431,7 +431,7 @@
         (end
           (apply #'handlers/extension method args)))))
 
-(defn process-alive?
+(defn ^:private process-alive?
   [pid]
   (let [{:keys [exit]} (shell/sh "kill" "-0" (str pid))]
     (= exit 0)))
--- a/src/clojure_lsp/main.clj
+++ b/src/clojure_lsp/main.clj
@@ -433,16 +433,16 @@
 
 (defn ^:private start-parent-process-liveness-probe!
   [ppid server]
-  (future (run!
-           (fn [_]
-             (Thread/sleep 5000)
-             (log/debug "Checking parent process" ppid "liveness")
-             (if (shared/process-alive? ppid)
-               (log/debug "Parent process" ppid "is running")
-               (do
-                 (log/info "Parent process" ppid "is not running - exiting server")
-                 (.exit server))))
-           (range))))
+  (async/go-loop []
+    (async/<! (async/timeout 5000))
+    (log/debug "Checking parent process" ppid "liveness")
+    (if (shared/process-alive? ppid)
+      (do 
+        (log/debug "Parent process" ppid "is running")
+        (recur))
+      (do
+        (log/info "Parent process" ppid "is not running - exiting server")
+        (.exit server)))))
 
 (def server
   (proxy [ClojureExtensions LanguageServer] []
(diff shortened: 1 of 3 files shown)


src/clojure_lsp/main.clj
```diff
@@ -430,6 +431,24 @@
         (end
           (apply #'handlers/extension method args)))))
 
+(defn process-alive?
```
ericdallo, line 434:
```suggestion
(defn ^:private process-alive?
```

src/clojure_lsp/main.clj
```diff
@@ -430,6 +431,24 @@
...
+  [pid]
+  (let [{:keys [exit]} (shell/sh "kill" "-0" (str pid))]
+    (= exit 0)))
+
+(defn start-parent-process-liveness-probe!
```
ericdallo, line 439:
```suggestion
(defn ^:private start-parent-process-liveness-probe!
```

src/clojure_lsp/main.clj
```diff
@@ -430,6 +431,24 @@
         (end
           (apply #'handlers/extension method args)))))
 
+(defn process-alive?
+  [pid]
+  (let [{:keys [exit]} (shell/sh "kill" "-0" (str pid))]
```
ericdallo, line 436: Should we wrap the call with a try-catch block? Not sure if `shell/sh` could throw an exception

bpringe: I'm going to check and add a try/catch if necessary.

src/clojure_lsp/main.clj
```diff
@@ -430,6 +431,24 @@
...
+    (= exit 0)))
+
+(defn start-parent-process-liveness-probe!
+  [ppid server]
+  (future (run!
```
ericdallo, line 441: I think you can use `go` macro since it uses clojure.async/go which is async too but is lighter than another thread, WDYT?

ericdallo: We already use that macro for almost all methods of this namespace

src/clojure_lsp/main.clj
```diff
@@ -430,6 +431,24 @@
...
+           (fn [_]
+             (Thread/sleep 5000)
+             (log/info "Checking parent process" ppid "liveness")
```
ericdallo, line 444: I'd change this one to avoid a lot of logs (I have an idea to remove debug logs on the production relese)
```suggestion
             (log/debug "Checking parent process" ppid "liveness")
```

src/clojure_lsp/main.clj
bpringe: Yeah, I thought about this but wasn't sure which is better for this, given that async `go` runs on the same thread but parks it, whereas the future runs in a new thread. I'm not sure if the use of `Thread/sleep` makes a difference in which one we should use. WDYT?

bpringe: I guess what I mean is will it sleep the thread and block other operations if we used async `go`? Worth a try to see if we don't know, I suppose.

bpringe: Oh I see we can use `<!` with `timeout`. I'm still unsure what the implications of each is (parking vs new thread for this usage), but you may be more familiar with this.

ericdallo: Hum, true, probably it will block the thread.. I think it's okay to go with the future

ericdallo: Oh, yes, probably a timeout may solve that, something like:
```clojure
(defun check-process! []
  (async/go
    (async/<! (async/timeout 1000))
    (if (alive?)
      (check-process!)
      (exit))))
```

src/clojure_lsp/shared.clj
```diff
@@ -1,14 +1,37 @@
...
+   [clojure.java.shell :as shell]
+   [clojure.string :as string]
+   [clojure.tools.logging :as log])
   (:import
    [java.net URI]
    [java.nio.file Paths]
    [org.eclipse.lsp4j Range]))
 
-(defn windows-os? []
+(def windows-os?
```
bpringe, line 11: I changed this because I don't think the OS can ever change mid jvm-run, but I'll change it back if you want. Just prevents a probably very tiny/insignificant amount of repeated work.

bpringe:
Oh, I should clarify that I had trouble reproducing the issue on Windows in order to check that the probe causes the server to exit (my Windows VM is slow and the timing has to be right to reproduce), but I did verify that the logs showed the probe was seeing the parent process as running, and I verified in the REPL that the `process-alive?` function worked as expected.

src/clojure_lsp/shared.clj
ericdallo: It's better this way, nice!

src/clojure_lsp/main.clj
```diff
@@ -430,6 +431,19 @@
         (end
           (apply #'handlers/extension method args)))))
 
+(defn ^:private start-parent-process-liveness-probe!
+  [ppid server]
+  (async/go-loop []
```
ericdallo, line 436: Nice

ericdallo (approved):
Looks really good to me, I tested and didn't see any logs regarding the loop on the logs, then I realized Emacs lsp-mode [do not send processId](https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-mode.el#L6620), I think we can merge and if any bugs check it again, thanks for the help @bpringe

PR #253, opened 2021-01-13
Read .nrepl-port file

danieltdt:
Read `.nrepl-port` file before creating embedded nrepl server

### DIFF
--- a/src/clojure_lsp/main.clj
+++ b/src/clojure_lsp/main.clj
@@ -8,7 +8,6 @@
    [clojure-lsp.shared :as shared]
    [clojure-lsp.window :as window]
    [clojure.core.async :as async]
-   [clojure.java.shell :as shell]
    [clojure.tools.logging :as log]
    [nrepl.server :as nrepl.server]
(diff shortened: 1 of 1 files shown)


ericdallo (commented):
Hey @danieltdt thanks for the PR, I'm not familiar with `.nrepl-port` file, could you explain why do you need these changes?
As I can see, if user has a `.nrepl-port` file, we won't start the nrepl server, right?

danieltdt:
Hi, @ericdallo 👋🏼 
The `.nrepl-port` is a [well-known](https://github.com/nrepl/nrepl/blob/master/doc/modules/ROOT/pages/building_servers.adoc#best-practices) file that nrepl servers create when starting the server process (and erases it when the process terminates).
Leiningen creates that file (with the port as content) when you run `lein repl`.
By check for that file before starting a new server, we can reuse an already running nrepl rather than creating a new server instance.

ericdallo (approved):
Oh, really useful!
Thank you for the explanation :D
LGTM!

PR #261, opened 2021-01-16
Use clj-kondo analysis instead of own parsing

### LINKED ISSUE #240: Move file path in hovers to bottom, under separator
Currently it looks like this in VS Code:

image

And this in Emacs:

image

We should see how it looks on both editors before merging a change.

snoe:
Lots of issues still, need to work on features one by one.

`checklist` file is holding my brain dump of todo items 

### To fix before merge

- [x] find-references are always returning the definition usage, it should return only when `includeDeclaration` is true
- [x] code lenses for references not working
- [X] semantic tokens not working - partially fixed, we can work that after merge, not critical
- [x] spec error on document-symbols
- [x] clean-ns - partially fixed, fix refers/imports after https://github.com/clj-kondo/clj-kondo/issues/1128
- [x] code actions fix
- [x] call hierarchy not working
- [x] hover range probably wrong
- [x] documentHighlighting does not working
- [x] document-symbol and workspace-symbol not working
- [X] ~refactorings commented~
- [x] inline-symbol refactor
- [x] rename refactor
- [x] completion
- [x] remove all `:file-envs` references
- [X] test sci defined and lint-as macro analysis
- [x] bug with filenames after jumping into jars (`analysis` assoced under filename like `zipfile:/....`)
- [x] remove unused old parse code

### clj-kondo issues/unknowns
- [x] signatures https://github.com/clj-kondo/clj-kondo/pull/1124
- [x] [pr] Need ns/refer/import info from clj-kondo findings to clean-ns https://github.com/clj-kondo/clj-kondo/pull/1132
- [ ] refers/imports after https://github.com/clj-kondo/clj-kondo/issues/1128
- [ ] [pr] keywords https://github.com/clj-kondo/clj-kondo/issues/1129
- [x] [pr] Need `:alias` on var-usages. https://github.com/clj-kondo/clj-kondo/issues/1133

Replaces
https://github.com/clojure-lsp/clojure-lsp/pull/176

Fixes #240


src/clojure_lsp/db.clj
```diff
@@ -11,34 +11,34 @@
 (defonce diagnostics-chan (async/chan 1))
 (defonce edits-chan (async/chan 1))
 
-(def version 5)
+(def version 1)
```
ericdallo, line 14: Why 1 not 6?

src/clojure_lsp/db.clj
```diff
@@ -11,34 +11,34 @@
...
 
 (defn make-spec [project-root]
-  (let [lsp-db (io/file (str project-root) ".lsp" (str "sqlite." version ".db"))]
+  (let [lsp-db (io/file (str project-root) ".lsp" "sqlite.db")]
```
ericdallo, line 17: So we will always use the same version of the db? what's the tradeoff now of that?

src/clojure_lsp/feature/hover.clj
```diff
@@ -19,47 +19,39 @@
...
+        doc-line (when (seq doc)
+                   (if markdown?
+                     (docstring->formatted-markdown doc)
+                     doc))]
+    (if markdown?
+      {:kind "markdown"
+       :value (cond-> (str opening-code sym-line closing-code)
+                signatures-line (str opening-code signatures-line closing-code)
+                doc-line (str line-break opening-code doc-line "\n```")
+                filename (str line-break "*" filename "*"))}
```
ericdallo, line 51: Nice, this is way better this way :)

src/clojure_lsp/feature/hover.clj
```diff
@@ -19,47 +19,39 @@
...
+        signatures-line (when (not show-docs-arity-on-same-line?)
+                          signatures)
+        doc-line (when (seq doc)
+                   (if markdown?
+                     (docstring->formatted-markdown doc)
+                     doc))]
+    (if markdown?
+      {:kind "markdown"
+       :value (cond-> (str opening-code sym-line closing-code)
+                signatures-line (str opening-code signatures-line closing-code)
```
ericdallo, line 49: The hover signatures seem to not work anymore

src/clojure_lsp/db.clj
snoe: The `version` is already considered in selecting from db so there's no reason to leave files hanging around

src/clojure_lsp/db.clj
snoe: changing the filename lets us reset

PR #267, opened 2021-01-24
GraalVM build

ericdallo:
This PR adds `graalvm/native-compile.sh` script to compile clojure-lsp with graalvm

With this PR, the script compiles `clojure-lsp` using a graalvm-ce11-21.0.0
Issues solved:

### To make it work:
- [X] __sqlite__: it needed to declare all reflection classes used by sqlite lib with a [fix](https://github.com/ericdallo/sqlite-jni-graal-fix), I could not make it work using only the `reflection.json`, we could improve that later.
- [x] __log__: I could not make log4j, slf4j, logback or log4j2 to work with graalvm so I changed our lib log to `timbre` which has the same features we use and it's easier to configure, still we need this fix to be merged: https://github.com/ptaoussanis/timbre/pull/334
- [X] __nrepl-server__: graalvm does not work with nrepl-server because of variables [using `future` at top-level](https://github.com/nrepl/nrepl/blob/master/src/clojure/nrepl/util/completion.clj#L170-L180), so I commented it for now, we can use as only dev deps anyway before the merge.
- [X] __dynamic proxy__: I used the graalvm's java with the agent flag to generate these files
- [X] __type-hint__: I needed to change some places to use explicit clojure type-hint since graalvm could not understand what kind of class was that on runtime.

### To allow merge:
- [x] Fix nrepl dep to be local or check for another lib as a replacement.
- [ ] ~Change timbre log for java.util.Logging~ Could not make it work
- [x] Remove all unnecessary classes from reflect.json auto generated.
- [x] setup CI to build for at least Linux
- [x] __test all features__: Since most of errors happen on runtime, we need to test every single feature to make sure no runtime exceptions happen (printed on `/tmp/clojure-lsp.out`)

### Performance:
clojure-lsp can be split into two kind of analysis: startup and after startup since most work is done during the classpath analysis on the project start for the first time, so with my tests this was the ~ result:

Multiple times tested on `NixOS 16 Gb ddr4 RAM, i7-8550U` with a medium/big project of classpath of __250__ files.

| Kind                       | JVM                                | GraalVM              |
|----------------------------|------------------------------------|----------------------|
| Startup TIME*               | ~70s                               | ~91s                 |
| Startup MEM                | max 2.04GB                         | max 1.48GB           |
| Startup CPU                | ~75-85% with max 91%               | ~20-30% with max 43% |
| After startup (cached) CPU | 50% then slowly decreases to ~4-5% | ~1-3%                |
| After startup (cached) MEM | ~736MB                             | ~384MB               |

*Considering kondo classpath scan

Huge thanks to @borkdude which helped me a lot fixing those issues and teaching me ;)


ericdallo:
Merging this to test Linux CI build

eerohele:
>nrepl-server: graalvm does not work with nrepl-server because of variables using future at top-level, so I commented it for now, we can use as only dev deps anyway before the merge.

I think we ought to fix this in nREPL, too. I can look into it when I get the chance.

ericdallo:
It'd be good to fix it on nrepl side, but still for a nrepl server, it'd would not work in runtime since we'd need to evaluate new code after the image was compiled I think

PR #277, opened 2021-01-26
include timestamp and PID in lsp logs

lucywang000:
Since multiple clojure-lsp instance logs into the same `/tmp/clojure-lsp.out` file, it'll be helpful to distinguish what each lsp instance is doing by including the timestamp and PID in each log message.

Before:
```
DEBUG clojure-lsp.main: :codeAction 40 ([:codeAction #{18908224820185}])
```

After:
```
2021/01/26 11:19:30 88291 DEBUG clojure-lsp.main :codeAction 109 ()
```


ericdallo (approved):
LGTM, really useful indeed, thank you for the contribution :)

Issue #268, opened 2021-01-24
Java heap space limit reached


anonimitoraf:
When I start clojure-lsp (in Emacs, clojure-lsp version 2021.01.22-13.04.28) for a Clojurescript project, after a few minutes of attempting to start, I get the following error:

```
java.lang.OutOfMemoryError: Java heap space
	at clojure.lang.Tuple.create(Tuple.java:28)
	at clj_kondo.impl.namespace$get_namespace.invokeStatic(namespace.clj:343)
	at clj_kondo.impl.namespace$get_namespace.invoke(namespace.clj:342)
	at clj_kondo.impl.namespace$resolve_name.invokeStatic(namespace.clj:400)
	at clj_kondo.impl.namespace$resolve_name.invoke(namespace.clj:396)
	at clj_kondo.impl.analyzer$analyze_call.invokeStatic(analyzer.clj:1319)
	at clj_kondo.impl.analyzer$analyze_call.invoke(analyzer.clj:1302)
	at clj_kondo.impl.analyzer$analyze_expression_STAR__STAR_.invokeStatic(analyzer.clj:1781)
	at clj_kondo.impl.analyzer$analyze_expression_STAR__STAR_.invoke(analyzer.clj:1692)
	at clj_kondo.impl.analyzer$analyze_children$fn__17033.invoke(analyzer.clj:62)
	at clojure.core$map$fn__5866.invoke(core.clj:2755)
	at clojure.lang.LazySeq.sval(LazySeq.java:42)
	at clojure.lang.LazySeq.seq(LazySeq.java:51)
	at clojure.lang.Cons.next(Cons.java:39)
	at clojure.lang.RT.boundedLength(RT.java:1792)
	at clojure.lang.RestFn.applyTo(RestFn.java:130)
	at clojure.core$apply.invokeStatic(core.clj:665)
	at clojure.core$mapcat.invokeStatic(core.clj:2783)
	at clojure.core$mapcat.doInvoke(core.clj:2783)
	at clojure.lang.RestFn.invoke(RestFn.java:423)
	at clj_kondo.impl.analyzer$analyze_children.invokeStatic(analyzer.clj:62)
	at clj_kondo.impl.analyzer$analyze_children.invoke(analyzer.clj:41)
	at clj_kondo.impl.analyzer$analyze_call.invokeStatic(analyzer.clj:1582)
	at clj_kondo.impl.analyzer$analyze_call.invoke(analyzer.clj:1302)
[...]

Issue #229, opened 2020-12-31
Reduce memory usage on startup


bpringe:
I want to open this to start/track a conversation about reducing clojure-lsp's memory usage.

I noticed the jvm opts are `:jvm-opts ^:replace ["-Xmx2g" "-server"]`. Is 2g necessary as a max? Potentially if you have n instances running they could use up to `2 x n` gigs of memory. Have lower limits been tried, and what are the results?

If there is an issue with a lower max heap size, it should be possible to make clojure-lsp hold less data in memory at any given time with some tweaks.

Since people are often also running REPLs in their projects, this memory consumption is added on top of the jvm instance for the REPL. I suppose this may be some people's arguments for driving all features through the REPL as opposed to a static analyzer. 

The REPL is already consuming a certain amount of memory, so starting another jvm for features the REPL can offer seems like unnecessary added memory consumption. I'm just throwing an argument against clojure-lsp as it stands now, but obviously I think this is a great project and it can be improved in this regard.

ericdallo:
It's a good question, I know @snoe increased the mem last time from 1 GB -> 2GB but I don't know why or if it was related to clj-kondo.
I think we can try reducing it indeed, I'm not sure clojure-lsp needs that amount of memory just for it's `db` atom.
We can test it reducing to 1GB, WDYT?

ericdallo:
I know that at that time we had a memory leak issue with clj-kondo that was already solved

bpringe:
We'll see if @snoe has some insights, as I'm sure he had a reason for the increase, but I think testing lower limits would be a good start. (

ericdallo:
I found this: https://github.com/clojure-lsp/clojure-lsp/commit/b241b795f45210dcbb2d15c2d04d9cd2616a07d6

bpringe:
I see, so it seems the memory issue was fixed or improved in kondo, but the heap size was still increase in lsp, maybe just to be safe?

On another note, VisualVM may be a good tool for profiling clojure-lsp. I've seen idle instances of clojure-lsp running at 1.7G, which seems way too high considering it's dealing with text files that combined should never reach that size, and also it shouldn't be holding that much in memory anyway.

ericdallo:
Yes, we probably are missing some improvements in performance, specially on this startup function: https://github.com/clojure-lsp/clojure-lsp/blob/master/src/clojure_lsp/crawler.clj#L145-L185

bpringe:
[...]

PR #278, opened 2021-01-26
clj-kondo batch analysis

ericdallo:
Related to #268 and #229 

The idea is to partition clj-kondo calls to decrease memory consumption, this will increase server startup a little though (tested and the startup  time is increased by ~1%-3%).
We may think in a good default batch size (ATM 50), or make it optional via some flag.

Log output:
```
INFO  clojure-lsp.crawler: Analyzing 157 paths with clj-kondo with batch size of 4 ...
INFO  clojure-lsp.crawler: Analyzing 1/4 batch paths with clj-kondo...
INFO  clojure-lsp.crawler: Analyzing 2/4 batch paths with clj-kondo...
INFO  clojure-lsp.crawler: Analyzing 3/4 batch paths with clj-kondo...
INFO  clojure-lsp.crawler: Analyzing 4/4 batch paths with clj-kondo...
INFO  clojure-lsp.crawler: Paths analyzed, took 59.44 secs. Caching for next startups...
```

More details of the improvement here: https://github.com/clojure-lsp/clojure-lsp/issues/268#issuecomment-767518212

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Fix document-symbol after #261 - Fixes #276
+- Reduce memory usage on startup batch analyzing classpath via clj-kondo. - Fixes #268
 
 ## 2021.01.25-22.56.05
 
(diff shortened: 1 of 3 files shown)


ericdallo:
c/c @borkdude

PR #279, opened 2021-01-26
inject pid before first call to log

lucywang000:
A follow-up for #277 

@ericdallo I wanted to push to the original PR but you merged it too fast :)


ericdallo (approved):
haha no problem, both are ok to me

PR #281, opened 2021-01-26
Remove ns from document-symbol, showing only functions/vars

ericdallo:
Showing the ns looks redundant and for some clients like Emacs, it's too much info, so this PR show only the functions/vars from that ns.

### DIFF
--- a/src/clojure_lsp/handlers.clj
+++ b/src/clojure_lsp/handlers.clj
@@ -192,22 +192,14 @@
 
 (defn document-symbol [{:keys [textDocument]}]
   (let [filename (shared/uri->filename textDocument)
-        local-analysis (get-in @db/db [:analysis filename])
-        namespace-definition (q/find-first (comp #{:namespace-definitions} :bucket) local-analysis)]
-    [{:name (or (some-> namespace-definition :name name)
-                filename)
-      :kind (f.document-symbol/element->symbol-kind namespace-definition)
-      :range full-file-range
-      :selection-range (if namespace-definition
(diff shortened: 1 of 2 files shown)


bpringe:
@ericdallo This works, but I'm not sure it's desirable, unless having the ns really causes some trouble in emacs. In the current version Calva uses that gives the ns in the outline, I can click the ns symbol to navigate to the ns form in the editor, which maybe some people would like.

It's not a huge deal I suppose, but it's something nice/convenient that would be taken away.

bpringe:
image

Meant to post this image. I can click `hello` and navigate to the ns form.

ericdallo:
Yeah, I thought about that, but it's kind of redundant IMO, user is already in that ns and can check it by the file, do you think a flag toggling the states would be enough with a default as now showing the ns?

bpringe:
I guess to me, the main point of the outline is quick navigation to symbols by clicking them in the outline rather than scrolling through the file (maybe some people don't like to click, but that seemed to be the main point to me), and the ns symbol is just another symbol that can be navigated to quickly with the outline.

What trouble does having the ns cause in emacs (just curious)?

In any case, if you do want to remove the ns, I think a setting to add it would be a good idea.

ericdallo:
I see, there is no issue at all, it's just the way emacs show the symbol, but I think this could be handled on client side.
 I agree with you, I think we can keep at is it :)

bpringe:
Client side makes sense to me :100:

PR #286, opened 2021-01-28
Add Homebew Github action

Andre0991:
Depends on https://github.com/Homebrew/homebrew-core/pull/69929


ericdallo:
I'm closing this since #321 should take a different approach and solve it with another github action later

PR #301, opened 2021-02-04
Remove install of native-image in native-unix-compile.sh

thiagokokada:
This will allow us to reuse this script in nixpkgs.


ericdallo (approved):
Thank you!

PR #317, opened 2021-02-12
Fix deps.edn run

rvlo:
Somehow the "test" dir is in lein's run classpath.

So, when doing: 
<pre>$ clojure -M:run
Syntax error (FileNotFoundException) compiling at (clojure_lsp/handlers.clj:1:1).
Could not locate clojure_lsp/code_lens__init.class, clojure_lsp/code_lens.clj or clojure_lsp/code_lens.cljc on classpath. Please check that namespaces with dashes use underscores in the Clojure file name.
</pre>


ericdallo (approved):
Oh the code lens file was my bad, nice catch!

Yeah, this happens because we have a custom `ClojureExtensions` class, and this is not a fully deps.edn project yet 😅 , I think we can go with this for now, but we should try to move the whole lein project to a deps one someday, thank you!

PR #325, opened 2021-02-17
Configurable db path with :sqlite-db-path

rvlo:
Main use case is to allow storing sqlite.db somewhere else.


docs/settings.md
```diff
@@ -46,6 +46,7 @@ This is an [example how Emacs `lsp-mode`](https://github.com/emacs-lsp/lsp-mode/
...
 | `project-specs`                 | A vector containing a map of key/value pairs, defining how `clojure-lsp` should find your project classpath                                                                                                                                                                                                                                                                                                          | Check the default [here](https://github.com/clojure-lsp/clojure-lsp/blob/master/src/clojure_lsp/crawler.clj#L53-L60)                  |
+| `sqlite-db-path`                | Where to store the project's analysis cache, used to speed up next `clojure-lsp` startup. A path relative to project root or an absolute path. When not present, defaults to `.lsp/sqlite.db`
```
ericdallo, line 49: Could you move the default value to the right column (last column)?

ericdallo (commented):
LGTM! just the documentation that needs to align, after that we are good to go :)

docs/settings.md
ericdallo:
```suggestion
| `sqlite-db-path`                | Where to store the project's analysis cache, used to speed up next `clojure-lsp` startup. A path relative to project root or an absolute path. | `.lsp/sqlite.db`
```

ericdallo (approved):

PR #338, opened 2021-02-23
fix more windows path issue

kimim:
Windows consider path start with D: or C: as absolute path. 
This PR fix the tests.

### DIFF
--- a/src/clojure_lsp/db.clj
+++ b/src/clojure_lsp/db.clj
@@ -17,7 +17,7 @@
 (defn ^:private get-sqlite-db-file-path [project-root]
   (let [configured (some-> (get-in @db [:settings :sqlite-db-path])
                            io/file)
-        default (io/file (str project-root) ".lsp" "sqlite.db")
+        default (io/file ".lsp" "sqlite.db")
         file (or configured default)]
     (if (.isAbsolute file)
       (.getAbsolutePath file)
--- a/test/clojure_lsp/db_test.clj
+++ b/test/clojure_lsp/db_test.clj
@@ -4,7 +4,8 @@
    [clojure-lsp.test-helper :as h]
    [clojure.string :as s]
    [clojure.test :refer [deftest testing is]]
-   [clojure.java.io :as io]))
+   [clojure.java.io :as io]
+   [clojure-lsp.shared :as shared]))
 
 (h/reset-db-after-test)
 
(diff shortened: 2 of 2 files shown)


src/clojure_lsp/db.clj
```diff
@@ -16,7 +16,7 @@
 (defn ^:private get-sqlite-db-file-path [project-root]
   (let [configured (some-> (get-in @db [:settings :sqlite-db-path])
                            io/file)
-        default (io/file (str project-root) ".lsp" "sqlite.db")
```
ericdallo, line 19: Not sure this should be removed, otherwise how can we make sure the db will be saved on project root?

kimim: project root will be added in following .isAbsolute test.
Otherwise, two project root will added before .lsp in windows, because /user/project/.lsp/sqlite.db is considered as a relative path, thus getAbsolutePath will turn it to: D:/user/project/user/project/.lsp/sqlite.db

kimim: like this:
```
lein test :only clojure-lsp.db-test/sqlite-db-file-setting

FAIL in (sqlite-db-file-setting) (db_test.clj:18)
when not set
expected: "D:\\user\\project\\.lsp\\sqlite.db"
  actual: "D:\\user\\project\\user\\project\\.lsp\\sqlite.db"
    diff: - "D:\\user\\project\\.lsp\\sqlite.db"
          + "D:\\user\\project\\user\\project\\.lsp\\sqlite.db"
```

ericdallo: I tested on linux and it works, I'm just curious how io/file return the path to the project + .lsp/config.edn, how it guesses the project root correctly?

kimim: Yes. That's the problem. The driver in absolute path may change according to project location, under C: or under D:

ericdallo: I see, so this could not be the same for other computes?
I think we should find some other way so, otherwise we can save the db on another path or get a wrong path

kimim: Hey, but we provide `project-path` as the parameter, Windows user should specify correct project root with driver characters.

ericdallo: yeah, use the project-root would be perfect, the current code it's using it

ericdallo: the project-root comes from the client, from emacs, lsp-mode sends the project-root correctly already

ericdallo: So, @kimim can we somehow keep using the project-root?

ericdallo:
Any news here @kimim ?

kimim:
Hello Eric,

There is no update to this PR. Everything works fine, except the unit test cases on my machine.
Shall we close this PR at the moment?

发件人: Eric Dallo<mailto:notifications@github.com>
发送时间: 05 March 2021 08:51
收件人: clojure-lsp/clojure-lsp<mailto:clojure-lsp@noreply.github.com>
抄送: Kimi MA<mailto:kimi.im@outlook.com>; Mention<mailto:mention@noreply.github.com>
主题: Re: [clojure-lsp/clojure-lsp] fix more windows path issue (#338)


Any news here @kimim<https://github.com/kimim> ?

―

ericdallo:
If you intend to fix it, we can keep it open, no problem, otherwise we can close

ericdallo:
@kimim I think this issue was fixed with the improvements we had on URI recently on clojure-lsp, feel free to open another PR if find any issues, thank you!

Issue #234, opened 2021-01-03
Idea - use datalevin for clojure-lsp db


FongHou:
Hi @ericdallo, @snoe 

Just want to bring up an idea. What do you think using https://github.com/juji-io/datalevin for clojure-lsp db persistent.

clojure-lsp currently uses sqlite3 pretty much as a flat file (ie. all data in single row in project table). datalevin seems a good fit for persisting clojure-lsp global db atom. It may enable some scalable and performant optimizations e.g. reduce memory usage since datalevin read query is very fast so the whole db doesn't have to be kept in memory anymore. Also, I remember there was a code analysis tool designed around datomic/datalog long time ago (though I couldn't find it online anymore), but I think it's still a cool idea to pursue long term.

Thanks!

ericdallo:
It looks really cool, but I'm not sure it'd improve performance that much, AFAIK we use the db just for loading already analyzed sessions and it's fast, what it's slow for us, is the fact to analyze all project/jars one time for new projects.
WDYT @snoe ?

snoe:
I've definitely imagined making queries against the usages. But when I started to write it up, it turned out there wasn't a lot of reuse from that ability. Basically there's a "query" for completions and one for "thing under cursor" so code seemed to fit the bill well enough.

There's an argument, I think for a number of indexes on the data.
Does datalevin supports field indexes (let alone indexes from expressions like sql can do)? I'm not sure how it provides the read performance off disk without having the db in memory.

The reason the whole db is not persisted is because there's a lot that can happen between shutting down the server and starting it up again. Doing something like rename on stale or incorrect information is very bad as it could lead to broken code if, say, a commit added a new usage before you start up your editor.

ericdallo:
I made a POC where it's working, I just replace the sqlite to datalevin, but we still use the atom for everything.
https://github.com/clojure-lsp/clojure-lsp/pull/345

Still having some issues while compiling the native-image with GraalVM

ericdallo:
The datalevin worked but discussing with @snoe we think that we'll not have any advantages just changing the sqlite to datalevin for what we use today, we probably can improve it using just a plainfile for the persisted analysis instead of sqlite, then we can probably fix #296 as well.

PR #345, opened 2021-02-27
Use datalevin instead of sqlite

ericdallo:
This is a POC to test datalevin native compiled performance instead of sqlite (don't work with static graalvm flag for now)


ericdallo:
This was a POC and it worked good, but we'll probably use a plainfile instead of sqlite and datalevin. Closes #234

PR #347, opened 2021-02-28
update datalevin version

huahaiy:
Datalevin 0.4.x has some slight changes in function names. Updated. I have tested, with this version of Datalevin, clojure-lsp compiles to native just fine.  

For now, we are including the C code manually and compile it to libdtlv.a.  I will figure out how to include it in the jar instead in the future.


ericdallo (approved):

ericdallo (approved):
Thank you for the help!
Will test it right now

PR #354, opened 2021-03-03
Avoid diagnostic thread crash when some token has no row/col information

lucywang000:
I have a macro that creates new bindings, and I'm writing a custom hook for it so clj-kondo could lint it correctly. Because the macro creates new bindings out of nowhere, the generated token nodes has no row/col information, which made the clojure-lsp diagnostic thread crash in the [`shared/->range`](https://github.com/clojure-lsp/clojure-lsp/blob/master/src/clojure_lsp/shared.clj#L89) function:

```
2021-03-03T07:13:38.979Z dev.local WARN [clojure-lsp.shared:95] - error in ->range, element =  {:end-row nil, :end-col nil, :type :unused-binding, :filename "foo.clj", :message "unused binding request", :row nil, :col nil, :level :warning}atch-8
                                              java.lang.Thread.run              Thread.java:  832
                                                               ...
clojure.core.async.impl.concurrent/counted-thread-factory/reify/fn           concurrent.clj:   29
                java.util.concurrent.ThreadPoolExecutor$Worker.run  ThreadPoolExecutor.java:  630
                 java.util.concurrent.ThreadPoolExecutor.runWorker  ThreadPoolExecutor.java: 1130
                                                               ...
             clojure.core.async.impl.channels.ManyToManyChannel/fn             channels.clj:  135
                       clojure.core.async.impl.ioc-macros/take!/fn           ioc_macros.clj:  991
      clojure.core.async.impl.ioc-macros/run-state-machine-wrapped           ioc_macros.clj:  982
              clojure.core.async.impl.ioc-macros/run-state-machine           ioc_macros.clj:  978
                     clojure-lsp.main/run/fn/state-machine--auto--                 main.clj:  396
                  clojure-lsp.main/run/fn/state-machine--auto--/fn                 main.clj:  396
               clojure-lsp.feature.file-management/analyze-changes      file_management.clj:  114
                            clojure-lsp.feature.diagnostics/notify          diagnostics.clj:   26
                                                 clojure.core/mapv                 core.clj: 6908
                                               clojure.core/reduce                 core.clj: 6833
                                       clojure.core.protocols/fn/G            protocols.clj:   13
                                         clojure.core.protocols/fn            protocols.clj:   75
                                 clojure.core.protocols/seq-reduce            protocols.clj:   31
                                       clojure.core.protocols/fn/G            protocols.clj:   19
                                         clojure.core.protocols/fn            protocols.clj:  136
                                                               ...
                                              clojure.core/mapv/fn                 core.clj: 6917
         clojure-lsp.feature.diagnostics/kondo-finding->diagnostic          diagnostics.clj:   12
                                        clojure-lsp.shared/->range               shared.clj:   92
                                                               ...
java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "x" is null
```

### DIFF
--- a/src/clojure_lsp/shared.clj
+++ b/src/clojure_lsp/shared.clj
@@ -87,12 +87,16 @@
   (string/replace uri project-root ""))
 
 (defn ->range [{:keys [name-row name-end-row name-col name-end-col row end-row col end-col] :as element}]
-  (when element
-    {:start {:line (max 0 (dec (or name-row row))) :character (max 0 (dec (or name-col col)))}
-     :end {:line (max 0 (dec (or name-end-row end-row))) :character (max 0 (dec (or name-end-col end-col)))}}))
+  (try
+    (when (and element (or name-row row))
+      {:start {:line (max 0 (dec (or name-row row))) :character (max 0 (dec (or name-col col)))}
+       :end {:line (max 0 (dec (or name-end-row end-row))) :character (max 0 (dec (or name-end-col end-col)))}})
+    (catch Exception e
+      (log/warn "error in ->range, element = " (pr-str element))
+      (throw e))))
 
 (defn ->scope-range [{:keys [name-row name-end-row name-col name-end-col row end-row col end-col] :as element}]
-  (when element
+  (when (and element (or row name-row))
     {:start {:line (max 0 (dec (or row name-row))) :character (max 0 (dec (or col name-col)))}
      :end {:line (max 0 (dec (or end-row name-end-row))) :character (max 0 (dec (or end-col name-end-col)))}}))
 


lucywang000:
Just found that for some reason another part is throwing with the same error:

```
                                              java.lang.Thread.run                        Thread.java:  832
                java.util.concurrent.ThreadPoolExecutor$Worker.run            ThreadPoolExecutor.java:  630
                 java.util.concurrent.ThreadPoolExecutor.runWorker            ThreadPoolExecutor.java: 1130
                               java.util.concurrent.FutureTask.run                    FutureTask.java:  264
               java.util.concurrent.Executors$RunnableAdapter.call                     Executors.java:  515
     org.eclipse.lsp4j.jsonrpc.json.ConcurrentMessageProcessor.run    ConcurrentMessageProcessor.java:  113
       org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.listen         StreamMessageProducer.java:   94
org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.handleMessage         StreamMessageProducer.java:  194
                  org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.consume                RemoteEndpoint.java:  187
       org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.handleNotification                RemoteEndpoint.java:  220
         org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.notify               GenericEndpoint.java:  152
  org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.lambda$null$0               GenericEndpoint.java:   65
                                                               ...
          jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke  DelegatingMethodAccessorImpl.java:   43
              jdk.internal.reflect.GeneratedMethodAccessor5.invoke
                 clojure-lsp.main.LSPTextDocumentService/didChange                           main.clj:  118
                                   clojure-lsp.handlers/did-change                       handlers.clj:   58
                    clojure-lsp.feature.file-management/did-change                file_management.clj:  122
                                               clojure.core/reduce                           core.clj: 6833
                                       clojure.core.protocols/fn/G                      protocols.clj:   13
                                         clojure.core.protocols/fn                      protocols.clj:   75
                                 clojure.core.protocols/seq-reduce                      protocols.clj:   31
                                       clojure.core.protocols/fn/G                      protocols.clj:   19
                                         clojure.core.protocols/fn                      protocols.clj:  136
                                                               ...
                 clojure-lsp.feature.file-management/did-change/fn                file_management.clj:  126
                  clojure-lsp.feature.file-management/replace-text                file_management.clj:   98
                       clojure-lsp.feature.file-management/offsets                file_management.clj:   82
                                                               ...
java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "x" is null
```

ericdallo (changes requested):
I think this only avoid a exception that can happens in a lot of places, we need to fix the root issue, do you have a small sample where this happens?
We need to understand why the analysis get empty row/col

ericdallo:
@lucywang000 when we call clj-kondo to analyze the code, we only [keep valid elements](https://github.com/clojure-lsp/clojure-lsp/blob/master/src/clojure_lsp/crawler.clj#L139-L142), we need to understand why this element doesn't have the `:row` and `:col`, we heavily rely on row and col for almost every feature, we could ignore that element but features wouldn't work for that, maybe we could fix on the kondo side?

lucywang000:
Hey @ericdallo , here is a minimal repro: https://github.com/lucywang000/bugs-repro/blob/master/repro-clojure-lsp/src/repro.clj

clj-kondo config: https://github.com/lucywang000/bugs-repro/blob/master/repro-clojure-lsp/.clj-kondo/config.edn
custom hook code: https://github.com/lucywang000/bugs-repro/blob/master/repro-clojure-lsp/.clj-kondo/hooks/custom.clj

ericdallo:
I'll take a look soon, thank you!

ericdallo:
So, first we scan the whole classpath:
```bash
$ clj -Sdeps '{:deps {clj-kondo {:mvn/version "2021.03.03"}}}' -m clj-kondo.main  --lint $(lein classpath)
```
Then we lint that specific file and get the findings to check what is returned by clj-kondo to clojure-lsp:
```bash
$ clj -Sdeps '{:deps {clj-kondo {:mvn/version "2021.03.03"}}}' -m clj-kondo.main  --lint src/repro.clj --config '{:output {:format :json}}' | jq .

{
  "findings": [
    {
      "type": "unused-binding",
      "filename": "src/repro.clj",
      "message": "unused binding foo",
      "row": null,
      "col": null,
      "end-row": null,
      "end-col": null,
      "level": "warning"
    },
    {
      "type": "unresolved-symbol",
      "filename": "src/repro.clj",
      "message": "Unresolved symbol: foo1",
      "row": 15,
      "col": 7,
      "end-row": 15,
      "end-col": 11,
      "level": "error"
    }
  ],
  "summary": {
    "error": 1,
    "warning": 1,
    "info": 0,
    "type": "summary",
    "duration": 363,
    "files": 1
  }
}

```
We can see that the `unused binding foo` has no row/col, this way clojure-lsp, we could fix a row/col but it'd be better to make clj-kondo return that correctly.
WDYT @borkdude?

snoe:
1) anaphoric macros are awful and should be avoided if you can.
2) I'll argue this warning is incorrect since it has no positioning. If a finding occurs in a hook without positioning, I think kondo should position it at the macro call (~(10,2) in this case).
```   
   ;; `clj-kondo --lint` prints (correctly):
   ;;   repro-clojure-lsp/src/repro.clj::: warning: unused binding foo
```
3) Same as analysis, we should be throwing out unpositioned `findings` and warning in the log about them as a heads up to fix in kondo. 
(@ericdallo I missed this, but I believe https://github.com/clojure-lsp/clojure-lsp/commit/8b9d1bd7c89ad044138232d7414ddb386571a707 should have also fixed the issue of leaking analysis in kondo instead of hiding the error in here).
4) there might be situations that are no-fix in kondo but let's make sure we have that conversation before addressing it here.

borkdude:
I agree with @snoe: log them and fix them in kondo but don't show to the user or at least not crash :).

ericdallo:
Yes, I agree, that PR of mine just fixed the analysis, this is a finding issue.
So, we can log and not consider that finding until we fix on kondo side

ericdallo:
Fixed the crash on f04d058 @lucywang000, we still need to fix on kondo side I imagine to clj-kondo return the position correctly, should we open a clj-kondo issue for that @borkdude ?

borkdude:
Yes please.

ericdallo:
https://github.com/clj-kondo/clj-kondo/issues/1193

lucywang000:
Thanks @ericdallo !

PR #357, opened 2021-03-04
Fix issue where a full text replacement was not handled

orestis:
I couldn't find any test for handle-change so I decided to open a PR without a test first, so that we can at least verify that it's wanted. I've tested this with Neovim but I couldn't get it to send any changes with ranges, so some manual verification needs to happen until a test is added...

### DIFF
--- a/src/clojure_lsp/feature/file_management.clj
+++ b/src/clojure_lsp/feature/file_management.clj
@@ -104,7 +104,6 @@
   "Handle a TextDocumentContentChangeEvent"
   [old-text change]
   (let [new-text (:text change)]
-    ;; there's also rangeLength but it's deprecated
     (if-let [r (:range change)] 
       (let [{{start-line :line
               start-character :character} :start
@@ -114,7 +113,6 @@
       ;; If range and rangeLength are omitted the new text is considered to be
       ;; the full content of the document.
       new-text)))
-
 
 (defn analyze-changes [{:keys [uri text version]}]
   (let [notify-references? (get-in @db/db [:settings :notify-references-on-file-change] false)]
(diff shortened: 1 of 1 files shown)


src/clojure_lsp/feature/file_management.clj
```diff
@@ -119,11 +135,9 @@
 (defn did-change [uri changes version]
   (loop [state-db @db/db]
     (let [old-text (get-in state-db [:documents uri :text])
-          final-text (reduce (fn [old-text {new-text :text {{start-line :line
-                                                             start-character :character} :start
-                                                            {end-line :line
-                                                             end-character :character} :end} :range}]
-                               (replace-text old-text new-text start-line start-character end-line end-character)) old-text changes)]
+          final-text (reduce (fn [old-text change]
+                               (handle-change old-text change)) old-text changes)]
```
ericdallo, line 139:
```suggestion
          final-text (reduce handle-change old-text changes)]
```

src/clojure_lsp/feature/file_management.clj
```diff
@@ -100,6 +100,22 @@
                          (subs original end-offset (count original))]]
     (string/join [prefix replacement suffix])))
 
+(defn ^:private handle-change 
+  "Handle a TextDocumentContentChangeEvent"
+  [old-text change]
+  (let [new-text (:text change)]
+    ;; there's also rangeLength but it's deprecated
```
ericdallo, line 107:
```suggestion
```

src/clojure_lsp/feature/file_management.clj
```diff
@@ -100,6 +100,22 @@
...
+      (let [{{start-line :line
+              start-character :character} :start
+             {end-line :line
+              end-character :character} :end} r]
+        (replace-text old-text new-text start-line start-character end-line end-character))
+      ;; If range and rangeLength are omitted the new text is considered to be
+      ;; the full content of the document.
+      new-text)))
+
+
```
ericdallo, line 118:
```suggestion
```

ericdallo (approved):
Added some suggestions but looks good, a test for that would be handy indeed, but don't need to add on this PR.

ericdallo (approved):
Tested and it works, thank you for your contribution!

PR #383, opened 2021-03-24
Fix keyword rename

rvlo:
Renaming keywords sometimes go terribly wrong. And it is very useful to be able to reliably do so when using re-frame or namespaced map keys (think datomic, crux or any other database-like use).

The tests show exactly how.


ericdallo:
Thank you!
I was not aware of the issue, I'll take a look soon today

snoe:
It's not obvious from the tests, but one thing that should not be possible is to rename a keyword from a destructuring binding.

Renaming with cursor at `|`:
`:x (let [{:keys [:x|]} foo] x)` => `(:x (let [{:keys [:y]} y)`

So renaming in this situation is an operation on the **local var** not on the keyword that shares the space.

rvlo:
@snoe, that test case is already implemented here:
https://github.com/clojure-lsp/clojure-lsp/blob/d694221b122d0a34d1e404425d52379b0ceddf2c/test/clojure_lsp/handlers_test.clj#L271-L287

In the case you mentioned, the keyword is actually in the `:locals` bucket. The `rename-changes` function acts as a dispatcher:
https://github.com/clojure-lsp/clojure-lsp/blob/d694221b122d0a34d1e404425d52379b0ceddf2c/src/clojure_lsp/feature/rename.clj#L92-L105

And even if it wasn't, even if in the `keywords` bucket, the `rename-keyword` function only handles namespaced keywords. As restricted here:
https://github.com/clojure-lsp/clojure-lsp/blob/d694221b122d0a34d1e404425d52379b0ceddf2c/src/clojure_lsp/feature/rename.clj#L107-L113

Please take these comments as me "speaking out loud" my understandings so somebody could correct me if I got something wrong.

ericdallo:
Just to add to here, I added rename I terrarium tests to master yesterday, they may help us to understand the behavior or fix it :)

ericdallo (approved):
Looks really good, thank you for the contribution! Especially the integration tests ;)

PR #395, opened 2021-04-05
Update installer version parsing

rnadler:
A line was added to the ```--version``` output in https://github.com/clojure-lsp/clojure-lsp/commit/68ea1eb0f790502a74b80711d8e6280856defa5e which broke the installer script. 

This is a simple change to resolve the issue.


ericdallo (approved):
Thanks for the contribution!

PR #397, opened 2021-04-06
Fix for duplicate edits in cljc keyword renaming

rvlo:
Renaming ::click to ::click-now was resulting in ::click-now-now in cljc
files because clj-kondo analysis includes duplicate entries, one with
{:lang :clj} and another with {:lang :cljs}


ericdallo:
Nice, would you mind creating an integration test for that case as well?

rvlo:
> Nice, would you mind creating an integration test for that case as well?

Not at all. Just changed clj files to cljc. Without the fix, we get duplicate edits. With the fix, it works as expected.

ericdallo (approved):
LGTM, @rvlo to merge, could you add an entry in `CHANGELOG.md`?

rvlo:
Like this?

Issue #399, opened 2021-04-07
code formatting stops certain paredit actions


benedekfazekas:
**Describe the bug**

`paredit-forward-slurp-sexp` not working properly when clojure-lsp is enabled on a clojure buffer. it works for the following scenario:
`(foo|) bar` -> `(foo bar)`
but it fails for
`(foo|) (bar)`
nothing happens

for the first, working scenario in the *Messages* buffer I can see
`LSP :: No formatting changes provided`

for the second, failing scenario in the *Messages* buffer:
```
LSP :: Applying 1 edits to `shell.clj' ...
Applying 1 edits to `shell.clj' ...done
```

killing the clojure-lsp server manually fixes the problem, also setting `(setq lsp-enable-indentation nil)` seems to be a workaround

have not tested other paredit actions but others may be affected as well.

**User details (please complete the following information):**
 - OS: MacOS
 - Editor emacs 27.1
 - Version: (post the result of `clojure-lsp --version`)

```
clojure-lsp 2021.03.30-20.42.34
clj-kondo 2021.03.22-SNAPSHOT
```

let me know if you need more info to reproduce, thanks for the amazing clojure-lsp!

ericdallo:
Yeah, I'm having the same issue recently, not sure it's a lsp-mode or clojure-lsp issue yet,
Could you test if it happens in older clojure-lsp versions?

benedekfazekas:
how older should i downgrade to?

benedekfazekas:
my lsp mode is `lsp-mode-20210407.736` just upgraded packages

ericdallo:
Maybe test 2 or 3 clojure-lsp previous releases, this helps finding the release that introduced the bug

benedekfazekas:
will try, i will get back to you if I find a working version

ericdallo:
I'm trying to repro right now and don't get the issue anymore... It seems to be some kind of corner case.
Also, I think this is related to #127

ericdallo:
I confirmed the issue started happens on https://github.com/clojure-lsp/clojure-lsp/releases/tag/2021.03.26-23.41.07, not sure why yet

ericdallo:
I confirmed the bug was introduced with [this `refactoring-async` merge](https://github.com/clojure-lsp/clojure-lsp/commit/b0d872357f3bb3681f2a324e2734a672928775c8), git reverting this commit fix the issue, @snoe could you help identify the issue on that merge please?

[...]

PR #401, opened 2021-04-07
Fix async-refactor branch

ericdallo:
This branch caused a regression on formatting for some reason: #399 and #396


ericdallo:
Merging the revert as temporary fix until we find the root cause

PR #409, opened 2021-04-24
Bump clj-flx with VERY significant performance improvements for workspace symbol filtering

anonimitoraf:
I'm sorry I procrastinated benchmarking.
I found out that my initial implementation was very slow.

For example, calculating the score of candidate `quite a very long candidate string` against search string `quite a long string` would take 2000ms.

I've now fixed this. Now the same calculation takes only 3ms. (Almost 1000x faster). [Relevant commit](https://github.com/anonimitoraf/clj-flx/commit/61b2081b65b7d3be14851bac03ea508147c90054)

More info in https://github.com/anonimitoraf/clj-flx#benchmarks.

I've not changed any tests and they all still pass :D (both in clojure-lsp and clj-flx)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


ericdallo (approved):
Seems good, would you mind add a CHANGELOG entry as well?

PR #410, opened 2021-04-26
Render brew class with template. Add Linux support

wdullaer:
The main goal of this PR is to add Linux support to the homebrew tap.

The current sed implemenation doesn't allow you to template 2 different SHA files.

<strike>The action has been updated to run a small go program that renders the ruby class as a template, which is much more robust.
The mac-latest virtual environment includes a go runtime by default: https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md#language-and-runtime</strike>

The corresponding changes to the brew repo are done in this PR (which should be merged first): https://github.com/clojure-lsp/homebrew-brew/pull/2

edit: I rewrote the go script as a babashka script.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -419,13 +419,19 @@ jobs:
         env:
           VERSION: ${{ steps.latest-tag.outputs.tag }}
         run: |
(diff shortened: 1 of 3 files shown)


ericdallo:
@wdullaer Would you mind add an entry to the `CHANGELOG.md` mentioning the support for Linux via brew?

ericdallo (approved):
Looks really good, thank you for your contribution!

PR #424, opened 2021-05-14
Process most of async client request after processing changes

ericdallo:
didChange LSP method is a critical piece of the server, it should process the changes of the client buffer along with process new diagnostics/analysis for other requests use updated analysis.

There was a issue introduced with async didChange processing some time ago where since we started to handling didChange completely async, some client requests done while the didChange was not totally complete were using outdated analysis of code, like semantic tokens, code lenses, document highlight.

So I introduced on 857bdc186e5585771c3b50b2c444065fb2b2e018 the sync didChange and fixed the out of sync for those features, but that made the didChange request wait for clj-kondo analysis which may take a while for large buffers (~28s on clojure.core buffer, ~7000 lines).

This PR rollback to handle async the code processing but adds a new `:processing-changes` to the db, and make features that highly rely on updated analysis like semantic-tokens, code lenses, document highlight to await the changes processing to then process those features.

This was the best approach I found and it worked very well even on clojure.core buffers, making didChange async/quickly and safely processing features with updated content (async as they were already).

### DIFF
--- a/src/clojure_lsp/handlers.clj
+++ b/src/clojure_lsp/handlers.clj
@@ -36,6 +36,15 @@
 (def ^:private full-file-range
   (shared/->range {:row 1 :col 1 :end-row 1000000 :end-col 1000000}))
 
+(defmacro process-after-changes [& body]
+  `(let [~'_time (System/nanoTime)]
+     (loop []
+       (if (> (quot (- (System/nanoTime) ~'_time) 1000000) 60000) ; one minute timeout
+         (log/warn "Timeout waiting for changes for body")
+         (if (:processing-changes @db/db)
+           (recur)
+           ~@body)))))
+
 (defn initialize [project-root client-capabilities client-settings]
   (when project-root
     (crawler/initialize-project project-root client-capabilities client-settings)))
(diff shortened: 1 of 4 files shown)


src/clojure_lsp/handlers.clj
```diff
@@ -36,6 +36,15 @@
 (def ^:private full-file-range
   (shared/->range {:row 1 :col 1 :end-row 1000000 :end-col 1000000}))
 
+(defmacro process-after-changes [& body]
+  `(let [~'_time (System/nanoTime)]
+     (loop []
+       (if (> (quot (- (System/nanoTime) ~'_time) 1000000) 60000) ; one minute timeout
```
ericdallo, line 42: I could not repro that even with clojure.core buffer, but I imagine this could happen for crazy long buffers that analysis could take more than one minute, maybe we should use even higher value here?

jcsims:
@ericdallo This made a huge difference for me so far in a 5000+ line Clojure file. I checked out this PR, made a local build with `make`, and set `lsp-clojure-custom-server-command` to this newly built `clojure-lsp`, and it's back to buttery smooth. I'll use this more tomorrow and see if there any issues I can surface with it. Thanks for this!

snoe:
I think this is fine to fix the immediate problem. However, the proliferation of core.async throughout has been worrying me deeply, and a fix that adds more indirection maybe should be considered temporary.

The performance problem still exists and needs to be addressed. We definitely tested kondo on clojure.core and it was way way faster than 28s initially, so something has slowed down in clojure-lsp. 

Running on clojure/core.clj `time clj-kondo --parallel --lint "core.clj" --config "{:output {:analysis {:arglists true :locals true :keywords true} :format :edn}}"` I get 1.3s so what is happing to balloon that 20x?

I suspect it's to do with diagnosing and analyzing other files on change, is that still happening?

The process-changes channel dropped timing info, this is critical to know when things are slower than they should be so operations need to print the full timing in the logs even if they get shuffled off to other threads/channels. There also used to be a log that showed how many operations were pending in main.clj, I believe that's been removed but it should probably go back too.

Performance tests should also be added to avoid this in the future - I think a target of 5s for didChange to fully update the client with diagnostics and hover info against clojure.core is more than reasonable.

ericdallo:
It makes sense @snoe, I totally agree.
We are not analysing other files on changes anymore, but maybe there is something slow on diagnostics.

I'll merge as a temp fix and improve the logs to add log to kondo runs via a optional debug flag, also not sure how to log the time when things go async

snoe:
@ericdallo I think you basically need to make ids when a request comes in an wrap the data on the channel with that id and start-time. maybe something like `{:id (str "didChange" (gensym)) :start 8121... :data data-for-channel}` but there's a number of other ways, like using an atom to keep track of in-flight threads.

ericdallo:
It makes sense @snoe, maybe the id one could be the easiest

PR #435, opened 2021-05-25
Improve performance for didOpen method

ericdallo:
We have some performance issues on processing diagnostics:
- During [kondo-finding->diagnostic](https://github.com/clojure-lsp/clojure-lsp/blob/master/src/clojure_lsp/feature/diagnostics.clj#L24), we call `(parser/loc-at-pos text row col)` for every diagnostic, which could spend ~50ms for every diagnostics on a buffer.
- [lint-public-vars](https://github.com/clojure-lsp/clojure-lsp/blob/master/src/clojure_lsp/feature/diagnostics.clj#L88) usually is fast for most projects/buffers as it depends on the number of public vars on the buffer, but for `clojure.core` it takes ~13s as it contains a lot of public vars, but I can't see a lot of projects with that quantity of public vars, and for those cases, user could disable the public linter easily for performance improvements.
- didOpen is a sync request, and it looks like that some editors wait for it to process other requests like `lsp-mode`, the issue here is that we lint the file sync during didOpen.

What was done to improve it:

- For now, I removed/rolled back the logic that was checking for expression using rewirte-clj as it was taking too much time (40s -> 13s on clojure.core), also I didn't notice (yet) a huge change on user XP on how the range of diagnostics are affected.
- The major change here was the change of didOpen lint to use an async lint version as even if we are taking too much time to process the diagnostics (ATM only for public vars), there is no reason to process sync the diagnostics during didOpen
- I changed some other places to use async-lint as well as they are places that should be async as well.

Also, I intend to create a performance section explaining the public var lint performance for huge buffers with a lot of public vars and how to disable it.


snoe:
good find, we absolutely should not be parsing the text ourselves ever - rely fully on kondo or cljfmt and push prs if need be.

ericdallo:
@snoe we were parsing only for checking if the diagnostic was starting with a `(`, not sure how to delegate that to upstream

PR #436, opened 2021-05-26
Fix a bug in `feature.file-management/did-open`

vmfhrmfoaj:
The bug is that wrongly get `:auto-add-ns-to-new-files?` option.


ericdallo (approved):
Oh, sorry about that and thanks for the fix!

PR #437, opened 2021-05-26
Fix uri->filename not fully decoding URI

LuisThiamNye:
I found that clojure-lsp was not loading my project's .lsp/config.edn file. This was because the project's root path contained whitespace, and so the `%20` in the URI was not correctly being decoded into the correct path by the `uri->filename` function.

### DIFF
--- a/src/clojure_lsp/shared.clj
+++ b/src/clojure_lsp/shared.clj
@@ -77,7 +77,7 @@
          jar-file
          ":"
          nested-file)
-    (str (string/replace uri #"^[a-z]+://" ""))))
+    (.getPath (URI. uri))))
 
 (defn filename->uri [^String filename]
   (let [jar-scheme? (= "jar" (get-in @db/db [:settings :dependency-scheme]))]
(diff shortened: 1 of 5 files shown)


ericdallo:
@LuisThiamNye could you check the tests failling?

LuisThiamNye:
The tests use "file://c.clj", which is causing those problems. Should this not be considered an invalid file URI scheme?

ericdallo:
@LuisThiamNye, I think the correct one is `file:///c.clj`, most tests use this one and I know this works with real paths

LuisThiamNye:
As I am assuming something like `file://c.clj` would never need to be processed, I updated the tests.

I also added an additional test for the original change.

ericdallo (approved):
LGTM! 

@bpringe, maybe this should fix that issue with some windows users?

bpringe:
@ericdallo You mean this one? https://github.com/BetterThanTomorrow/calva/issues/1050

PR #439, opened 2021-05-30
Fix inconsistent URI encoding and path representation

LuisThiamNye:
This is a follow-up to #437, which I found was not a full solution to its problem. In fact, it might have introduced more problems on the native image due to reflection and a lack of type hints.

## Problems:

- Both encoded and 'decoded' URIs are being passed around. The URIs supplied to the server are being decoded by `interop/document->decoded-uri` and `handlers/did-open` before they reach `uri->filename`, which causes an exception when constructing a `java.net.URI` and prevents features from working.
    - E.g. in the handlers, `file:///some%20path` gets converted to `file:///some path` (the latter is no longer a valid URI)
    - **Proposed solution:**
        - Do not decode URIs unless converting them to a file path (relates to assumption 1)
- `:source-paths` in `crawler/initialize-project` are set using `.getAbsolutePath` (uses native path representation) which on Windows may be inconsistent with the unix-style output of `shared/uri->filename`
    - **Proposed solution:**
        - Ensure all filenames are correct absolute paths in the representation native to the current platform (assumption 5)

## Assumptions:

1. URIs should have a valid format:
    - Windows file URIs should contain a drive letter (or an authority component)
    - URIs should be properly %-encoded (ie no spaces)
    - URIs should always be absolute
    - `"jar:file:////home/some/.m2/some-jar.jar!/clojure/core.clj"` with four slashes is an invalid URI unless this is a UNC path where 'home' is a server
    - `URLDecoder` should not be used to decode URIs.
2. All URIs processed by the server are encoded the same way
    - i.e. the program will never see both `"zipfile:///something.jar%3A%3Asomething/file.cljc"` and `"zipfile:///something.jar::something/file.cljc"`; Windows drive letters will always be either lower case or upper case.
    - This is because URIs are used for uniquely identifying files.
3. `Paths.get().toAbsolutePath()` is the most reliable way of converting a URI into a unique, absolute file path.
    - Assumes (1)
4. `Paths/get` cannot be used for non-`file:` URIs
5. "filename"s are absolute paths in the representation matching the current platform (eg `C:\path` for Windows and `/path` for Unix)
6. Jar filenames are in the format: `C:\canonical\native\path.jar:entry/name.clj` — the entry path always uses `/` delimiters.
    - Assumes:
        - This is also how clj-kondo formats these paths
        - clojure-lsp only obtains its jar filenames from clj-kondo
        - (5)
7. On Windows devices that run the unit tests: the result of `(-> "/" io/file .getCanonicalPath)` is `"C:\\"`

## Summary of Changes

- Assuming (1)
    - Modify test: `shared-test/filename→uri` "when it is a jar via jarfile" — use three slashes for the URI as 'home' is probably not a server in `"jar:file:////home/some/.m2/some-jar.jar!/clojure/core.clj"`
    - For `shared/filename->uri`, add tests for URIs containing Windows drive letters
    - Use correct URI encoding for `integration-test.integration.helper/file->uri`
    - Use valid `jar:file` scheme for jar file URIs in `queries-test`
    - Use `file-uri` helper to ensure URIs used in tests contain a drive letter (`C:`)on Windows.
    - Replace `document->decoded-uri` with `document->uri` (which does not decode the URI) and remove tests for the decoding.
- Assuming (1), (3), (4), (5), (6)
    - Modify `shared/uri->filename` to handle `jar:file` and `zipfile` and to output a native, absolute file path.
- Assuming (1), (5), (6)
    - Modify `shared/filename->uri` to use correct URI encoding. Assume the function only receives absolute paths, and remove the test for relative path inputs.
- Assuming (5)
    - Use `file-path` helper to ensure file paths used in tests use the Windows path representation (with backslashes and drive letter `C:`)
    - In `file-management/uri->namespace`, use the correct file path separator and a new `shared/relativize-filepath` function to ensure the namespace is also obtained on Windows.
    - In `feature.rename/namespace->uri`, use new `shared/join-filepaths` function to ensure a correct URI for the current platform.
- Replace `shared/project-related-path` with the more general `shared/relativize-filepath` — Note: the output no longer has a leading slash (as it is a relative path).
- Given that in most parts of the code, `project-root` is a URI
    - In `db.clj`, replace `project-root` with `project-root-path` to avoid confusion.
    - Modify tests for `config/resolve-config` to pass URIs for the `project-root` instead of file paths.
- Use type hints to avoid reflection which can cause exceptions with the GraalVM native image.

## Further questions

- Should `project-root` be renamed to `project-root-uri` or `project-uri`?
- Should `filename` be renamed to `filepath`? (To me, 'name' implies solely the name of the file without the directory information)

### DIFF
--- a/src/clojure_lsp/crawler.clj
+++ b/src/clojure_lsp/crawler.clj
@@ -14,7 +14,9 @@
     [clojure.set :as set]
     [clojure.string :as string]
     [digest :as digest]
-    [taoensso.timbre :as log]))
+    [taoensso.timbre :as log])
+  (:import
+   (java.net URI)))
 
 (defn ^:private to-file ^java.io.File
   [^java.nio.file.Path path
@@ -217,10 +219,16 @@
     (log/info "Analyzing source paths for project root" root-path)
     (analyze-paths! source-paths false)))
 
-(defn initialize-project [project-root-uri client-capabilities]
+(defn initialize-project [project-root-uri client-capabilities client-settings]
   (let [project-settings (config/resolve-config project-root-uri)
         root-path (shared/uri->path project-root-uri)
-        settings (-> (merge (:settings @db/db) project-settings)
+        default-settings {:uri-format {:upper-case-drive-letter? (->> project-root-uri URI. .getPath
+                                                                      (re-find #"^/[A-Z]:/")
+                                                                      boolean)
+                                       :encode-colons-in-path? (string/includes? project-root-uri "%3A")}}
+        settings (-> (merge default-settings
+                            client-settings
+                            project-settings)
                      (update :source-paths (fn [source-paths] (mapv #(str (.getAbsolutePath (to-file root-path %))) source-paths)))
                      (update :cljfmt cljfmt.main/merge-default-options))]
     (when-let [log-path (:log-path settings)]
@@ -228,6 +236,7 @@
     (swap! db/db assoc
            :project-root-uri project-root-uri
            :project-settings project-settings
+           :client-settings client-settings
            :settings settings
            :client-capabilities client-capabilities)
     (analyze-project! project-root-uri)))
--- a/src/clojure_lsp/crawler.clj
+++ b/src/clojure_lsp/crawler.clj
@@ -217,10 +217,10 @@
     (log/info "Analyzing source paths for project root" root-path)
     (analyze-paths! source-paths false)))
 
-(defn initialize-project [project-root-uri client-capabilities client-settings]
+(defn initialize-project [project-root-uri client-capabilities]
   (let [project-settings (config/resolve-config project-root-uri)
         root-path (shared/uri->path project-root-uri)
-        settings (-> (merge client-settings project-settings)
+        settings (-> (merge (:settings @db/db) project-settings)
                      (update :source-paths (fn [source-paths] (mapv #(str (.getAbsolutePath (to-file root-path %))) source-paths)))
                      (update :cljfmt cljfmt.main/merge-default-options))]
     (when-let [log-path (:log-path settings)]
@@ -228,7 +228,6 @@
     (swap! db/db assoc
            :project-root-uri project-root-uri
            :project-settings project-settings
-           :client-settings client-settings
            :settings settings
            :client-capabilities client-capabilities)
     (analyze-project! project-root-uri)))
--- a/src/clojure_lsp/crawler.clj
+++ b/src/clojure_lsp/crawler.clj
@@ -209,26 +209,26 @@
           (log/info "Manual GC after classpath scan took" (float (/ (- (System/nanoTime) start-time) 1000000000)) "seconds")
           (db/save-deps root-path project-hash classpath analysis))))))
 
-(defn ^:private analyze-project! [project-root]
-  (let [root-path (shared/uri->path project-root)
+(defn ^:private analyze-project! [project-root-uri]
+  (let [root-path (shared/uri->path project-root-uri)
         settings (:settings @db/db)
         source-paths (get settings :source-paths)]
     (analyze-classpath! root-path source-paths settings)
     (log/info "Analyzing source paths for project root" root-path)
     (analyze-paths! source-paths false)))
 
-(defn initialize-project [project-root client-capabilities client-settings]
-  (let [project-settings (config/resolve-config project-root)
-        root-path (shared/uri->path project-root)
+(defn initialize-project [project-root-uri client-capabilities client-settings]
+  (let [project-settings (config/resolve-config project-root-uri)
+        root-path (shared/uri->path project-root-uri)
         settings (-> (merge client-settings project-settings)
                      (update :source-paths (fn [source-paths] (mapv #(str (.getAbsolutePath (to-file root-path %))) source-paths)))
                      (update :cljfmt cljfmt.main/merge-default-options))]
     (when-let [log-path (:log-path settings)]
       (logging/update-log-path log-path))
     (swap! db/db assoc
-           :project-root project-root
+           :project-root-uri project-root-uri
            :project-settings project-settings
            :client-settings client-settings
            :settings settings
            :client-capabilities client-capabilities)
-    (analyze-project! project-root)))
+    (analyze-project! project-root-uri)))
--- a/src/clojure_lsp/handlers.clj
+++ b/src/clojure_lsp/handlers.clj
@@ -31,7 +31,6 @@
   (:import
     [java.net
      URL
-     URI
      JarURLConnection]))
 
 (def ^:private full-file-range
@@ -48,16 +47,7 @@
 
 (defn initialize [project-root-uri client-capabilities client-settings]
   (when project-root-uri
-    (swap! db/db assoc
-           :client-settings client-settings
-           :settings (cond-> client-settings
-                       (not (map? (:uri-format client-settings)))
-                       (assoc :uri-format
-                              {:upper-case-drive-letter? (->> project-root-uri URI. .getPath
-                                                              (re-find #"^/[A-Z]:/")
-                                                              boolean)
-                               :encode-colons-in-path? (string/includes? project-root-uri "%3A")})))
-    (crawler/initialize-project project-root-uri client-capabilities)))
+    (crawler/initialize-project project-root-uri client-capabilities client-settings)))
 
 (defn did-open [{:keys [textDocument]}]
   (let [uri (:uri textDocument)
--- a/src/clojure_lsp/handlers.clj
+++ b/src/clojure_lsp/handlers.clj
@@ -23,7 +23,6 @@
     [clojure-lsp.queries :as q]
     [clojure-lsp.shared :as shared]
     [clojure.pprint :as pprint]
-    [clojure.string :as string]
     [medley.core :as medley]
     [rewrite-clj.node :as n]
     [rewrite-clj.zip :as z]
--- a/src/clojure_lsp/handlers.clj
+++ b/src/clojure_lsp/handlers.clj
@@ -30,7 +30,6 @@
   (:import
     [java.net
      URL
-     URLDecoder
      JarURLConnection]))
 
 (def ^:private full-file-range
@@ -50,7 +49,7 @@
     (crawler/initialize-project project-root client-capabilities client-settings)))
 
 (defn did-open [{:keys [textDocument]}]
-  (let [uri (-> textDocument :uri URLDecoder/decode)
+  (let [uri (:uri textDocument)
         text (:text textDocument)]
     (f.file-management/did-open uri text))
   nil)
--- a/src/clojure_lsp/handlers.clj
+++ b/src/clojure_lsp/handlers.clj
@@ -44,9 +44,9 @@
            (recur)
            ~@body)))))
 
-(defn initialize [project-root client-capabilities client-settings]
-  (when project-root
-    (crawler/initialize-project project-root client-capabilities client-settings)))
+(defn initialize [project-root-uri client-capabilities client-settings]
+  (when project-root-uri
+    (crawler/initialize-project project-root-uri client-capabilities client-settings)))
 
 (defn did-open [{:keys [textDocument]}]
   (let [uri (:uri textDocument)
@@ -135,7 +135,7 @@
 (defn ^:private server-info []
   (let [db @db/db]
     {:type :info
-     :message (with-out-str (pprint/pprint {:project-root (:project-root db)
+     :message (with-out-str (pprint/pprint {:project-root-uri (:project-root-uri db)
                                             :project-settings (:project-settings db)
                                             :client-settings (:client-settings db)
                                             :port (or (:port db)
@@ -266,24 +266,24 @@
 
 (defn prepare-call-hierarchy
   [{:keys [textDocument position]}]
-  (let [project-root (:project-root @db/db)]
+  (let [project-root-uri (:project-root-uri @db/db)]
     (f.call-hierarchy/prepare textDocument
                               (inc (:line position))
                               (inc (:character position))
-                              project-root)))
+                              project-root-uri)))
 
 (defn call-hierarchy-incoming
   [{:keys [item]}]
   (let [uri (:uri item)
         row (inc (-> item :range :start :line))
         col (inc (-> item :range :start :character))
-        project-root (:project-root @db/db)]
-    (f.call-hierarchy/incoming uri row col project-root)))
+        project-root-uri (:project-root-uri @db/db)]
+    (f.call-hierarchy/incoming uri row col project-root-uri)))
 
 (defn call-hierarchy-outgoing
   [{:keys [item]}]
   (let [uri (:uri item)
         row (inc (-> item :range :start :line))
         col (inc (-> item :range :start :character))
-        project-root (:project-root @db/db)]
-    (f.call-hierarchy/outgoing uri row col project-root)))
+        project-root-uri (:project-root-uri @db/db)]
+    (f.call-hierarchy/outgoing uri row col project-root-uri)))
--- a/src/clojure_lsp/handlers.clj
+++ b/src/clojure_lsp/handlers.clj
@@ -23,6 +23,7 @@
     [clojure-lsp.queries :as q]
     [clojure-lsp.shared :as shared]
     [clojure.pprint :as pprint]
+    [clojure.string :as string]
     [medley.core :as medley]
     [rewrite-clj.node :as n]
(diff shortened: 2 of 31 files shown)


ericdallo:
Thank you very much for the detailed description, that helps a lot reviewing those things.
We already had a lot of issues regarding decoding/uri/parsing for Windows and unix, so I agree we can improve that and make a better standard.

I can test this soon, but as this is something critical, and there are editors (LSP clients) that handle/send different "uris" and we can have issues with specific OSs, maybe we should confirm this change work with the current most used editors: Emacs, vim, and Calva(Maybe @bpringe can help us here). Also, It'd be great to make sure everything works on Windows, Linux, and MacOS. WDYT? 
I suppose you tested it on Emacs and Windows with both JVM/native image?

> Should project-root be renamed to project-root-uri or project-uri?

I like `project-root-uri`

> Should filename be renamed to filepath? (To me, 'name' implies solely the name of the file without the directory information)

I'm just not sure if kondo calls it as filename as well where this could cause a confusion in some destructurings/places, otherwise, it sounds good to me

LuisThiamNye:
> I'm just not sure if kondo calls it as filename as well where this could cause a confusion in some destructurings/places, otherwise, it sounds good to me

After a brief search, it does seem that clj-kondo uses the `:filename` key, so I'll just leave it as it is for now.

> I like project-root-uri

In that case, I will rename it.

> there are editors (LSP clients) that handle/send different "uris" and we can have issues with specific OSs

My hope was that as long as the URIs conform to [the specification](https://microsoft.github.io/language-server-protocol/specification.html#uri) and that they are encoded by the same rules, there should not be a problem. However, I am now realising that there may be inconsistency due to the fact that the composition of `shared/uri->filename` followed by `shared/filename->uri` is not an identity.

For instance, I briefly experimented with `vscode-uri` to find that `C:\` is encoded as `file:///c%3A/` — colons are encoded. However, Java's stuff does not encode the colons. In other words:

`(-> "file:///c%3A/x.clj" shared/uri->filename shared/filename->uri)` ⇒ `"file:///c:/x.clj"`

I can see this being a problem in `refactor.transform/inline-symbol?` at `(get-in @db/db [:documents (shared/filename->uri filename)])`

Thus, I think it would make sense to ensure all URIs received from external locations get passed through some conforming function that ensures that all URIs in the `db` follow the same encoding rules.

But as I think about this some more, I realise this approach would not work if the client expects to receive file URIs in the exact same format as it sent them out — there would be a mismatch.

Possible solution: where URIs are received from the client, they are recorded in a table mapping the file path to the URI. So, `filename->uri` can first attempt to look up a known URI from this table to ensure that there are never two different strings floating around that represent the same URI.

> Also, It'd be great to make sure everything works on Windows, Linux, and MacOS. WDYT? I suppose you tested it on Emacs and Windows with both JVM/native image?

I am currently using the native image on Mac (Emacs lsp-mode). Because Mac and Linux are both Unix-like, if it works on one it will probably work on the other.

Also, I have managed to get it running on a Windows VM with `client-vscode` using `clojure -M:run`. Finding references, highlighting and formatting appear to work. Once I feel like all the necessary changes have been made, I could then set up Calva and compile a native Windows image.

There are four unit test assertions failing on Windows but I think that is a separate issue I will open.

> maybe we should confirm this change work with the current most used editors: Emacs, vim, and Calva

Yes, it would be useful to establish some benchmark to do some formal testing. So far, symbol highlighting and renaming have been good indicators to me.

As I am using my custom native image with Emacs, I'll notice any problems as I run into them. I'll try setting up Calva too.

*Assuming everything goes right*, it should be relatively straightforward to plug in the native image to each editor.

```
         | macOS | Windows | Linux
Calva    |       |         |    
lsp-mode |       |         |    
Vim      |       |         |    
```

I haven't used Vim for any development so I'm probably not the most qualified person to test it.

ericdallo:
> Possible solution: where URIs are received from the client, they are recorded in a table mapping the file path to the URI. So, filename->uri can first attempt to look up a known URI from this table to ensure that there are never two different strings floating around that represent the same URI.

To be honest, I don't think there are clients that send URI not following the LSP spec, I think testing the major editors is enough, also I'd avoid adding that kind of logic just because an editor not following the standard

I'll keep testing this branch for a couple of days on NixOS, Emacs 28 with lsp-mode.

> I haven't used Vim for any development so I'm probably not the most qualified person to test it.

I think @snoe can help with that

LuisThiamNye:
> I don't think there are clients that send URI not following the LSP spec

[This issue](https://github.com/Microsoft/vscode/issues/2990) suggests that colon encoding may be invalid and that there is a workaround to it, so I'll go with the assumption that Java is correct and matches the LSP spec.

However, I still see the problem of whether the Windows drive letter should be upper or lower case in the URI — the specification is ambiguous. I am inclined to believe that lower case is the more correct approach since that would agree with `vscode-uri`. The problem is that java.io.File.getCanonicalPath (used by clj-kondo to produce "filenames") enforces an *upper case* drive letter, and java.net.URI does not enforce either way.

So then it might make sense for `filename->uri` to ensure a lower case drive letter in the final URI. But that could be problematic if the client chooses sends URIs with an upper case drive letter — something I have not seen any hard rules against.

Should it be assumed that URI drive letters are always lower case *(and that URIs follow the exact same rules)*? Personally, I do not feel very comfortable with that.

Related: [https://github.com/neovim/neovim/pull/14003](https://github.com/neovim/neovim/pull/14003)

---

Another issue I briefly alluded to is that on Windows, `feature.completion-test/test-completion` testing `complete-a` and `complete-alpaca` fail: the top two completion results are swapped around from what they should be.

In the `completion` function, the results are collected into a hash set, and then they are ordered with `sort-by :label`. Since the top two results have the same `:label` of `"alpaca"`, the result should be unpredictable.

Upon inspection, I think either the tests need to be adjusted to account for this or the implementation code needs to have a deterministic way of sorting results with the same label.

Originally, I thought this issue was unrelated, but it only occurs after the changes in this PR, suggesting that the different paths/URIs affect how the set is ordered. What is the best solution here?

ericdallo:
> Should it be assumed that URI drive letters are always lower case (and that URIs follow the exact same rules)? Personally, I do not feel very comfortable with that.

Good question, it'd be great to use the same pattern as kondo as they probably already had issues like that.

> Originally, I thought this issue was unrelated, but it only occurs after the changes in this PR, suggesting that the different paths/URIs affect how the set is ordered. What is the best solution here?

Both solutions work for me, but I'd go with the sorting by label and something else deterministic like the uri or something like that.

LuisThiamNye:
> it'd be great to use the same pattern as kondo

As far as I can tell, clj-kondo only deals with file paths, not URIs. Matching kondo's pattern would involve normalizing file paths to have upper case drive letters. Then what about URIs?

I installed Calva on Windows and the URIs it sends look like this:

`"file:///c%3A/Users/luist/dev/clojure-lsp/clojure-lsp/test/clojure_lsp/test_helper.clj"`

This matches the output of `URI.toString()` of `vscode-uri`. Thus, I do not think we can assume the URIs received from the client follow a single specification: it is unknown whether the drive letter will be upper case or whether the colons will be encoded.

Some ideas to handle this situation:

- Keep a memory of `uri->filename` conversions as I have described before
- Keep a memory of the client's preference of drive letter capitalisation and colon encoding.
- Do not convert filenames to URIs and modify code to work around that.

> I'd go with the sorting by label and something else deterministic like the uri or something like that

Here is an example result:

```clojure
{:label "alpaca"
:data {"name" "", "filename" "c:\\a.cljc", "name-row" 1, "name-col" 35}
:kind :property
:detail "user"
:additional-text-edits [{:range {:start {:line 0, :character 0}
                                 :end {:line 0, :character 9}}
                         :new-text "(ns user \n (:require\n [user :as alpaca]))"}]}
```

We could use `(get-in _ [:data "filename"])` to access the URI. Or how about simply sorting with `(juxt :label :detail)` since the detail property is explicitly given in the assertion:

```clojure
[{:label "alpaca" :kind :property :detail "user"}
 {:label "alpaca" :kind :property :detail "alpaca.ns"}
 {:label "alpha" :kind :variable}
 {:label "ba" :detail "alpaca.ns"}]
```

Although, this would mean that the top two results specified in the assertion will need to be swapped around. Would that be a problem?

I am also noticing that the tests for "complete-a" and "complete-alpaca" look exactly the same to me. I assume that is not meant to be the case?

ericdallo:
> Some ideas to handle this situation:
Keep a memory of uri->filename conversions as I have described before
Keep a memory of the client's preference of drive letter capitalisation and colon encoding.
Do not convert filenames to URIs and modify code to work around that.

The 2 option looks the best to me, but I'd like to avoid that kind of thing as we'd need to check that the first time we convert a URI or something like that. How we handle that today? I mean, if most/all clients send in the same pattern I'd not add/change that logic as it works well ATM. 


> Although, this would mean that the top two results specified in the assertion will need to be swapped around. Would that be a problem? 

I don't think it's a problem.

> I am also noticing that the tests for "complete-a" and "complete-alpaca" look exactly the same to me. I assume that is not meant to be the case?

Maybe the test is not clear enough, but it's matching the labels/the correct quantity of completion items

LuisThiamNye:
> if most/all clients send in the same pattern I'd not add/change that logic as it works well ATM.

I think the reason it may currently handle Calva's encoding of colons for some things is because URIs are 'decoded' with URLDecoder which [should not be done](https://stackoverflow.com/a/17870390).

> The 2 option looks the best to me, but I'd like to avoid that kind of thing as we'd need to check that the first time we convert a URI or something like that. How we handle that today?

`crawler/initialize` looks like a good candidate for setting the URI configuration, since it receives the project root as a URI.

Another option is to not store URIs in the `db` as keys but instead use filepaths. Assuming clients do not require the responses to have URIs in their exact format, this would mean that `filename->uri` would not need to worry about the output format. Although, that would mean quite a bit of refactoring. Does this sound like a better solution?

> Maybe the test is not clear enough, but it's matching the labels/the correct quantity of completion items

I understand what the tests do, but when I compare the two tests, I do not see why they should ever produce different results.

Here is the test for "complete-a"

```clojure
(testing "complete-a"
    (h/assert-submaps
     [{:label "alpaca" :kind :property :detail "user"}
      {:label "alpaca" :kind :property :detail "alpaca.ns"}
      {:label "alpha" :kind :variable}
      {:label "ba" :detail "alpaca.ns"}]
     (f.completion/completion (h/file-uri "file:///b.clj") 3 3)))
```

Here is the test for "complete-alpaca"

```clojure
(testing "complete-alpaca"
    (h/assert-submaps
     [{:label "alpaca" :kind :property :detail "user"}
      {:label "alpaca" :kind :property :detail "alpaca.ns"}
      {:label "alpha" :kind :variable}
      {:label "ba" :detail "alpaca.ns"}]
     (f.completion/completion (h/file-uri "file:///b.clj") 3 3)))
```

Unless the tests have side effects, "complete-alpaca" looks redundant to me.

LuisThiamNye:
Or we make sure to sanitise URIs in all the places they are received from the client, before they are put into the db.

ericdallo:
> crawler/initialize looks like a good candidate for setting the URI configuration, since it receives the project root as a URI.

Oh, indeed, I forgot it was sent as a URI not a path

> Unless the tests have side effects, "complete-alpaca" looks redundant to me.

I see, I think this test was included to handle completion of labels with the exact string (not just starting with), but it does seems the same to me too.

> Or we make sure to sanitise URIs in all the places they are received from the client, before they are put into the db.

yeah, that sounds safer I think than use filepath on everything

LuisThiamNye:
At the moment it seems it has been narrowed down to two options:

1. Initialise URI formatting preferences in `crawler/intialize` and use this in `filename->uri`
2. Sanitise all URIs that the server receives

Which sounds more ideal?

My impression: Option 1 looks much more straightforward. Option 2 sounds good but I'm concerned it would be easier to let unsanitised URIs slip through the gaps, as LSP messages come in different shapes.

I'll probably start working on one of these and see how that turns out.

---

As for the tests, what should be done about them. Should I open a new issue?

ericdallo:
I'd go with first option as we are having issues with that since begining and trying follow the 2 option, maybe we should try 1 option.

Please open a new issue about the tests.

bpringe:
I did not read this whole thread, but please ping me when/if something could use testing in Calva.

LuisThiamNye:
An overview of the latest changes:

- Rename `project-root` to `project-root-uri`
- Infer URI format settings from the project-root-uri in handlers.clj — and store this in the `db` along with other `:settings`
    - Enforce these settings in `filename->uri`
- Enforce upper case drive letters in filenames in `uri->filename`
- Sort completion results by `:detail` for results with the same `:label`
    - Change the order of completion results in the test assertions

src/clojure_lsp/crawler.clj
```diff
@@ -209,26 +209,25 @@
...
+        settings (-> (merge (:settings @db/db) project-settings)
                      (update :source-paths (fn [source-paths] (mapv #(str (.getAbsolutePath (to-file root-path %))) source-paths)))
                      (update :cljfmt cljfmt.main/merge-default-options))]
     (when-let [log-path (:log-path settings)]
       (logging/update-log-path log-path))
     (swap! db/db assoc
-           :project-root project-root
+           :project-root-uri project-root-uri
            :project-settings project-settings
-           :client-settings client-settings
```
ericdallo, line 231: Why remove the client settings? Without that we can surely know if a setting was setted via a client (lsp-mode, calva) or a user (.lsp/config.edn)

ericdallo: Oh got it, you just did it sooner

ericdallo: Hum, it's a little bit odd we add the settings the call before this function and then change it again on this function, can't we just make in a single place?

LuisThiamNye: I moved it to handlers/initialize where I am also setting the settings for the URI format.

ericdallo: Yeah, my point is that we `swap!` those settings before `initialize-project` and right after calling it here, overriding the `:settings`, can't we do that logic in a single function in a single place to make it simpler?

ericdallo:
Oh, I needed to rollback a commit on master and I messed up this branch commit history for some reason, sorry about that

src/clojure_lsp/crawler.clj
LuisThiamNye: Yes, that would be better.

LuisThiamNye: I've move all the settings initialisation into the initialize-project function, similar to how it was before.

LuisThiamNye:
With the latest change, I just made sure that `uri->path` returns a Path object with an upper case drive letter, so that the source-paths are correctly normalised.

I've done some testing on Windows and Mac with Calva and client-vscode — things like rename and formatting work as expected. Also, the log shows the server responding with URIs like `file:///c%3A/...` to match the client.

The patch will work under the assumption that the client will be consistent about the format they send their URIs in. That is to say, they would not send `file:///C:/path` and later send `file:///c:/path`. In my limited experience so far (with Calva), this has been true.

ericdallo:
Nice, I'll take a closer look at the changes today and make some tests as well

bpringe:
I'm loving the effort and attention to detail and testing on this PR. :heart: Love to see clojure-lsp continue improving as it has been.

src/clojure_lsp/handlers.clj
```diff
@@ -23,14 +23,14 @@
     [clojure-lsp.queries :as q]
     [clojure-lsp.shared :as shared]
     [clojure.pprint :as pprint]
+    [clojure.string :as string]
```
ericdallo, line 26:
```suggestion
```

test/clojure_lsp/shared_test.clj
```diff
@@ -1,35 +1,75 @@
 (ns clojure-lsp.shared-test
   (:require
-   [clojure-lsp.db :as db]
-   [clojure-lsp.shared :as shared]
-   [clojure.test :refer [deftest is testing]]))
+    [clojure-lsp.db :as db]
+    [clojure-lsp.shared :as shared]
+    [clojure-lsp.test-helper :as h]
+    [clojure.string :as string]
```
ericdallo, line 6:
```suggestion
```

src/clojure_lsp/interop.clj
```diff
@@ -1,12 +1,13 @@
 (ns clojure-lsp.interop
   (:require
+    [clojure-lsp.shared :as shared]
```
ericdallo, line 3:
```suggestion
```

ericdallo (approved):
Nice tests!

LGTM, thanks for the hard work!

PR #446, opened 2021-06-15
Use last definition when showing signature help

nenadalm:
**Problem**
In case there are redefinitions, clojure-lsp seems to show signature help for first definition. Last definition should be used as it overwrites previous definitions.

**Example**
Clojure itself is using redefinition of `map` where first definition should probably use `declare` instead of `def` (maybe `declare` didn't exist at the time the code was written?). This probably causes `map` to not have signature help.

- here `map` is declared using `def`: https://github.com/clojure/clojure/blob/3c9307e27479d450e3f1cdf7903458b83ebf52d7/src/clj/clojure/core.clj#L1713
- and here it's implemented: https://github.com/clojure/clojure/blob/3c9307e27479d450e3f1cdf7903458b83ebf52d7/src/clj/clojure/core.clj#L2731

**Side effect**
As a side effect of this change, cljs is preferred over clj in this test: https://github.com/clojure-lsp/clojure-lsp/compare/master...nenadalm:signature_help?expand=1#diff-1a73eefdddf48f04bcc32d792584b66cf82fd4ea953f2087304e271daab95015R209 - I don't know if it matters which dialect is used? I guess not, so I just updated the test.

### DIFF
--- a/src/clojure_lsp/queries.clj
+++ b/src/clojure_lsp/queries.clj
@@ -39,10 +39,6 @@
   (->> analysis
        (remove-keys #(string/starts-with? (-> % name shared/filename->uri) "file://"))))
 
-(defn ^:private find-first-order-by-project-analysis [pred? analysis]
-  (or (find-first pred? (mapcat val (filter-project-analysis analysis)))
-      (find-first pred? (mapcat val (filter-external-analysis analysis)))))
-
 (defn ^:private find-last-order-by-project-analysis [pred? analysis]
   (or (find-last pred? (mapcat val (filter-project-analysis analysis)))
       (find-last pred? (mapcat val (filter-external-analysis analysis)))))
@@ -91,7 +87,7 @@
 
 (defmethod find-definition :namespace-usages
   [analysis element]
-  (find-first-order-by-project-analysis
+  (find-last-order-by-project-analysis
     #(and (= (:bucket %) :namespace-definitions)
           (= (:name %) (:name element))
           (match-file-lang % element))
@@ -115,7 +111,7 @@
 (defmethod find-definition :keywords
   [analysis element]
   (when (:ns element)
-    (find-first-order-by-project-analysis
+    (find-last-order-by-project-analysis
       #(and (= (:bucket %) :keywords)
             (= (:name %) (:name element))
             (:reg %)
--- a/src/clojure_lsp/queries.clj
+++ b/src/clojure_lsp/queries.clj
@@ -25,6 +25,9 @@
     nil
     coll))
 
+(defn- find-last [pred coll]
+  (find-first pred (reverse coll)))
+
 (defn ^:private remove-keys [pred m]
   (apply dissoc m (filter pred (keys m))))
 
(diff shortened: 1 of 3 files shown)


ericdallo (changes requested):
Thanks for the help but changing the queries could bring other issues that is what that test was testing, imagine there is a function foo in both project and in classpath analysis,  with your change the go to definition goes to the classpath but it should go to the project definition.

I think we should try to just declare on the signatureHelp somehow

nenadalm:
> Thanks for the help but changing the queries could bring other issues that is what that test was testing, imagine there is a function foo in both project and in classpath analysis, with your change the go to definition goes to the classpath but it should go to the project definition.
> 
> I think we should try to just declare on the signatureHelp somehow

I am not too familiar with this project, but based on: https://github.com/clojure-lsp/clojure-lsp/pull/446/files#diff-e0f84cbb17c81fa54eae9b5d2745f4544a28b57c92f4b851acb14a673b822d70R46-R48 it seems that function is first still searched in project (`filter-project-analysis`) and then elsewhere (`filter-external-analysis`) so it should probably still work in a way that project definitions are preferred over other definitions?

ericdallo:
Yes, you are right, it might work, but I still I'd suggest taking another approach as we are starting to change some code to check last while other check first, it's odd... It'd be great if we could try avoid declare instead of getting the last, WDYT?

nenadalm:
> It'd be great if we could try avoid declare instead of getting the last, WDYT?

I am not sure what do you mean by `avoid declare`? Should we somehow detect that https://github.com/clojure/clojure/blob/3c9307e27479d450e3f1cdf7903458b83ebf52d7/src/clj/clojure/core.clj#L1713 is used instead of `declare` and skip it? I am not sure how to do that.

To me it makes sense that when I am looking for var, I am looking for last definition - at least in a file. So maybe alternative would be looking for last definition in first found file?

> we are starting to change some code to check last while other check first

It would be probably ok to check last of everything (`:namespace-usages`, `:keywords`)? I could try to change that and then there would be no problem with doing something first and something last?

ericdallo:
> To me it makes sense that when I am looking for var, I am looking for last definition - at least in a file. So maybe alternative would be looking for last definition in first found file?

Looks valid to me, as clojure use the same idea of redefined vars.

> It would be probably ok to check last of everything (:namespace-usages, :keywords)? I could try to change that and then there would be no problem with doing something first and something last?

Yeah, looks better, following the above idea of the last of the first file.

Even with these changes, we need to test it to check if we don't break anything, also some integration-test for that would be great, LMK if need any help

nenadalm:
I've updated the pr so that last definition is always used (first in project, then elsewhere if not found in project).

src/clojure_lsp/queries.clj
```diff
@@ -36,9 +39,9 @@
   (->> analysis
        (remove-keys #(string/starts-with? (-> % name shared/filename->uri) "file://"))))
 
-(defn ^:private find-first-order-by-project-analysis [pred? analysis]
-  (or (find-first pred? (mapcat val (filter-project-analysis analysis)))
-      (find-first pred? (mapcat val (filter-external-analysis analysis)))))
+(defn ^:private find-last-order-by-project-analysis [pred? analysis]
```
ericdallo, line 42: this function name is not correct anymore as we find the last but checking the project before, but I can't think in a better name

ericdallo (approved):
It seems to work great, thanks for the contribution!

ericdallo:
@nenadalm would you mind adding a entry to CHANGELOG.md?

nenadalm:
Thanks for the merge. I see that you already updated the changelog.

Issue #215, opened 2020-12-24
Idea: REPL-oriented helpers


vemv:
Hi!

I'd be interested in clojure-lsp for its refactorings.

Personally I don't want to integrate my editor with clojure-lsp, which is not really a judgement on it (if you're curious: I use forked/customized cider, clojure-mode, refactor-nrepl etc and don't really want to ditch that hard work I once did),

So, invoking clojure-lsp from the repl would allow me to use its goodies while not having to touch my editor setup. I also can imagine how repl-oriented helpers can widen the reach of clojure-lsp to more people, because other people might be change-averse like me, and also there's a great deal of repl-orientation in the community :)

## Suggestion

* For a select set of operations, offer a "repl helper" function
  * e.g. `rename-function` or `rename-ns` make sense from the repl, `cycle-privacy` seems far less likely to be invoked from a repl
* Make the API favor vars over textual references
  *  e.g. `(rename #'foo :to 'bar)` seems preferable to `(rename "the/file.clj:10:20" to 'bar)`, as it's higher-level
    * `#'foo` (var syntax) make the API robust and self-evident, since the Clojure compiler forbids references to non-existing vars
* Perhaps, augment operations' intelligence when invoked from the repl
  * When invoked from the repl, one can access Clojure's var system, which has metadata, etc
  * My understanding is that when clojure-lsp is invoked in a vanilla way, it has to use heuristics for certain things

FWIW, I've followed this kind of approach in https://github.com/nedap/formatting-stack/ which invokes cljfmt, clj-kondo, eastwood, etc from the repl, which assuming an existing repl process, has great performance and occasionally great runtime-powerered intelligence. It has worked really well although admittedly it's not for everyone.

With a bit of luck, if the idea took off it might result in more people contributing to clojure-lsp even if they're not direct users, filling a niche similar to https://github.com/clojure-emacs/refactor-nrepl (a project which I love but tends to perform badly in big projects due to tools.analyzer)

WDYT?

Thanks - V

ericdallo:
Hi, It seems a really good idea indeed :) 
I just not sure how hard it would be, we could start with a `clojure-lsp.repl` ns accessing the refactoring ns exposing the common code action functions, WDYT?
I'm not sure we need to do anything more to other libs use this project as repl though

vemv:
Thanks for the response!

> we could start with a clojure-lsp.repl ns accessing the refactoring ns exposing the common code action functions, WDYT?

Sure :)

[...]

PR #455, opened 2021-06-26
clojure-lsp as API / CLI

ericdallo:
Fixes #https://github.com/clojure-lsp/clojure-lsp/issues/215

This PR introduces an API to be accessed as a lib and CLI to use clojure-lsp features outside an editor.
The first (only for now) method available is the `clean-ns` where it uses the `clean-ns` refactoring.

```bash
clojure-lsp clean-ns
```
with optional `--project-root` and `--namespace` argumetns

### DIFF
--- a/src/clojure_lsp/api.clj
+++ b/src/clojure_lsp/api.clj
@@ -1,44 +1,15 @@
 (ns clojure-lsp.api
   (:require
-   [clojure-lsp.client :as client]
-   [clojure-lsp.crawler :as crawler]
-   [clojure-lsp.db :as db]
-   [clojure-lsp.handlers :as handlers]
-   [clojure-lsp.interop :as interop]
-   [clojure-lsp.queries :as q]
-   [clojure-lsp.shared :as shared]))
+    [clojure-lsp.internal-api :as internal-api]))
 
-(defn ^:private start-analysis!
-  [project-root]
-  (let [project-uri (shared/filename->uri project-root)]
-    (crawler/initialize-project
-      project-uri
-      {:workspace {:workspace-edit {:document-changes true}}}
-      (interop/clean-client-settings {:dependency-scheme "jar"})
-      {:lint-project-files-after-startup? false
-       :text-document-sync-kind :full})))
+(defn clean-ns!
+  "Organize ns form, removing unused requires/refers/imports and sorting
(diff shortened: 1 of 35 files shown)


src/clojure_lsp/api.clj
```diff
@@ -0,0 +1,15 @@
+(ns clojure-lsp.api
+  (:require
+    [clojure-lsp.internal-api :as internal-api]))
+
+(defn clean-ns!
+  "Organize ns form, removing unused requires/refers/imports and sorting
+  alphabetically.
+  Options should be passed as keyword arguments and it accepts:
+
+  `:project-root` a string representing the path to the project root.
```
vemv, line 10: > `:project-root` a string representing the path to the project root.

Maybe accepting it as a File instead would invite people to do the right thing (namely, not hardcoding `/`s).

Might also be handy to `{:pre [(.exists the-file)]}` - APIs that fail-fast at the right level of abstraction tend to be friendlier

vemv (commented):
LGTM!

(FWIW and all that)

src/clojure_lsp/api.clj
ericdallo: Thank you! I was wondering if this would be a file or a string path indeed, clojure-lsp would accept both with or without the `/`, but it seems to makes sense come as a file indeed :)

ericdallo: Done!

vemv:
🚀

vemv:
Btw, I assume that so far no state has to be initialized, maintained etc?

That was an interesting point that popped up in https://github.com/clojure-lsp/clojure-lsp/issues/215

ericdallo:
@vemv actually, the state is still used, but similar to how clj-kondo works, it tries to find cache folder `.lsp/sqlite.db` otherwise it scans during the call, which for large projects should take some seconds while others could be very fast, this is probably the result of the performance improvements done this year including the use of GraalVM :)

ericdallo:
I found some improvements that can be done specifically for the API later, like caching the analyzed during multiple calls, but I'll add more API features before tackling that :)

vemv:
Makes sense, thanks!

Looking forward to hack some things on top of LSP as soon as I get the chance

PR #459, opened 2021-06-28
Added comment to Makefile suggesting updating the clojure launcher

simon-brooke:
See issue #454 

The Makefile comment here is slightly more general than that I suggested in the issue report, since, if people hit issues other than the one I encountered, updating their clojure launcher script to current is probably a good idea anyway.


ericdallo (approved):
Thank you!

PR #475, opened 2021-07-12
Respect XDG specification regarding empty XDG_CONFIG_HOME

### LINKED ISSUE #474: Empty XDG_CONFIG_HOME should default to "$HOME/.config"
Not all GNU/Linux distributives set the `XDG_CONFIG_HOME` variable to some value. For example Fedora doesn't. [According to XDG spec](https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html):

> `$XDG_CONFIG_HOME` defines the base directory relative to which user-specific configuration files should be stored. **If `$XDG_CONFIG_HOME` is either not set or empty, a default equal to `$HOME/.config` should be used.**

Right now, `clojure-lsp` is checking if variable is set, and if it is not it defaults to `$HOME/.lsp`:

https://github.com/clojure-lsp/clojure-lsp/blob/dd2d3082a08022a3d5a2ca8896de031dc08c2a94/src/clojure_lsp/config.clj#L65-L68

This doesn't exactly follow the XDG conventions. Instead, when `(get-env "XDG_CONFIG_HOME")` returns `nil`, the `(io/file (get-property "user.home") ".config" ".lsp" "config.edn")` should be checked for existence. And if it doesn't exist, the ` (io/file (get-property "user.home") ".lsp" "config.edn")` should be tried next.

**To Reproduce**

1. Unset `XDG_CONFIG_HOME` variable
2. Put `.lsp` directlry to `$HOME/.config/ 
3. Start `clojure-lsp` in any project, and see that server configuration is not applied

**Expected behavior**

Config is read from `$HOME/.config/.lsp/config.edn`

 - OS: Fedora 34
 - Version: clojure-lsp 2021.06.01-16.19.44 clj-kondo 2021.06.01

andreyorst:
fix #474

This changes the `get-home-config-file` function to respect XDG specification regarding empty `XDG_CONFIG_HOME`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- API/CLI
+  - Fix empty `XDG_CONFIG_HOME` not defaulting to `$HOME/.config` #474
+
 ## 2021.07.12-12.30.59
 
 - Editor
--- a/src/clojure_lsp/config.clj
+++ b/src/clojure_lsp/config.clj
@@ -63,12 +63,13 @@
   (.exists f))
 
 (defn ^:private get-home-config-file []
-  (if-let [xdg-config-home (get-env "XDG_CONFIG_HOME")]
-    (io/file xdg-config-home ".lsp" "config.edn")
-    (let [xdg-config-default (io/file (get-property "user.home") ".config" ".lsp" "config.edn")]
-      (if (file-exists? xdg-config-default)
-        xdg-config-default 
-        (io/file (get-property "user.home") ".lsp" "config.edn")))))
+  (let [xdg-config-home (or (get-env "XDG_CONFIG_HOME")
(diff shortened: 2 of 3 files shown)


ericdallo:
Thanks, that looks great, could you fix the lint and add a entry to changelog?

Also, would be perfect to add a test for that, WDYT? calling the function from test with `#'get-home-config-file` as it's private

andreyorst:
> Also, would be perfect to add a test for that, WDYT? calling the function from test with `#'get-home-config-file` as it's private

got it!

Also, I've actually understood that my the logic is still is not right, because if `XDG_CONFIG_HOME` is set, but there's no config, we should also try `~/.lsp/` directory. I think I need to rework the logic a bit more here.

ericdallo:
Yeah, looks a good fallback, one more reason to have tests for that :p

ericdallo (approved):

PR #478, opened 2021-07-13
Clean up hover output

NoahTheDuke:
Instead of returning a text string to the editor, we can return multiple strings or objects, each of which will be formatted as a unit. (Per the [spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_hover).) This removes the need for hand-writing the dashes in most cases. I also removed the additional linebreak in the docstring portion, and merged the symbol and signature into a single code block.

With markdown and multi-arity params on separate lines: 
image

With markdown and multi-arity params on same line:
image

With text and multi-arity params on separate lines:
image

With text and multi-arity params on same line:
image

I briefly tried this out with vs code and Calva and it looks good there as well, tho I don't have screenshots handy.

### DIFF
--- a/src/clojure_lsp/feature/hover.clj
+++ b/src/clojure_lsp/feature/hover.clj
@@ -53,7 +53,7 @@
                 doc-line (str line-break doc-line)
                 filename (str (format "%s*[%s](%s)*"
                                       line-break
-                                      (string/replace filename #"\\" #"\\\\")
+                                      (string/replace filename #"\\" "\\\\")
                                       (shared/filename->uri filename))))}
       ;; Default to plaintext
       (cond->> []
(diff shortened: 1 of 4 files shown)


src/clojure_lsp/feature/hover.clj
```diff
@@ -34,31 +34,37 @@
...
                      doc))]
     (if markdown?
       {:kind "markdown"
        :value (cond-> (str opening-code sym-line closing-code)
-                signatures-line (str "\n" opening-code signatures-line closing-code)
-                doc-line (str line-break doc-line "\n")
+                doc-line (str line-break doc-line)
                 filename (str (format "%s*[%s](%s)*"
                                       line-break
                                       (string/replace filename #"\\" "#\\\\")
```
ericdallo, line 56: Would you mind fixing this added recently?
```suggestion
                                      (string/replace filename #"\\" #"\\\\")
```

ericdallo (approved):
Looks great!
Would you mind add an entry to CHANGELOG.md under `Unreleased`?

NoahTheDuke:
@ericdallo I think it's good to go now!

ericdallo (approved):

PR #483, opened 2021-07-15
Change thread refactoring to work when not on parentheses

NoahTheDuke:
https://user-images.githubusercontent.com/603677/125862187-7ee2d65b-0436-4024-8b71-fac526cc3e1f.mov

### DIFF
--- a/src/clojure_lsp/refactor/transform.clj
+++ b/src/clojure_lsp/refactor/transform.clj
@@ -109,6 +109,13 @@
        (not (contains? thread-invalid-symbols
                        (some-> zloc z/next z/sexpr)))))
 
+(defn can-thread-at-cursor? [zloc]
+  (or (can-thread? zloc)
+      (and (= (z/tag zloc) :token)
+           (= (z/tag (z/up zloc)) :list)
+           (not (contains? thread-invalid-symbols
+                           (some-> zloc z/up z/next z/sexpr))))))
+
 (defn thread-first
   [zloc]
   (when (can-thread? zloc)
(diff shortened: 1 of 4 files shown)


src/clojure_lsp/refactor/transform.clj
```diff
@@ -109,6 +109,13 @@
...
                        (some-> zloc z/next z/sexpr)))))
 
+(defn can-thread-at-cursor? [zloc]
+  (or (can-thread? zloc)
+      (and (= (z/tag zloc) :token)
+           (= (z/tag (z/up zloc)) :list)
+           (not (contains? thread-invalid-symbols
+                           (some-> zloc z/up z/next z/sexpr))))))
+
 (defn thread-first
```
ericdallo, line 119: Should we change all other references that use `can-thread?` to use the new `can-thread-at-cursor?`

NoahTheDuke: Good catch. I've renamed the functions to be more clear: `can-thread?` checks both if the current zloc is a list or if it's a token within a list, and `can-thread-list?` only checks if the current zloc is a list. We want only the latter when calling `thread-all` to prevent constantly rechecking the updated zloc, as `(z/up next-loc)` points to the in-progress thread macro rewrite, which will always start with `->` or `->>`.

NoahTheDuke:
Is there a flaky test? My changes don't intersect with the failed test (`clojure-lsp.handlers-test/did-open`).

ericdallo:
> Is there a flaky test? My changes don't intersect with the failed test (clojure-lsp.handlers-test/did-open).

Yes, I never faced this test error, but it seems to be flaky indeed, you can re-run the check to confirm that.

Thanks for the changes, I'll test it manually later today

ericdallo (approved):
It works, thank you!

I just realized we could in the future add a code action to the refactoring `unwind-all` if inside a thread

PR #486, opened 2021-07-20
Update clients.md

askonomm:
Adding Sublime Text editor to the list of clients, which I love dearly, and use daily for Clojure development thanks to Clojure LSP.


ericdallo (approved):
Thank you!

PR #491, opened 2021-07-25
Support bb classpath/source-paths integration

ericdallo:
Depends on https://github.com/babashka/babashka/issues/733


ericdallo:
Done here: https://github.com/clojure-lsp/clojure-lsp/pull/503

PR #516, opened 2021-08-13
Clean client settings when passed strings instead of keywords

NoahTheDuke:
Certain LSP clients (such as [coc.nvim](https://github.com/neoclide/coc.nvim)) use json when defining the [Initialization Options](https://clojure-lsp.github.io/clojure-lsp/settings/#initializationoptions) and so don't have a concept of `:keywords`. To better handle those clients, try to parse certain setting inputs:
* if it's a keyword, use it
* if it's a string that's written like a keyword, trim the `:` and convert to keyword
* if it's a string otherwise, convert to keyword
* return `nil` otherwise

There are no tests for `clean-client-settings` or I would have added a couple. Is this something you'd like?


ericdallo:
Thank you! Yeah, would love to see tests for that case if you don't mind

ericdallo:
Can you also rebase with master? I fixed some flaky unit/integration tests

ericdallo (approved):
Looks great! thank you!

PR #541, opened 2021-09-02
Test aarch64 graalvm image

ericdallo:


iruYo:
Hello! I came by this by chance since I couldn't get clojure-lsp running on my raspberry pi. 
I tried compiling it myself, but got stuck with some missing libraries error I couldn't fix.
Are there still plans to get this working on aarch64?

PR #590, opened 2021-10-07
Fix expand-let bug that occurs when a list form precedes let

athos:
This PR fixes an issue that `expand-let` results in a wrong form in a certain situation. 

## Issue

Concretely, `expand-let` fails in the expansion if there is another list form immediately before the `let` form:

```clojure
(require '[rewrite-clj.zip :as z]
         '[clojure-lsp.refactor.transform :as t])

(def s "(+ (* 3 3) (let [x 4] (* x x)))")
(def loc (-> s (z/of-string) (z/find-value z/next 'let) (z/up)))

(t/expand-let loc)
;; Expected: (let [x 4] (+ (* 3 3) (* x x)))
;; Actual:   (let [x 4] (* 3 3))
```

You can see what's happening by tracing transitions of the zipper's "current location" step by step as follows (Here, "the current location" will be denoted with `<< >>`):

```clojure
(-> let-loc     ;; (+ (* 3 3) <<(let [x 4] (* x x))>>)
    (z/splice)  ;; (+ (* 3 3) <<let>> [x 4] (* x x))
    (z/remove)  ;; (+ (* 3 <<3>>) [x 4] (* x x))
    (z/next)    ;; (+ (* 3 3) <<[x 4]>> (* x x))
    (z/remove)  ;; (+ (* 3 <<3>>) (* x x))
    (z/find z/up #(= (z/tag %) :list)) ;; (+ <<(* 3 3)>> (* x x))
```
(The code was taken from [clojure_lsp/refactor/transform.clj](https://github.com/clojure-lsp/clojure-lsp/blob/4c87f082f5650eaecb01c5a388cca69c68181915/src/clojure_lsp/refactor/transform.clj#L287-L292))

It seems that using `z/remove` together with `z/next` is somewhat brittle because it can be affected by the context surrounding the `let` form.

## Proposal

This PR proposes a workaround that avoids the fragility of `z/remove` and `z/next` by inserting a dummy element `::dummy`:

```clojure
(-> let-loc                  ;; (+ (* 3 3) <<(let [x 4] (* x x))>>)
    (z/insert-child ::dummy) ;; (+ (* 3 3) <<(::dummy let [x 4] (* x x))>>)
    (z/splice)               ;; (+ (* 3 3) <<::dummy>> let [x 4] (* x x))
    (z/right)                ;; (+ (* 3 3) ::dummy <<let>> [x 4] (* x x))
    (z/remove)               ;; (+ (* 3 3) <<::dummy>> [x 4] (* x x))
    (z/right)                ;; (+ (* 3 3) ::dummy <<[x 4]>> (* x x))
    (z/remove)               ;; (+ (* 3 3) <<::dummy>> (* x x))
    (z/find z/up #(= (z/tag %) :list)) ;; <<(+ (* 3 3) ::dummy (* x x))>>
    (z/edit->
     (z/find-value z/next ::dummy)
     (z/remove))             ;; <<(+ (* 3 3) (* x x))>>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
   
 - Editor
   - Add reference code lens for ns forms. #578
+  - Fix expand-let bug that occurs when a list form precedes let. #590
 
 
 ## 2021.09.30-15.28.01


athos:
OK, will do later 👍

athos:
Sorry, I'm not sure what category this change should be in. Editor or API?

ericdallo:

PR #612, opened 2021-10-24
Clojuredocs raw

PEZ:
I tried to follow the prior art. Thought of it as a mix of the server-info/raw command (which takes no argument and returns the results) and cursor-info (which takes arguments but doesn't return results).

When executed, it crashes with the following log:

```
[Trace - 10:08:48 AM] Sending request 'clojure/clojuredocs/raw - (149)'.
Oct 24, 2021 10:08:48 AM org.eclipse.lsp4j.jsonrpc.RemoteEndpoint fallbackResponseError
SEVERE: Internal error: java.lang.reflect.InvocationTargetException
java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
	at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.lambda$null$0(GenericEndpoint.java:67)
	at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.request(GenericEndpoint.java:120)
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.handleRequest(RemoteEndpoint.java:261)
	at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.consume(RemoteEndpoint.java:190)
	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.handleMessage(StreamMessageProducer.java:194)
	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.listen(StreamMessageProducer.java:94)
	at org.eclipse.lsp4j.jsonrpc.json.ConcurrentMessageProcessor.run(ConcurrentMessageProcessor.java:113)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:835)
Caused by: java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:567)
	at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.lambda$null$0(GenericEndpoint.java:65)
	... 11 more
Caused by: java.lang.UnsupportedOperationException: clojuredocsRaw
	at clojure_lsp.server.proxy$clojure_lsp.ClojureExtensions$ExtraMethods$LanguageServer$4129860f.clojuredocsRaw(Unknown Source)
	... 16 more

[Trace - 10:08:48 AM] Received response 'clojure/clojuredocs/raw - (149)' in 2ms. Request failed: Internal error. (-32603).
```

My inline defs in the handler function never gets defined so it must crash before that, I am thinking. But I have no clue where.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
   - Improve rename feature UX to output errors when it's not possible rename.
   - Add new `Unwind thread once` and `Unwind whole thread` code actions to undo a thread call.
   - Improve code actions performance request async all actions.
-  - Add new LSP custom method `clojure/cursordocs/raw` which takes a symbol and a namespace (both strings) and returns any Clojuredocs entry found, otherwise `null`.
+  - Add new LSP custom method `clojure/clojuredocs/raw` which takes a symbol and a namespace (both strings) and returns any Clojuredocs entry found, otherwise `null`.
 
 ## 2021.10.20-16.49.47
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
   - Improve rename feature UX to output errors when it's not possible rename.
   - Add new `Unwind thread once` and `Unwind whole thread` code actions to undo a thread call.
   - Improve code actions performance request async all actions.
+  - Add new LSP custom method `clojure/cursordocs/raw` which takes a symbol and a namespace (both strings) and returns any Clojuredocs entry found, otherwise `null`.
 
 ## 2021.10.20-16.49.47
 
--- a/docs/capabilities.md
+++ b/docs/capabilities.md
@@ -125,3 +125,4 @@ Other clients might provide a higher level interface to `workspace/executeComman
 | clojure/serverInfo/raw |                                | Use to retrieve from server the server configuration information      |
 | clojure/serverInfo/log |                                | Use to log to user the server configuration information               |
 | clojure/cursorInfo/log | `[document-uri, line, column]` | Use to log to user the debugging information for the symbol at cursor |
+| clojure/clojuredocs/raw | `[symbol-name, symbol-ns]` | Use to retreive any Clojuredocs entry. Will return `null` if no entry found.  |
--- a/resources/META-INF/native-image/clojure-lsp/clojure-lsp/reflect-config.json
+++ b/resources/META-INF/native-image/clojure-lsp/clojure-lsp/reflect-config.json
@@ -33,6 +33,13 @@
   "methods":[{"name":"<init>","parameterTypes":[] }]
 },
 {
+  "name":"org.eclipse.lsp4j.ClojuredocsParams",
+  "allDeclaredMethods":true,
+  "allPublicMethods": true,
+  "allDeclaredFields":true,
+  "methods":[{"name":"<init>","parameterTypes":[] }]
+},
+{
   "name":"com.google.gson.JsonObject",
   "allPublicMethods":true
 },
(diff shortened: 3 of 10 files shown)


ericdallo:
@PEZ could you resolve the conflicts as well?
I'd like to add this new command to next release

PEZ:
I'll look at this today, hopefully. I'm not sure about the API yet, but I have a client now so I can see what they negotiate. 😄

PEZ:
I have resolved the conflicts. However, still unsure about the API. Calva doesn't really know the fully qualified namespace of symbols. With this API I will need to make Calva able to resolve those. Yet, since clojure-lsp knows how to resolve it, I want to explore what the call should look like in order for us to utilize that.

PEZ:
@ericdallo: I realized that for Calva's current need of this command, Calva will always have a REPL connection, and I can use that to resolve the namespace. This is now good to go as far as I can tell.

ericdallo:
@PEZ regarding calva, maybe you can use the clojure-lsp command `cursorInfo` which brings information about the symbol at point, and then get the ns from kondo, this way you don't need a repl to query clojuredocs :)

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
   - Improve rename feature UX to output errors when it's not possible rename.
   - Add new `Unwind thread once` and `Unwind whole thread` code actions to undo a thread call.
   - Improve code actions performance request async all actions.
+  - Add new LSP custom method `clojure/cursordocs/raw` which takes a symbol and a namespace (both strings) and returns any Clojuredocs entry found, otherwise `null`.
```
ericdallo, line 21:
```suggestion
  - Add new LSP custom method `clojure/clojuredocs/raw` which takes a symbol and a namespace (both strings) and returns any Clojuredocs entry found, otherwise `null`.
```

ericdallo:
Thank you for the contributtion!

CHANGELOG.md
PEZ: Good catch!

PEZ:
Now, is there a way I can check for this capability? `try/catch` seems rude...

ericdallo:
What do you mean with capability? @PEZ this is another clojure-lsp custom method which is already available if clojure-lsp is initialized

PEZ:
The `clojuredocs` one. Users of older clojure-lsp clients won't have this custom method. I'm thinking maybe the LSP protocol has a way to query for if a command is available? Otherwise, maybe we should add that to serverInfo, an entry with `customCommands` or something.

ericdallo:
I see, the LSP protocol has a client and server capabilities, when starting the client say what capabilities support and server respond with the ones it supports as well, but I'm not sure about custom capabilities.
It'd be awesome if we could use the same existing capabilities communication.
https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#serverCapabilities

PEZ:
I don't understand how to use it for this. But I can make a version check for now. 😄

ericdallo:
Yeah, we can go with the version check for now as it'll be necessary changes on clojure-lsp, I'll try to implement that soon and let you know how will be the API

PEZ:
> maybe you can use the clojure-lsp command cursorInfo which brings information about the symbol at point, and then get the ns from kondo, this way you don't need a repl to query clojuredocs :)

Nice. I didn't know what this command does. Very good to know!

PR #631, opened 2021-11-07
[docs] Freezing on MacOS

ikhaldeev:
clojure-lsp uses com.taoensso/timbre for logging, and by default, timbre tries to add hostname into log entry. on my MacOS it wasn't able to resolve the hostname which caused freezes. 

Issue 
Every minute or so request to lsp takes 5 seconds.

Solution
Add MacOS hostname to /etc/hosts

sudo sed -i bak "s^127\.0\.0\.1.*^127.0.0.1 localhost $(hostname)^g" /etc/hosts
sudo sed -i bak "s^::1.*^::1 localhost $(hostname)^g" /etc/hosts

Reference
https://github.com/ptaoussanis/timbre/issues/323
https://stackoverflow.com/questions/39636792/jvm-takes-a-long-time-to-resolve-ip-address-for-localhost/39698914

### DIFF
--- a/src/clojure_lsp/logging.clj
+++ b/src/clojure_lsp/logging.clj
@@ -7,7 +7,8 @@
 #_{:clj-kondo/ignore [:unresolved-var]}
 (defn setup-logging [db]
   (let [log-path (str (java.io.File/createTempFile "clojure-lsp." ".out"))]
-    (log/merge-config! {:appenders {:println {:enabled? false}
+    (log/merge-config! {:middleware [#(assoc % :hostname_ "")]
+                        :appenders {:println {:enabled? false}
                                     :spit (log/spit-appender {:fname log-path})}})
     (log/handle-uncaught-jvm-exceptions!)
     (swap! db assoc :log-path log-path)))


ericdallo:
oh, quite a bug..
Thanks for the PR, I wonder if we can change the timbre log to not add the hostname as it'd probably avoid those issues WDYT?

ikhaldeev:
I've checked the timbre log library and found that hostname is resolved at the very last moment. But unfortunately, there is no easy way to disable it.
Here is the code of default-output-fn from timbre
https://github.com/ptaoussanis/timbre/blob/master/src/taoensso/timbre.cljc#L47

I've implemented two different versions of the fix.
1. Using middleware we can set value for hostname_ and prevent any actual resolving of hostname.
https://github.com/ikhaldeev/clojure-lsp/commit/c76426b2e8e1ba577aaee01c2e3ae9cb3572dba8

2. Implement a custom output function. In this case, it is just a copy of the default one, without an option for disabling stacktraces (as clojur-lsp needs this) and without hostname.
https://github.com/ikhaldeev/clojure-lsp/commit/f7d5f625a8c3e175af4723d80a2a2bfbe3fe282a

ericdallo:
I see, I liked the first one, it looks simple, if that will remove the hostname and fix this issue, looks good to me, WDYT?

ikhaldeev:
Looks good to me as well. Should I create PR?
timbre are aware of the issue with resolving hostname, there is a reference to the issue in my first post.

ericdallo:
> timbre are aware of the issue with resolving hostname, there is a reference to the issue in my first post.
I see

> Looks good to me as well. Should I create PR?
Please, you can do on this PR

ericdallo (approved):
Thank you for this contribution!

PR #635, opened 2021-11-09
Add flake

ericdallo:
Co-authored-by: kkharji

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,58 @@
+{ lib, stdenv, graalvm11-ce, gcc, babashka, fetchurl, fetchFromGitHub, clojure
+, writeScript, jre, makeWrapper }:
+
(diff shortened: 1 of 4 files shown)


ericdallo:
The issue now is to make the jar build work, the issue is that we will need to make it work offline or in a separated derivation like nixpkgs does with maven cases

ericdallo:
@kkharji I almost fixed the jar build, it seems right now the issue is when compiling the java classes with `americano` via `clojure -X:javac`, it seems that `americano` checks network for some reason(?)
I think it almost there if we fix that

kkharji:
Oh nice, Thank you I was really stuck with jar thing, I will merge it to the original pr and do further testing, btw you are welcome to push commit to the same PR, I'll just make sure I won't force push

PR #643, opened 2021-11-17
Prevent potential macro expansion error

noprompt:
This patch prevents a potential error where passing more than two arguments to `logging-time` could result in a macro expansion error.

I spotted this while skimming through the code. All of the calls to `logging-time` currently pass no more than two arguments so this was never an issue.


ericdallo:

PR #644, opened 2021-11-18
use name of tap installed

drewverlee:


ericdallo:
@drewverlee actually this was on purpose, since there is a `clojure-lsp` one on homebrew-core that uses only JVM and not graalvm native image, so we recommend removing that if you have installed and then installing the official from this org: https://github.com/clojure-lsp/homebrew-brew

PR #645, opened 2021-11-20
Expose cursor info raw API

ikhaldeev:
Expose cursor info raw API to be used in the editors.

Sample use case - in emacs lsp-mode:
`lsp-clojure-dwim`
1. find a definition for the symbol
2. find references if the cursor is already on the definition
3. fallback to cider for finding runtime definitions
4. open code actions if no symbol under cursor (useful for threading actions when the cursor is on parentheses)

here is a naive implementation of 'lsp-clojure-dwim` that uses a new API method:
https://gist.github.com/ikhaldeev/eb82213ff3233158112ff29160b16283

I haven't found a way to utilize clojure/cursorInfo/log for this use case.

### DIFF
--- a/integration-test/integration/cursor_info_test.clj
+++ b/integration-test/integration/cursor_info_test.clj
@@ -1,13 +1,13 @@
 (ns integration.cursor-info-test
   (:require
-   [clojure.test :refer [deftest testing is]]
+   [clojure.test :refer [deftest is testing]]
    [integration.fixture :as fixture]
    [integration.helper :as h]
    [integration.lsp :as lsp]))
 
 (lsp/clean-after-test)
 
-(deftest definition
+(deftest cursor-info
   (lsp/start-process!)
   (lsp/request! (fixture/initialize-request))
   (lsp/notify! (fixture/initialized-notification))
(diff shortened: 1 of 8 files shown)


integration-test/sample-test/src/sample_test/cursor_info/a.clj
```diff
@@ -0,0 +1,6 @@
+(ns sample-test.src.sample-test.cursor-info.a)
+
+(defn some-public-func [a]
+  a)
+
+
```
ericdallo, line 6:
```suggestion
```

integration-test/integration/cursor_info_test.clj
```diff
@@ -0,0 +1,36 @@
+(ns integration.cursor-info-test
+  (:require
+   [clojure.test :refer [deftest testing is]]
+   [integration.fixture :as fixture]
+   [integration.helper :as h]
+   [integration.lsp :as lsp]))
+
+(lsp/clean-after-test)
+
+(deftest definition
```
ericdallo, line 10:
```suggestion
(deftest cursor-info
```

ericdallo (approved):
Thanks for the PR and integration test!
From a clojure-lsp point of view it makes sense to expose that, I just not sure what you are trying to solve on lsp-mode side 🤔 
Why existing lsp-mode methods are not enough for what you want?

ikhaldeev:
I want one command (and one keybinding) for all these actions. It might be possible to implement everything with existing commands, but it looks much easier using cursor info.

ericdallo:
@ikhaldeev lint is still failing, you can test it local with `make lint-clean` and `make lint-format`

ikhaldeev:
@ericdallo lint task is successful, but graalvm-integratin-test is failing with timeout and it was successful on previous runs. Should I fix something?

ericdallo:
@ikhaldeev the integration tests sometimes are flaky so it's not your changes

ericdallo:
Thank you!

PR #654, opened 2021-11-29
Add naive sort to clean-ns, through settings

snoe:
A config setting to override the difficult to replicate clojure style guide `:require` sorting with a `:naive` sort that will follow tools like unix `sort` command or cursive's sort lines command.

`"clean" {:sort {:require :naive}`

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -65,5 +65,9 @@
                                    :jvm-opts ["-Xmx2g" "-server"]
                                    :main-class clojure-lsp.main
                                    :aliases [:debug]}}
+           :run-debug {:extra-deps {nrepl/nrepl {:mvn/version "0.8.3"}
+                                    cider/cider-nrepl {:mvn/version "0.27.2"}}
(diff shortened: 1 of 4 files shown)


ericdallo:
@snoe I can't understand easily why we need that, can we add unit-tests for that? is `:naive` the better name for that? maybe alphabetically?

snoe:
> @snoe I can't understand easily why we need that, can we add unit-tests for that? is `:naive` the better name for that? maybe alphabetically?

We have team members who essentially `:<>!sort` namespaces in vim, or in cursive call the sort-lines command. The big difference is that namespaces without a surrounding vector will sort last. Which actually helps convey information as, in our project, those namespaces are required for things like defmulti or protocol extensions that only exist to be loaded.

I thought about `alphabetically` but that's inaccurate. I guess java calls it `lexicographic` which I'll switch to.

ericdallo:
I see!

`lexicographic` sounds better indeed

ericdallo:
I see!

`lexicographic` sounds better indeed

PR #671, opened 2021-12-17
Avoid high CPU and lockup when kondo throws exceptions

Cnly:
I have a snippet that would trigger an uncaught clj-kondo error and will further cause clojure-lsp to set my CPU on fire for a minute plus other functions (e.g. document highlighting) to stop working totally:

```clj
;; a.clj

(ns ^{:clj-kondo/config '{:linters
                          {:unresolved-symbol
                           {:exclude [(a/my-macro1 EXTRA [$])]}}}}  ;; Remove "EXTRA" and things will recover
  a)

(defmacro my-macro1 [a b c]
  (a b c))

(my-macro1 $ ? !)
```

LSP functions all stop working because the `:processing-changes` flag never gets set back to `false` when clj-kondo errors, and the `process-after-changes` macro spins to wait for that flag, causing high CPU until it times out. This PR fixes those problems. I've been away from the codebase for quite a while, so please point out if I'm not doing something right :)

I'll also make a patch for clj-kondo so the exception actually gets fixed.

### DIFF
--- a/src/clojure_lsp/handlers.clj
+++ b/src/clojure_lsp/handlers.clj
@@ -41,7 +41,9 @@
        (if (> (quot (- (System/nanoTime) ~'_time) 1000000) 60000) ; one minute timeout
          (log/warn "Timeout waiting for changes for body")
          (if (:processing-changes @db/db)
-           (recur)
+           (do
+             (Thread/sleep 100)
+             (recur))
            ~@body)))))
 
 (defn ^:private report-startup-progress
--- a/src/clojure_lsp/handlers.clj
+++ b/src/clojure_lsp/handlers.clj
@@ -37,13 +37,13 @@
 
 (defmacro process-after-changes [& body]
   `(let [~'_time (System/nanoTime)]
-     (loop []
+     (loop [backoff# 1]
        (if (> (quot (- (System/nanoTime) ~'_time) 1000000) 60000) ; one minute timeout
          (log/warn "Timeout waiting for changes for body")
          (if (:processing-changes @db/db)
            (do
-             (Thread/sleep 100)
-             (recur))
+             (Thread/sleep backoff#)
+             (recur (min 200 (* 2 backoff#)))) ; 2^0, 2^1, ..., up to 200ms
            ~@body)))))
 
 (defn ^:private report-startup-progress
--- a/.cljfmt.edn
+++ b/.cljfmt.edn
@@ -3,7 +3,8 @@
  :remove-consecutive-blank-lines? true
  :insert-missing-whitespace? true
  :align-associative? false
- :indents {catch-kondo-errors [[:inner 0]]}
+ :indents {#re "^(?!catch-kondo-errors).*" [[:block 0]]
+           catch-kondo-errors [[:inner 0]]}
  :test-code
  (comment
    (:require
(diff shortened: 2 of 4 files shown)


Cnly:
OK I think this is interesting. CI is failing for something I haven't touched I guess... and the results on [my repo](https://github.com/Cnly/clojure-lsp/runs/4560629168?check_suite_focus=true) also differ what's shown [here](https://github.com/clojure-lsp/clojure-lsp/runs/4560672488?check_suite_focus=true). I saw commits on master also have similar problems... @ericdallo any idea?

src/clojure_lsp/handlers.clj
```diff
@@ -41,7 +41,9 @@
        (if (> (quot (- (System/nanoTime) ~'_time) 1000000) 60000) ; one minute timeout
          (log/warn "Timeout waiting for changes for body")
          (if (:processing-changes @db/db)
-           (recur)
+           (do
+             (Thread/sleep 100)
```
ericdallo, line 45: I wonder how much this will affect all features that use this macro, should this be something smaller? not sure what value, but 100ms sounds a little too much for some features?

Cnly: That's an ad hoc number indeed. What's your suggestion? Maybe it's better to have an incremental backoff?

ericdallo:
@Cnly there is one or other flaky tests (It'd be great to fix those but they are not easy to fix)

src/clojure_lsp/handlers.clj
ericdallo: yeah, something exponential sounds better, just not sure what buckets should we use

ericdallo: we could leave the 200 as max bucket and start with little buckets of 50? not sure it worths though

Cnly: It looks like we didn't have problems when there's no backoff, so I think we can simply start from 2^0, 2^1, ... and set 200ms as max.

ericdallo: sounds like a good start

src/clojure_lsp/handlers.clj
```diff
@@ -37,13 +37,13 @@
...
-     (loop []
+     (loop [backoff# 1]
        (if (> (quot (- (System/nanoTime) ~'_time) 1000000) 60000) ; one minute timeout
          (log/warn "Timeout waiting for changes for body")
          (if (:processing-changes @db/db)
            (do
-             (Thread/sleep 100)
-             (recur))
+             (Thread/sleep backoff#)
+             (recur (min 200 (* 2 backoff#)))) ; 2^0, 2^1, ..., up to 200ms
```
ericdallo, line 46: is there any chance we blow up the maximum value allowed for a long on backoff? I was just wordering, but probably only if we waited too many time which is acceptable I think

ericdallo: oh not really, because of min, seems safe

Cnly: Yep, should be safe

ericdallo:
Could you also add a entry on CHANGELOG?

ericdallo:
I'll add to the changelog, thank you for the help!

Cnly:
Okay thanks, I'm still looking into the other PR as it's having some problems :)

PR #672, opened 2021-12-17
Allow absolute paths in deps.edn :local/root

Cnly:
Previously if I have something like

```clj
clj-kondo/clj-kondo {:local/root "/Users/me/git/clj-kondo"}
```

as a dependency in `deps.edn`, lsp will fail with

```
java.lang.IllegalArgumentException: /Users/me/git/clj-kondo is not a relative path
```

This PR fixes that issue.

### DIFF
--- a/src/clojure_lsp/source_paths.clj
+++ b/src/clojure_lsp/source_paths.clj
@@ -80,7 +80,8 @@
                            (remove nil?))
           deps-files-by-local-root (->> local-roots
                                         (map (fn [local-root]
-                                               (let [sub-deps-file (io/file root-path local-root "deps.edn")]
+                                               (let [local-root (shared/relativize-filepath local-root root-path)
+                                                     sub-deps-file (io/file root-path local-root "deps.edn")]
                                                  (when (shared/file-exists? sub-deps-file)
                                                    [local-root sub-deps-file]))))
                                         (remove nil?))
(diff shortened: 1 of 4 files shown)


ericdallo:
@Cnly d you think it's related with this issue?
https://clojurians.slack.com/archives/CPABC1H61/p1639747789213400

src/clojure_lsp/source_paths.clj
```diff
@@ -80,7 +80,8 @@
                            (remove nil?))
           deps-files-by-local-root (->> local-roots
                                         (map (fn [local-root]
-                                               (let [sub-deps-file (io/file root-path local-root "deps.edn")]
+                                               (let [local-root (shared/relativize-filepath local-root root-path)
+                                                     sub-deps-file (io/file root-path local-root "deps.edn")]
```
ericdallo, line 84: Could you create a test for that?

ericdallo: seems like CI is failing because of this change

Cnly: Right, should be fixed now with a new testcase.

ericdallo (approved):

Cnly:
Should I also add a changelog entry for this one?

ericdallo:
Thank you very much for the help!

PR #673, opened 2021-12-17
fix: make local/root paths relative to deps.edn

hugoduncan:
When resolving local/root dependencies, a realtive path needs to be with
respect to the deps.edn file that contains the dependency.

### DIFF
--- a/src/clojure_lsp/source_paths.clj
+++ b/src/clojure_lsp/source_paths.clj
@@ -49,6 +49,15 @@
          set
          (set/union root-source-paths))))
 
+(defn relative-to-deps-file [deps-file]
+  (let [p (.getParentFile (io/file deps-file))]
+    (fn [f]
+      (if (.isAbsolute (io/file f))
+        f
+        ;; NOTE; calling .getCanonicalFile on this would get of some "../" path
+        ;; components.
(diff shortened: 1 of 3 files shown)


src/clojure_lsp/source_paths.clj
```diff
@@ -49,6 +49,13 @@
          set
          (set/union root-source-paths))))
 
+(defn relative-to-deps-file [deps-file]
+  (let [p (.getParentFile (io/file deps-file))]
+    (fn [f]
+      (if (.isAbsolute (io/file f))
+        f
+        (shared/normalize-file (io/file p f))))))
+
```
ericdallo, line 58: We can't have reflections because of graalvm, also would you mind changing to:
```clojure
(defn ^:private relative-to-deps-file [file ^java.io.File deps-file]
  (if (.isAbsolute (io/file file))
    file
    (shared/normalize-file (io/file p f))))
```
and use the function on a anonymous function on the caller?

ericdallo:
I'll merge and add to changelog + the minor change on the function + reflection fix
Thank you very much for the fix!

hugoduncan:
You beat me to it - thank you!

PR #674, opened 2021-12-18
Add babashka pod

### LINKED ISSUE #555: Consider exposing API via babashka pod
When doing so, we can call the API directly from babashka scripts.

More info on how clj-kondo does this: https://github.com/clj-kondo/clj-kondo/blob/master/src/pod/borkdude/clj_kondo.clj

ericdallo:
Fixes #555


src/pod/clojure_lsp/api.clj
```diff
@@ -0,0 +1,135 @@
...
+      (when-not (identical? ::EOF message)
+        (let [op (-> (get message "op")
+                     read-string
+                     keyword)
+              id (some-> (get message "id")
+                         read-string
+                         (or "unknown"))]
+          (case op
+            :describe (do (write {"format" "transit+json"
+                                  "namespaces" [{:name "borkdude.tdn.pod"
```
borkdude, line 98: There is a borkdude.tnd.pod from tools-deps-native here. Probably want to rename that.

borkdude:
Cool stuff!

src/pod/clojure_lsp/api.clj
ericdallo: Oh, thanks!

PR #685, opened 2022-01-07
Add move coll entry code action + refactor

ericdallo:
Co-authored-by: @mainej Jacob Maine

- Add `Move coll entry down` and `Move coll entry up` code actions
- Add `move-coll-entry-down` and `move-coll-entry-up` commands

move-coll-entry


tami5:
😍

PR #686, opened 2022-01-08
 Move coll entry

mainej:
- Allow movement of destructured bindings
- Treat commas as whitespace
- Add tests, including several commented out failing tests.

TODO
- [ ] Remove support for sets? 
   - @ericdallo you wrote `can-move-entry?` so I wanted to check with you on this. Since the elements of a set are never paired, I don't think this refactoring should be enabled for them. What do you think?
- [ ] Fix movement in single line maps. See failing tests.
- [ ] Fix movement when cursor starts on trailing comment. See failing tests.
- [ ] Fix movement of commas. See failing tests.
- [ ] Changelog
- [ ] Docs

I have a plan for several of the failing tests, but the fix I have in mind will be a reasonably large change to the central algorithm. I won't be surprised if that change makes some other tests fail, so it may take a while to get everything working again.

### DIFF
--- a/src/clojure_lsp/feature/move_coll_entry.clj
+++ b/src/clojure_lsp/feature/move_coll_entry.clj
@@ -2,10 +2,25 @@
   (:require
    [clojure.string :as string]
    [rewrite-clj.node :as n]
-   [rewrite-clj.zip :as z]
-   [taoensso.timbre :as log]))
+   [rewrite-clj.zip :as z]))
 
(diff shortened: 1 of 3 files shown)


ericdallo:
> Since the elements of a set are never paired, I don't think this refactoring should be enabled for them. What do you think?

Makes sense to me, we can remove this feature for sets so.

mainej:
> remove this feature for sets

👌 I'll do that in a separate PR.

PR #688, opened 2022-01-10
Add features to move-coll-entry

mainej:
This PR is a re-write of move-coll-entry to use a different method of allocating comments to entries. It now supports several features that were not possible before:
1. Move entries terminated by commas
        {1 2,
         3 4}
         ->
        {3 4,
         1 2}
2. Move entries within single line maps
   * `{1 2 3 4} -> {3 4 1 2}`
   * `{1 2, 3 4} -> {3 4, 1 2}`
3. Move entry when starting on trailing whitespace
        {1 2 ;; two |comment
         3 4}
         ->
        {3 4
         |1 2 ;; two comment}

It also does a better job of moving the cursor, which is now reliably placed at the beginning of the entry in its new location.

As a separate matter, this PR removes support for moving entries within sets. In the future it may be possible to add support for moving items one at a time (as opposed to in pairs) within sets, vectors and lists. And the comment allocation in this PR should facilitate that. But, for now, it is not supported.

@ericdallo I found a few cases where can-move-*? returns the wrong results. This happens:

* When on a padding line after the first pair, and `move-up` is invoked.
* When on a padding line before the last pair, and `move-down` is invoked.

Movement should not be allowed in either of these situations, but can-move-*? reports differently. Unfortunately, there's no quick way to know we're in one of these situations without doing most of the work of allocating whitespace. I changed the code to return nil in these cases. Is that the right thing to do to make the action a no-op? Or should I return the original unchanged zloc somehow?

### DIFF
--- a/src/clojure_lsp/feature/move_coll_entry.clj
+++ b/src/clojure_lsp/feature/move_coll_entry.clj
@@ -111,19 +111,20 @@
           (let [;; find the start of the next padding
                 zloc    (->> elem-loc
                              z/right*
-                             ;; move right until the first newline, comma or key
+                             ;; seek to the first newline, comma or regular node
                              ;; TODO: handle case where comma precedes comment
                              ;; `a 1, ;; one comment`
                              ;; Conceptually, the comma should stay put while
                              ;; the entry and comment move.
                              (z/skip z/right* (comp #{:whitespace :comment}
                                                     z/tag))
+                             ;; relinquish that node
                              z/left*
                              ;; then give up as much whitespace as possible,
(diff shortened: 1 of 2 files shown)


ericdallo:
>  I changed the code to return nil in these cases. Is that the right thing to do to make the action a no-op?
Yes, right thing, this function should be fast and return any thruthy value since we run everytime during code actions processing when user moves the cursor

ericdallo (approved):
Thank you!

ericdallo:
I intend to do some final tests on the original branch tonight and then I merge to master, thank you for the huge help!

PR #698, opened 2022-01-16
Test tree on clojure lsp initialization

### LINKED ISSUE #697: Find every test in the project by default at init to fulfill test tree managers

As the new feature of test-tree for Calva was released it was noticeable that the test tree is only populated after open some files, and only with the tests from opened files.

Maybe it would be great to do a quick scan in the project to detect every test after the initialization of clojure-lsp.

erickisos:
Closes #697


src/clojure_lsp/crawler.clj
```diff
@@ -216,5 +219,8 @@
         (stubs/generate-and-analyze-stubs! settings db))
       (producer/publish-progress (:producer @db) 95 "Analyzing project files" progress-token)
       (log/info "Analyzing source paths for project root" root-path)
-      (analyze-source-paths! (:source-paths settings) db))
+      (analyze-source-paths! (:source-paths settings) db)
+      (producer/publish-progress (:producer @db) 97 "Analyzing test files" progress-token)
```
ericdallo, line 223: I think we should check for `when-not (:api? @db)` and we should not this on this same thread or we will block the clojure-lsp startup time, maybe just move this to [here](https://github.com/clojure-lsp/clojure-lsp/pull/698/files#diff-fa419531b6cbafb2899aa71bcd5a5acee775a79480a43235c6f29fc5074cb1b6R215) would solve both issues, WDYT?

src/clojure_lsp/crawler.clj
```diff
@@ -49,21 +49,24 @@
     (swap! db update :findings merge (group-by :filename (:findings result)))
     analysis))
 
+(defn ^:private analyze-test-paths! [paths db]
+  (map #(producer/refresh-test-tree % db) paths))
```
ericdallo, line 53: I think we should pass a URI not a filename to this function, so probably we need to call `shared/filename->uri` before calling the producer.

ericdallo: Hum, the paths are not necessarly filenames, right? they could be `/path/to/src/` and not `/path/to/src/my-file.clj`, and this would make the the test tree not work.
We should re-think a way to get all the URIs of the project, not only the paths.
Probably find it via clj-kondo analysis filtering project analysis only

ericdallo:
Thank you for the help @erickisos!

Just some points to make this ready to merge beside the comments:

- I merged a huge work I was doing that splitter the project into API and CLI, #696, you will probably need to rebase with master, sorry for not telling that before
- It'd be great if we could make a test for that, I don't see a way to unit test that easily, but I think we can create a integration test that uses bb under the hood to test that after the project started we should listen and wait for a testTree notification without `didOpen` any file, WDYT?
- Could you add an entry to CHANGELOG?

src/clojure_lsp/crawler.clj
ericdallo: > we should not this on this same thread or we will block the clojure-lsp startup time

I just realized there is a already a go block inside the producer call, so it's fine

erickisos:
Of course @ericdallo , sorry, I forgot the "draft" in the PR, however let me check those points then!

src/clojure_lsp/crawler.clj
erickisos: That's right, I was thinking if we had something like that in the project, but based on your comment it sounds like we don't have that feature at the moment, I will check if we can get that list of filenames from clj-kondo.

Thx Greg!

ericdallo: @erickisos probably something like:
```clojure
(->> (:analysis @db)
       (map :filename)
       (filter #(string/starts-with "file:/"))) ;; filter project only filenames
```

erickisos: just in case, I tried with one similar approach, and checking around with something like this
```clojure
(:documents @db)
```

And it seems like at least kondo and clojure-lsp don't keep track of all the files from the project.

WDYT about adding that kind of function in clojure-lsp? Or could be better if I add that in clj-kondo? I mean, we can just keep that in the clojure-lsp crawler.

ericdallo: the `:documents` are filled with the current opened documents on the editor and information about version and content, I think the best would be to use kondo analysis for that, which is more reliable, something like this should work :
```clojure
(-> (:analysis @db/db)
      (q/filter-project-analysis db/db)
      keys)
```

lib/src/clojure_lsp/crawler.clj
```diff
@@ -50,21 +50,24 @@
     (swap! db update :findings merge (group-by :filename (:findings result)))
     analysis))
 
+(defn ^:private analyze-test-paths! [paths db]
+  (map #(producer/refresh-test-tree % db) paths))
```
ericdallo, line 54:
```clojure
(defn ^:private analyze-test-paths! [paths db]
  (let [project-files (-> (:analysis @db/db)
                          (q/filter-project-analysis db/db)
                          keys)]
    (->> project-files
         (map #(shared/filename->uri % db))
         (mapv #(producer/refresh-test-tree % db)))))
```

ericdallo:
@erickisos I ended up committing this feature on master since we'd need to do more changes besides those mentioned, thanks for your huge help anyway!

Issue #684, opened 2022-01-04
Move map and bindings entries


mainej:
When I'm editing a map I often want to move one key/value pair before the prior pair (or after the next pair). Similarly, in a let binding, I often want to move a binding form and its value before the prior form (or after the subsequent form). This is a rote transformation, but it often takes many steps. This is especially true if the first or last pair is changing, because line-wise changes in the editor tend to break the structure of the map or binding, because they move the opening or closing brackets.

For example, if I'm on the `:b 2` line in this form:

```clojure
            {:a 1
             :b 2
             :c 3}
```
I'd like to be able to move it up, like so:
```clojure
            {:b 2
             :a 1
             :c 3}
```
Or down, like so:
```clojure
            {:a 1
             :c 3
             :b 2}
```

Ideally, comments would stay "affiliated" to their respective pairs. For example, consider the comments related to `:b` below, both on the line before it and on the same line, at the end.
```clojure
            {:a 1 ;; comment after a
             ;; comment before b
             :b 2 ;; comment after b
             ;; multiline
             ;; comment for c
             :c 3}
```
This would become (moving up):
```clojure
            {;; comment before b
             :b 2 ;; comment after b
             :a 1 ;; comment after a
             ;; multiline
             ;; comment for c
             :c 3}
```
Or (moving down):
```clojure
            {:a 1 ;; comment after a
             ;; multiline
             ;; comment for c
             :c 3
             ;; comment before b
             :b 2 ;; comment after b
             }
```

> Note an edge case demonstrated in the last example: because of the comment `;; comment after b`, the closing bracket of the map needs to be put on its own line, to avoid being commented out.

Would you be willing to add this refactoring to clojure-lsp? Or is there an established way to extend the code actions that clojure-lsp exposes?

I have a prototype that does the described refactoring. It preserves affiliated comments, as described above. It uses `clj-rewrite` and operates on strings, but doesn't have a full editor integration which is why I'm asking here. I'm not sure it's very idiomatic use of `clj-rewrite`, but it at least proves it's possible. Anyway, I'd be happy to share that code.

ericdallo:
@mainej It's been on my mind to add this feature for some time but I never had the time, the issue with that is the comments manipulation, if you have a code that provides this feature and want to share, I'd make things a lot easier :)

mainej:
@ericdallo I'm happy to share... see below. 

[...]

PR #701, opened 2022-01-18
Extend move-coll-entry to elementwise movements

mainej:
Originally `move-coll-entry` moved elements in vectors and maps in a pairwise fashion. This extends `move-coll-entry` so that it will move element-wise instead.  That is, it will move a pair of elements in some situations but a single element in others. It tries to be smart about which situation it is being called in. It will move pairwise in maps and vectors that establish bindings, and element-wise in sets, lists and regular vectors.

As before, it does its best to preserve comment affiliation. Comments will be kept near the element as it moves, including comments that appear on the lines before an element or a comment at the end of the same line as the element.

/cc @snoe, who originally suggested this addition to move-coll-entry in #684.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
   - Improve 'create function' refactor code action handling multiple cases. #682
   - Fix 'resolve macro as ...' code action not working.
   - Fix `showDocumentRequest` issues when triggered via some refactor code action.
-  - Add new code actions + commands `Move coll entry down` and `Move coll entry up` to move map/vectors pair entries. #684
+  - Add new code actions + commands `Move coll entry down` and `Move coll entry up` to move entries within collections. #684, #701
 
 - API/CLI
   - Make `format`, `clean-ns` and `rename` features not need to scan whole classpath, analyzing only project code improving performance a lot.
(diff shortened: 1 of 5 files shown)


mainej:
FWIW, I could have implemented this as separate actions: two for pairwise up/down and two for element-wise up/down, which would mean four keybindings. Having separate keybindings is more precise, but I'd rather commit one pair of keybindings to muscle memory, which is why I made element-wise movement an extension of move-coll-entry. There's a small chance it won't cover every use-case (like swapping keys with values) but personally I'm OK with doing that the old fashioned way.

ericdallo:
@mainej sounds great to me, thank you! I'll take a look tonight and test it

mainej:
@ericdallo I've pushed a few clean-ups to this PR. Since doing that one of the integration tests fails intermittently. I _think_ it's just a flaky test, but it's possible something I changed broke it. Let me know if I can help track down the problem.

ericdallo:
@mainej, unfortunately, we have one or other flaky integration-test, but usually re-runing fix it.
Also, I'll test your changes right now

ericdallo:
Thank you for the huge contribution!
Looks good to me and I didn't find any issues manually testing it

mainej:
Great, thanks @ericdallo! And thanks for fitting this review in between the big refactoring you have going on. Looks like lots of changes!

ericdallo:
Yes, I hope that help in the future we break the project or separate features for a more modular project :)

PR #702, opened 2022-01-20
Add support for classpath file lookup.

ccidral:
This is my first shot at implementing [nREPL sideloader support in Calva](https://github.com/BetterThanTomorrow/calva/issues/1451), although this might be used by any other editors and IDEs that want to follow this approach to sideloading. The idea here is that Calva (or any other editor) relies on clojure-lsp for loading files (classes or resources) from the classpath in order to send them across to the nREPL session. Please let me know if you think there's a better approach.

This PR adds a jsonrpc endpoint for looking up a file in the classpath. It accepts a file path as argument, tries to find the file in the classpath, and then returns its binary contents in base64 format, or nil if not found. First it looks into jar files, and then into the source paths if none of the jars contain the requested file. Perhaps this lookup order should be reversed because source paths appear before jar files in the classpath?

On the first time the endpoint is called, it builds an index containing all files found in jars from the classpath. Initially the indexed files point to their containing jar files, and then they get "expanded" with their actual binary contents as they are requested. This is a naive attempt to speed up lookups, so there might be better ways to accomplish that. Please see more details about the index at `classpath_lookup.clj`.

### DIFF
--- a/cli/resources/META-INF/native-image/clojure-lsp/clojure-lsp/reflect-config.json
+++ b/cli/resources/META-INF/native-image/clojure-lsp/clojure-lsp/reflect-config.json
@@ -33,6 +33,13 @@
   "methods":[{"name":"<init>","parameterTypes":[] }]
 },
 {
+  "name":"clojure_lsp.feature.classpath.ClasspathLookupParams",
+  "allDeclaredMethods":true,
+  "allPublicMethods": true,
+  "allDeclaredFields":true,
+  "methods":[{"name":"<init>","parameterTypes":[] }]
+},
+{
   "name":"clojure_lsp.feature.clojuredocs.ClojuredocsParams",
   "allDeclaredMethods":true,
   "allPublicMethods": true,
--- a/cli/src-java/clojure_lsp/ClojureLanguageServer.java
+++ b/cli/src-java/clojure_lsp/ClojureLanguageServer.java
@@ -11,6 +11,7 @@ import clojure.java.api.Clojure;
 import clojure.lang.IFn;
 
 import clojure_lsp.feature.cursor_info.CursorInfoParams;
+import clojure_lsp.feature.classpath.ClasspathLookupParams;
 import clojure_lsp.feature.clojuredocs.ClojuredocsParams;
 
 /**
@@ -44,5 +45,8 @@ public interface ClojureLanguageServer extends LanguageServer {
 
     @JsonRequest("clojure/clojuredocs/raw")
     CompletableFuture<Object> clojuredocsRaw(ClojuredocsParams clojuredocsParams);
+
+    @JsonRequest("clojure/classpath/lookup")
+    CompletableFuture<Object> lookupClasspath(ClasspathLookupParams params);
 
 }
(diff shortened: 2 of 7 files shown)


ericdallo:
@ccidral The implementation looks correct, although I feel like we should discuss a little more this need. 
We don't have any REPL code or anything related on clojure-lsp, although this PR just provides classpath info for downstream usages as they want. The problem we want to solve seems a little specific for Calva maybe, have you considered getting clojure-lsp classpath info on calva and crawling and caching jar entries on calva? This info is already available for calva via `serverInfo/raw` request.

ccidral:
Thanks for the feedback! At first I considered doing that in Calva but then I thought maybe other editors or IDEs could benefit from this functionality instead of replicating it themselves. Perhaps this problem transcends Calva as any editor or tool that wants to implement nREPL sideloading, or even retrieve classes and resources in general for any reason, will need this capability. Of course if classpath lookup is beyond the scope of what clojure-lsp is supposed to provide, then it makes sense to do it in Calva.

ericdallo:
Well, classpath lookup is indeed something clojure-lsp does and I think it's ok provide that info, but do we need to add all that jar index logic instead of providing just the classpath string standard that we already have? or do you think this is not enough for most cases?

ccidral:
The purpose of the index is to prevent clojure-lsp from searching, reading and extracting jar files every time a file lookup is requested. We can definitely remove it but I believe it would impose a significant overhead on lookup requests. As I commented on the code, the index is a naive attempt to optimize for performance, so there's probably a better ways of optimizing.

I was just thinking here, maybe classpath "lookup" is not the best term and is causing some confusion. What the lookup does is it returns the contents (bytes) of a file from the classpath, so it's not just "looking up" a file, it's really loading the requested file and returning its contents. When the file is found in a jar, it extracts the file from the jar.

ericdallo:
> I was just thinking here, maybe classpath "lookup" is not the best term and is causing some confusion. What the lookup does is it returns the contents (bytes) of a file from the classpath, so it's not just "looking up" a file, it's really loading the requested file and returning its contents. When the file is found in a jar, it extracts the file from the jar.

Good point, that what I was referring to, maybe we should rename to `lookup-classpath-content` since we already call "classpath lookup" in some other places. 

Anyway, I don't see any strong reason to not add this on clojure-lsp so, but would you mind adding unit and integration-tests for that feature? otherwise, we can break it any time and don't know since only downstreams would be probably using this feature

ccidral:
I'm a bit concerned about memory consumption though, the index may eat up a big chunk of memory if the classpath has lots of jars. I'll keep an eye on that. We might need a more efficient jar index implementation in the future.

> would you mind adding unit and integration-tests for that feature? otherwise, we can break it any time and don't know since only downstreams would be probably using this feature

Absolutely!

ericdallo:
> I'm a bit concerned about memory consumption though, the index may eat up a big chunk of memory if the classpath has lots of jars. I'll keep an eye on that. We might need a more efficient jar index implementation in the future.

Yes, it's good to keep an eye on that indeed

ericdallo:
Closing this as it's pretty obsolete by now, LMK if you wanna continue in a new PR

PR #707, opened 2022-01-20
Move coll entry tests

mainej:
This adds a few tests for element-wise movement and a few to document discrepancies between `can-move-*?` and `move-*`.

It also shortens the syntax for writing tests about movement from comments and whitespace.


ericdallo (approved):

ericdallo (approved):
Looks good! thank you!

PR #713, opened 2022-01-22
Fix problems

borkdude:


borkdude:
Forget this one, use https://github.com/clojure-lsp/clojure-lsp/pull/714

PR #714, opened 2022-01-22
Fix dependency problems which causes clj-kondo to loop forever

borkdude:


ericdallo (approved):

ericdallo:
Thank you very much for the help!

PR #716, opened 2022-01-23
When parsing, allow zloc to be whitespace or comment

mainej:
Before this change, when parsing a file's contents into a rewrite-clj
zipper, the returned zloc was never a comment or whitespace node. That
is, it was always sexpr-able. If the parser was invoked while the cursor
was on a comment or whitespace node, the zipper was moved to the z/next
sexp following the comment or whitespace, if any.

After this change, the zloc's node may be whitespace or a comment. This
preserves the cursor's position more accurately, and in turn, allows
move-coll-entry to move an entry affiliated with a comment or
whitespace.

When I was initially researching this change I expected many test
failures. I figured many refactorings would expect to start at an
sexpr-able node. As the diff shows, there were actually only a few
failing tests, and a correspondingly small number of implementation
fixes:

1. The tests for the parser itself needed to be adjusted slightly
2. When starting on whitespace, the thread-all refactorings expected to
   be on a non-whitespace/comment node. This behavior was easily
   restored.

So, at the moment, this patch is pretty simple. However, I'm a bit
concerned that most tests don't use the parser; they work at a lower
level, where they are handed a non-whitespace/comment node. So, there's
a chance that other refactorings will break.

We can either accept this possibility, sprinkling in `(z/skip-whitespace
z/up zloc)` when we find problems. Or we can do a little extra work to
make two nodes available to `f.refactor/call-refactor` and
`f.code-actions/all`, both `zloc` as before, a non-whitespace/comment
node, and a new node `ws-zloc` that may be whitespace or a comment. Then
individual actions could opt into the node they need.


lib/test/clojure_lsp/feature/move_coll_entry_test.clj
```diff
@@ -10,12 +11,11 @@
...
 (def ^:private uri (h/file-uri "file:///a.cljc"))
 
-(defn zloc-at [pos]
+(defn zloc-at [[row col]]
   (-> @db/db
       (get-in [:documents uri])
       :text
-      (z/of-string {:track-position? true})
-      (z/find-last-by-pos pos)))
+      (parser/loc-at-pos row col)))
```
mainej, line 18: This looks like a small change, but is actually important. By going through the parser, it proves that move-coll-entry works even when starting on whitespace/comment, without needing `:track-position?`. This is the same codepath that `f.refactor/call-refactor` and `f.code-actions/all` follow.

ericdallo:
Thanks for this, it's something we would really need to do some time.

> So, at the moment, this patch is pretty simple. However, I'm a bit
concerned that most tests don't use the parser; they work at a lower
level, where they are handed a non-whitespace/comment node. So, there's
a chance that other refactorings will break.

Yes, unfortunately, we are not used to testing comments on most of the cases, we should test it more indeed IMO.

> We can either accept this possibility, sprinkling in (z/skip-whitespace z/up zloc) when we find problems. Or we can do a little extra work to
make two nodes available to f.refactor/call-refactor and
f.code-actions/all, both zloc as before, a non-whitespace/comment
node, and a new node ws-zloc that may be whitespace or a comment. Then
individual actions could opt into the node they need.

Not sure would be worth it, handling the comments/whitespace with the default parser would be ideal and each feature could handle it specifically for the best UX.

Since this is a small but critical change, I'll have to test each refactoring/feature from https://clojure-lsp.io/features/ to make sure we didn't break anything 🤷🏻

mainej:
> I'll have to test each refactoring/feature from https://clojure-lsp.io/features/

That's... a lot of features. I can help if you'd like. Do you want to split them up somehow?

ericdallo:
@mainej thanks, if you manage to test the most or the major refactor commands, that should help a lot already: https://clojure-lsp.io/features/#refactoring :)

mainej:
> https://clojure-lsp.io/features/#refactoring

I went through these refactorings and didn't notice any explicit errors when invoking them. There's some slightly odd behavior in some cases:

## Invoking from comment

- cycle privacy: If comment was to right of `^:private`, removes comment. Also happens when cursor starts anywhere else in function, and was probably always like this.
- extract function: Creates function whose body is the comment
- create private function: Emacs doesn't offer this. Is it the same as extract function? Should it be removed from docs?
- introduce let: Creates `(let [x x])` binding, though val 'x' doesn't exist
- move to let: Creates `x ;; comment` binding, with no value

## Invoking from whitespace

- extract function: Creates function whose body is the whitespace
- create private function: Emacs doesn't offer this. Is it the same as extract function? Should it be removed from docs?
- introduce let: Creates `(let [x x])` binding, though val 'x' doesn't exist
- move to let: Creates `x` binding, with no value

## Elsewhere:
- thread first all: Deletes interior comments (probably an unrelated bug)
- thread last all: Deletes interior comments (probably an unrelated bug)

For the purposes of this PR, I propose modifying extract-function, introduce-let and move-to-let, so that they operate on a real node, not the comment/whitespace node. Would you expect them to choose the outer node `z/up`? I think that would be closest to the original behavior. Or perhaps the next node? That might be more intuitive. I'm not sure whether to use `z/next` or `z/right` for that, but I could figure it out.

mainej:
Edit: Emacs _does_ offer create-function. It behaves reasonably.

mainej:
Patch updated: extract-function, introduce-let and move-to-let all operate on the next real node to the right, not on the comment node.

ericdallo:
Nice, I was expecting we would need to do one or other changes regarding existing refactors, glad you already debugged it! 

> Patch updated: extract-function, introduce-let and move-to-let all operate on the next real node to the right, not on the comment node.

Sounds good! not sure if we should use z/right or z/next, I think it depends on the refactor? not sure, that's why I think we should check if that works for all cases otherwise change it.
Your debugging list already sounds like we are pretty good on the refactoring side, I think we can check other features like completion, references, rename and so on

mainej:
>  not sure if we should use z/right or z/next, I think it depends on the refactor

Yeah, I agree, depends on the refactor.

I did some further testing and had to make a revision: extract-function, introduce-let and move-to-let all operate on the next real node to the right _if there is one_, otherwise, they operate on the next node up. If that node doesn't exist (i.e., you're at the end of the file), they'll throw an exception. Looking at the code, that's probably always been the case. If you'd like, I could make it a bit more defensive to avoid those errors.

I've also started paying attention to the logs, which I should have been doing from the beginning. That pointed out a problem with completion, where errors were thrown when the code tried to complete from a comment. That's fixed in 1a0db27.

ericdallo:
>  Looking at the code, that's probably always been the case. If you'd like, I could make it a bit more defensive to avoid those errors.

If the code to fix that would not be complex, looks good to me to handle this corner case.

> I've also started paying attention to the logs, which I should have been doing from the beginning. That pointed out a problem with completion, where errors were thrown when the code tried to complete from a comment. That's fixed in 1a0db27.

Nice, yeah, while developing clojure-lsp it's quite important to check the server logs for exceptions throw on async process which will show only on the logs, Also I recommend developing using a REPL on the running clojure-lsp process if you build it with just `make`, it should add cider-nrepl to the jar which you can use to jack in, I have a Emacs shortcut for that:
```elisp
(defun lsp-clojure-nrepl-connect ()
  "Connect to the running nrepl debug server of clojure-lsp."
  (interactive)
  (let ((info (lsp-clojure-server-info-raw)))
    (save-match-data
      (when-let (port (and (string-match "\"port\":\\([0-9]+\\)" info)
                           (match-string 1 info)))
        (cider-connect-clj `(:host "localhost"
                             :port ,port))))))
```

It request clojure-lsp `serverInfo` and get the cider-nrepl port and then connect the reply, so everytime you eval on clojure-lsp, it should affect the running clojure-lsp session 🤯 

Feel free to improve the emacs section [here ](https://clojure-lsp.io/development/#the-clojure-way) about this

mainej:
> handle this corner case.

Will do. I'll do more testing while tailing the logs to see what else I can find.

> I recommend developing using a REPL on the running clojure-lsp process

👍 I've been doing this for the last week or so; it's really helpful. I'll add the instructions and snippet you wrote to the docs. Maybe this one too:

```emacs-lisp
  (defun lsp-clojure-nrepl-tail-log ()
    "Tail the logs of the running nrepl debug server of clojure-lsp."
    (interactive)
    (let ((info (lsp-clojure-server-info-raw)))
      (save-match-data
        (when-let (file (and (string-match "\"log-path\":\"\\([^\"]+\\)" info)
                             (match-string 1 info)))
          (find-file file)
          (auto-revert-tail-mode)))))
```

All this setup is pretty powerful, but since I got it running I've been having problems with the editor stalling. It's not using excessive CPU or anything, it just stops responding. Have you seen that too? Have other people reported it, or is it only a problem when using the `make` version of the executable?

ericdallo:
@mainej never saw that I think, try the native bin if it happens later

mainej:
Apologies @ericdallo ... this PR has become a moving target. I'm still finding and fixing errors that happen when refactorings are invoked in unusual locations. I'll let you know when it's ready for a real review, although it's going to be quite large by then.

I ended up changing many of the tests to go through the parser. This makes it a lot easier to write tests about comments and whitespace because there's much less fiddling with zippers to get to the right spot. I think it's also better to see how the refactorings behave after passing through the parser, since that's how the real code path works. Anyway, that's the reason for many of the test changes, although there are also quite a few new tests.

ericdallo:
@mainej thank you for your efforts!
I think this kind of PR is the reason to have good integration tests with bb 😅
I can probably take a look later this week and test those missing features. Feel free to keep it updated and thanks again for the help!

mainej:
> good integration tests with bb

Ah, the dream ... fast integration tests! Have you been experimenting with bb for that?

I've been thinking about my ideal tests. I'd rather that they operate at more of an integration level. Right now in tests you construct chunks of code, turn them into zlocs, reposition the zipper, define cursor positions, pass all of that into a refactoring, and get back more zlocs and cursor positions. But in my ideal, the tests would reflect the way a user experiences clojure-lsp. You'd pass in a "workspace" (a series of "screens" each representing a file with its code and possibly a cursor position or positions), call a refactoring, and get back another workspace, with all the refactorings applied. The tests would assert that a particular refactoring transforms one workspace into another.

That gets into implementing the editor side of LSP—writing something that can apply the edits that clojure-lsp produces. Actually, perhaps lsp4j implements some of that already? It provides both a Server and a Client, but I haven't looked at what the Client can do.

I don't know... sounds like a lot of work just to make the tests marginally more declarative. It's always hard to commit to large investments in test infrastructure. Nice to think about though!

ericdallo:
> Ah, the dream ... fast integration tests! Have you been experimenting with bb for that?

Not sure I understood, [we do have](https://github.com/clojure-lsp/clojure-lsp/blob/master/cli/bb.edn#L3) integration tests with bb ATM :)

mainej:
Oh, I see... it is nice that those start quickly. And I see what you mean, they prove the full client/server interaction. They'll be helpful for testing those extra features.

But it's not really what I was looking for. I meant that I want the unit tests to be at a higher level. Consider the suppress-diagnostic tests, which look something like this:
```clojure
(let [results (-> (h/code "(defn foo [b]"
                          "  (+ c 1 2))")
                  z/of-string
                  (z/find-value z/next 'c)
                  (transform/suppress-diagnostic "unresolved-var"))]
  (h/assert-submaps
   [{:loc   (h/code "#_{:clj-kondo/ignore [:unresolved-var]}"
                    "  ")
     :range {:row 2 :col 3 :end-row 2 :end-col 3}}]
   (map #(update % :loc z/string)
        results)))
```

I'd rather write a test like this:
```clojure
(is (= (-> (editor/editor)
           (editor/with-file ["(defn foo [b]"
                              "  #_{:clj-kondo/ignore [:unresolved-var]}"
                              "  (+ |c 1 2))"]))
       (-> (editor/editor)
           (editor/with-file ["(defn foo [b]"
                              "  (+ |c 1 2))"])
           (editor/choose-action :suppress-diagnostic "unresolved-var"))))
```

The `editor/choose-action` command would check that the action was available, call it, splice the returned edits into the editor's files, add new files, and re-position the cursor. With something like that I'd be more confident that everything about a refactoring was working.

A naive version of an `editor` test helper might not be so hard. On the other hand, writing a full text editor is challenging, to say the least! You'd have to decide what level it works at—you could call directly into `clojure-lsp.feature.refactor/call-refactor`, but if you were ambitious, you could communicate across JSON-rpc. And you'd have to decide how much to support... you'd probably want to be able to display available actions, but would you want to represent code lenses, or symbol usages, or semantic tokens? At some point the test helper would become as complex as the real code.

Anyway, it's just an idea I've been toying with. I think it'd be nice, but there are many other things I'd want to work on first.

ericdallo:
@mainej nice idea, we actually kind of had these kind of tests not as high level as you meant on the handlers_test.clj, but then we found integration-tests would fix better and have unit tests for the specific features/small components rather a whole.. integration-test 😅 .
It's a idea, but to me looks like another set of integration-tests which we already can achieve with the bb tests

lib/src/clojure_lsp/feature/add_missing_libspec.clj
```diff
@@ -269,32 +279,35 @@
...
 (defn find-require-suggestions [zloc missing-requires db]
-  (when-let [sexpr (z/sexpr zloc)]
-    (if-let [alias-str (namespace sexpr)]
+  (when-let [sym (safe-sym zloc)]
+    (if-let [alias-str (namespace sym)]
       (find-alias-require-suggestions alias-str missing-requires db)
-      (find-refer-require-suggestions sexpr missing-requires db))))
+      (find-refer-require-suggestions sym missing-requires db))))
 
 (defn add-require-suggestion [zloc chosen-ns chosen-alias chosen-refer db]
```
ericdallo, line 287: Not sure I understood the changes of this function on this PR, why do we need this? Could you elaborate a little more on the issue we are solving here?

ericdallo: @mainej

lib/src/clojure_lsp/feature/add_missing_libspec.clj
```diff
@@ -269,32 +279,35 @@
...
+                                                      (not= chosen-alias (namespace sym)))))
+        chosen-require  (if chosen-alias
+                          (add-known-alias zloc (symbol chosen-alias) (symbol chosen-ns) db)
+                          (add-known-refer zloc (symbol chosen-refer) (symbol chosen-ns) db))]
+    (seq
+      (concat chosen-require
+             ;; TODO: the intention of the following code appears to be to
+             ;; convert un-aliased symbols to aliased symbols, e.g.
+             ;; `clojure.string/split` -> `my-str/split`. However, when on a
+             ;; `clojure.string/split` node, the UI will only ever suggest
```
ericdallo, line 300: It's possible to pass the alias if using the command manually, not via code action right?

ericdallo:
@mainej I tested most features and everything seems to be working, good job!
I'd like to just understand better the changes on add-suggestion-requires

ericdallo (approved):
I'm actually merging this as you didn't change the behavior of add-suggestion-require and this PR seems to add a lot of value, but please let me know what you think. Thank you for the help!

ericdallo:
I just noticed we didn't add a changelog entry, would you mind add it later?

mainej:
Thanks for merging this @ericdallo! I'll review and respond to your questions separately. If any problems come up, please let me know here or on Slack.

lib/src/clojure_lsp/feature/add_missing_libspec.clj
mainej: Yeah, my apologies... This should have changed the whitespace handling only, but it turned into more. I can revert some of it, if you'd like.

I think that I was confused about how the functions in this namespace work. There's a group that are dedicated to finding potential requires and imports (the code actions). And there's a group that inserts requires and imports into the ns (the commands). 

I _thought_ that the commands just took whatever the user gave them and inserted that into the ns (except for the "missing" commands, which magically figure out the correct require/import.) But that's not how most of them work. Instead, they take the user's input, then _check_ whether that input would have been suggested as a code action, and _then_ insert that into the namespace.

I can revert to that behavior, if you'd like. But, I still think that my initial impressions is valid. Shouldn't the commands do whatever the user says, even if that's not something that would have been suggested by the actions? That would let the user construct an invalid require/import, but perhaps they have a reason for that.

In any case, that should have been discussed in a separate Issue or PR. Sorry about that. In the meantime, the code allows the user to make more changes than they would have been able to before. But, since you can't figure that out via the actions, most users will never notice.

Anyway, please let me know whether you'd like me to revert this, or revert it and create a new PR, or whatever you think is best.

lib/src/clojure_lsp/feature/add_missing_libspec.clj
mainej: Yes, that's true. I realized that too, after I worked on this code. But I forgot to remove this TODO. I'll plan to remove it in a future PR.

Issue #638, opened 2021-11-11
Support disabling function symbol comletion when there is snippet equivalent


tami5:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when with seeing duplication of symbols, and hate that fact that snippet is less prioritized in my editor. 

For me instead of `letf|<cr>`, I have to `letf|<down><cr>`

**Describe the solution you'd like**

- An option to filter out symbols that has snippet with the same name.
- Doc for builtin symbols instead of snippet description.

**Additional context**
![image](https://user-images.githubusercontent.com/65782666/141341375-b4bc9633-7c03-4a72-a86c-5f3536ffd497.png)
![image](https://user-images.githubusercontent.com/65782666/141341394-6b738cd5-d510-4be3-b724-0a60b8b01419.png)
![image](https://user-images.githubusercontent.com/65782666/141341406-5670c9dd-32e2-4808-8a43-0bb9095798f6.png)

ericdallo:
@tami5 I'd like to understand this one better, for me we show the elements before the snippets, at least for me on emacs:
Is this the same behavior your are seeing?
![image](https://user-images.githubusercontent.com/7820865/141378213-92f52512-bba3-4d03-8eef-e1ea389f0a6b.png)

tami5:
> Is this the same behavior your are seeing?

Yes exactly, 9/10 what the user want is the snippet. it's rare that I'd want to complete and confirm a symbol by it's own. What I'm suggesting is an option to filter out function/macro symbol when there's a symbol with the same name.

![@floke ](https://user-images.githubusercontent.com/292349/118822916-6af02080-b86d-11eb-9990-942fe9b55541.png)
[this plugin](https://github.com/folke/lua-dev.nvim) went extra setup and made ALL lua function transformed to snippets which made writing lua code fun and intuitive 😆. When  the user confirm the function, the function argument becomes named $placeholders,.

Of course this would be tricky to implement because we'd need to take in consideration some cases:

- Is in threading macro, if so remove the first argument and complete the rest,
- What if the function has multiple arguments, easy the fewer arguments get prioritized.
- Is inside `()`, ignore appending `(`)
- Is inside `(|arg1 arg2 arg3)` then just symbol?

Symbol completion is cool, but completion for code structure is awesome.  "tab" over to fill the blank 🤣 and that's it.

[...]

PR #720, opened 2022-01-25
completion-item-resolve-fix

### LINKED ISSUE #719: completionItem/resolve does not return insertText
**Describe the bug**
When calling `completionItem/resolve` with a snippet completionItem the result should also contain that snippet.

[calva issue](https://github.com/BetterThanTomorrow/calva/issues/1483)
**To Reproduce**

1. type "kwarg", hit tab

sends:
```
Sending request 'completionItem/resolve - (99)'.
Params: {
    "label": "def",
    "detail": "clojure.core/def",
    "insertTextFormat": 2,
    "insertText": "(def ${1:name} ${0:value})",
    "kind": 18,
    "data": {
        "filename": "/clojure.core.clj",
        "name": "def",
        "ns": "clojure.core"
    }
}
Received response 'completionItem/resolve - (99)' in 39ms.
Result: {
    "label": "def",
    "kind": 18,
    "detail": "clojure.core/def",
    "data": {
        ":filename": "/clojure.core.clj",
        ":name": "def",
        ":ns": "clojure.core"
    }
}
```

**Expected behavior**
get an enriched version back

PR incoming

Cyrik:
Fixes #719 

not sure if it's ok to change the casing of `:plain-text`, but I couldn't find anything that relies on it.

### DIFF
--- a/cli/integration-test/integration/completion_test.clj
+++ b/cli/integration-test/integration/completion_test.clj
@@ -21,7 +21,6 @@
                                       :preselectSupport true
                                       :documentationFormat ["markdown" "plaintext"]
                                       :resolveSupport {:properties ["documentation" "detail" "additionalTextEdits"]}}}}}}))
-  ;; :client-capabilities :text-document :completion :completion-item :snippet-support
   (lsp/notify! (fixture/initialized-notification))
   (lsp/notify! (fixture/did-open-notification "completion/a.clj"))
   (lsp/notify! (fixture/did-open-notification "completion/b.clj"))
--- a/cli/integration-test/integration/completion_test.clj
+++ b/cli/integration-test/integration/completion_test.clj
@@ -32,7 +32,7 @@
          :kind 3
          :detail "clojure.core/definterface"
          :data {:filename "/clojure.core.clj", :name "definterface", :ns "clojure.core"}}]
-       (lsp/request! (fixture/completion-request "completion/a.clj" 2 3 1))))
+       (lsp/request! (fixture/completion-request "completion/a.clj" 2 3))))
     (testing "get snippets"
       (h/assert-contains-submaps
        [{:label "defn"
@@ -41,4 +41,4 @@
         ;;  :insertText "(defn ${1:name} [$2]\n  ${0:body})"
         ;;  :insertTextFormat 2
          :data {:filename "/clojure.core.clj" :name "defn" :ns "clojure.core", :snippet-kind 6}}]
-       (lsp/request! (fixture/completion-request "completion/a.clj" 2 4 1))))))
+       (lsp/request! (fixture/completion-request "completion/a.clj" 2 4))))))
--- a/cli/integration-test/integration/completion_test.clj
+++ b/cli/integration-test/integration/completion_test.clj
@@ -15,7 +15,6 @@
                   {:textDocument
                    {:completion
                     {:contextSupport true
-                     :completionItemKind {:valueSet (vec (range 1 26))}
                      :completionItem {:snippetSupport true
                                       :commitCharactersSupport true
                                       :preselectSupport true
@@ -36,9 +35,9 @@
     (testing "get snippets"
       (h/assert-contains-submaps
        [{:label "defn"
-        ;;  :kind 15 
-        ;;  :detail "clojure.core/defn" 
-        ;;  :insertText "(defn ${1:name} [$2]\n  ${0:body})"
-        ;;  :insertTextFormat 2
+         :kind 15
+         :detail "clojure.core/defn" 
+         :insertText "(defn ${1:name} [$2]\n  ${0:body})"
+         :insertTextFormat 2
          :data {:filename "/clojure.core.clj" :name "defn" :ns "clojure.core", :snippet-kind 6}}]
        (lsp/request! (fixture/completion-request "completion/a.clj" 2 4))))))
--- a/cli/integration-test/integration/completion_test.clj
+++ b/cli/integration-test/integration/completion_test.clj
@@ -1,0 +1,45 @@
+(ns integration.completion-test
+  (:require
+   [clojure.test :refer [deftest testing]]
+   [integration.fixture :as fixture]
+   [integration.helper :as h]
+   [integration.lsp :as lsp]))
+
+(lsp/clean-after-test)
+
+(deftest completion
+  (lsp/start-process!)
+  (lsp/request! (fixture/initialize-request
+                 {:initializationOptions {:lint-project-files-after-startup? false}
+                  :capabilities
+                  {:textDocument
+                   {:completion
+                    {:contextSupport true
+                     :completionItemKind {:valueSet (vec (range 1 26))}
+                     :completionItem {:snippetSupport true
+                                      :commitCharactersSupport true
+                                      :preselectSupport true
+                                      :documentationFormat ["markdown" "plaintext"]
+                                      :resolveSupport {:properties ["documentation" "detail" "additionalTextEdits"]}}}}}}))
+  ;; :client-capabilities :text-document :completion :completion-item :snippet-support
+  (lsp/notify! (fixture/initialized-notification))
+  (lsp/notify! (fixture/did-open-notification "completion/a.clj"))
+  (lsp/notify! (fixture/did-open-notification "completion/b.clj"))
+
+  (testing "normal completions"
+    (testing "get completions"
+      (h/assert-contains-submaps
+       [{:label "definterface"
+         :kind 3
+         :detail "clojure.core/definterface"
+         :data {:filename "/clojure.core.clj", :name "definterface", :ns "clojure.core"}}]
+       (lsp/request! (fixture/completion-request "completion/a.clj" 2 3 1))))
+    (testing "get snippets"
+      (h/assert-contains-submaps
+       [{:label "defn"
+        ;;  :kind 15 
+        ;;  :detail "clojure.core/defn" 
+        ;;  :insertText "(defn ${1:name} [$2]\n  ${0:body})"
+        ;;  :insertTextFormat 2
+         :data {:filename "/clojure.core.clj" :name "defn" :ns "clojure.core", :snippet-kind 6}}]
+       (lsp/request! (fixture/completion-request "completion/a.clj" 2 4 1))))))
--- a/cli/integration-test/integration/fixture.clj
+++ b/cli/integration-test/integration/fixture.clj
@@ -14,11 +14,17 @@
 
 (defn initialize-request
   ([]
-   (initialize-request {:lint-project-files-after-startup? false}))
-  ([settings]
+   (initialize-request {:initializationOptions {:lint-project-files-after-startup? false}}))
+  ([params]
    (lsp-json-rpc :initialize
-                 {:rootUri (h/file->uri (io/file h/root-project-path))
-                  :initializationOptions settings})))
+                 (merge {:rootUri (h/file->uri (io/file h/root-project-path))}
+                        params))))
+
+(defn completion-request [path row col kind]
+  (lsp-json-rpc :textDocument/completion
+                {:textDocument {:uri (h/file->uri (h/source-path->file path))}
+                 :position {:line row :character col}
+                 :context {:triggerKind kind}}))
 
 (defn definition-request [path row col]
   (lsp-json-rpc :textDocument/definition
--- a/cli/integration-test/integration/fixture.clj
+++ b/cli/integration-test/integration/fixture.clj
@@ -20,11 +20,10 @@
                  (merge {:rootUri (h/file->uri (io/file h/root-project-path))}
                         params))))
 
-(defn completion-request [path row col kind]
+(defn completion-request [path row col]
   (lsp-json-rpc :textDocument/completion
                 {:textDocument {:uri (h/file->uri (h/source-path->file path))}
-                 :position {:line row :character col}
-                 :context {:triggerKind kind}}))
+                 :position {:line row :character col}}))
 
 (defn definition-request [path row col]
   (lsp-json-rpc :textDocument/definition
--- a/cli/integration-test/integration/helper.clj
+++ b/cli/integration-test/integration/helper.clj
@@ -58,13 +58,6 @@
      (doseq [m# maps#]
        (let [m-keys (keys m#)]
          (is (some #(= m# (some-> % (select-keys m-keys))) res#)))))) 
-(comment
-  (let [maps# [{:a 1 :b 1} {:a 3}]
-        res# [{:a 1 :b 1} {:a 3 :b 5}]]
-    (doseq [m# maps#]
-      (let [m-keys (keys m#)]
-        (is (some #(= m# (some-> % (select-keys m-keys))) res#)))))
-  )
 
 (defn delete-folder
   "Recursively delete a directory."
--- a/cli/integration-test/integration/helper.clj
+++ b/cli/integration-test/integration/helper.clj
@@ -50,6 +50,22 @@
        (doseq [[r# m#] (map vector res# maps#)]
          (assert-submap m# r#)))))
 
+(defmacro assert-contains-submaps
+  "Asserts that maps are contained submaps of result in results. "
+  [maps result]
+  `(let [maps# ~maps
+         res# ~result]
+     (doseq [m# maps#]
+       (let [m-keys (keys m#)]
+         (is (some #(= m# (some-> % (select-keys m-keys))) res#)))))) 
+(comment
+  (let [maps# [{:a 1 :b 1} {:a 3}]
+        res# [{:a 1 :b 1} {:a 3 :b 5}]]
+    (doseq [m# maps#]
+      (let [m-keys (keys m#)]
+        (is (some #(= m# (some-> % (select-keys m-keys))) res#)))))
+  )
+
 (defn delete-folder
   "Recursively delete a directory."
   [^String filename]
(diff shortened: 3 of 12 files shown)


ericdallo:
@Cyrik we had this discussion before [here](https://github.com/clojure-lsp/clojure-lsp/issues/638)

Cyrik:
Thanks for the link.
I think it makes sense, but we would not have to build the "is it a snippet" logic into every editor to add a ~ at the end. That should either happen in LSP or the snippet kind should be returned by default and optionally it returns the actual type.

Also, there is no "use completion, but not snippet" hotkey in vscode, I think. So if we don't provide both completions there is no way for the user to complete without the snippet. That's not necessarily a problem, but it will be confusing.
I like the way vscode solves this: always provide both, but show snippets above everything else.

ericdallo:
@Cyrik the point is why one would not choose a snippet but a regular completion? the completion just help achieve what you want.

BTW, it seems there is happening some issues with this PR:
image

It'd be really nice if we could add integration tests for completion

ericdallo:
> 
I see your point for the current snippets. But we really should at least mark them as snippets, otherwise, it's really confusing. I can do switch the kind on the Calva side if you prefer to return the actual type.

Maybe @tami5 can help understand why this

tami5:
> > I see your point for the current snippets. But we really should at least mark them as snippets, otherwise, it's really confusing. I can do switch the kind on the Calva side if you prefer to return the actual type.
> 
> Maybe @tami5 can help understand why this

hmmm I'm a bit lost. The kind needs to be snippet instead of the actual type? Not sure why or what issue does having this solve. 

Normally snippet or not while writing code you want to glance what the symbol is, not what confirmation might produce, and if that the case, having duplications say defn the function symbol or defn the snippet is extra confusing and redundant.

> not have to build the "is it a snippet" logic into every editor to add a ~ at the end.

not sure what you mean here, but isn't client (like neovim) suppose to process lsp in unified manner

Cyrik:
> hmmm I'm a bit lost. The kind needs to be snippet instead of the actual type? Not sure why or what issue does having this solve.
Normally snippet or not while writing code you want to glance what the symbol is, not what confirmation might produce, and if that the case, having duplications say defn the function symbol or defn the snippet is extra confusing and redundant.

@tami5 vscode does not give you very much information by default. It's mostly just the icon + small text. So if the icon does not show you that you are selecting a snippet, you would expect just the text. Currently the defn snippet returns the function type.

> not sure what you mean here, but isn't client (like neovim) suppose to process lsp in unified manner

Usually, the clients just use the returned completion and don't modify it. So if we wanted to have the neovim behavior, both calva and emacs would have to add it. See the emacs discussion https://github.com/emacs-lsp/lsp-mode/pull/3209

tami5:
> @tami5 vscode does not give you very much information by default. It's mostly just the icon + small text. So if the icon does not show you that you are selecting a snippet, you would expect just the text. Currently the defn snippet returns the function type.

Yah this maybe a shortcoming of a vscode. The issue isn't with having the supposedly incorrect type returned by server but actually that the user is unsure whether the item will expand or not. At the end of the day, It is really a client not a server concern. Perhaps if it's reasonable with @ericdallo, a temporary option, "that is disabled by default" adds  `~` or `$` for those clients still catching up to this trend of; I guess expandable items that are labeled as snippets.
 
> Usually, the clients just use the returned completion and don't modify it. So if we wanted to have the neovim behavior, both calva and emacs would have to add it. 

I saw what @yyoncho mentioned about not wanting to follow neovim way, well no issues maybe `$` then, the point is that lsp clients must somehow show the user that; hey, this item will expand!.

Cyrik:
@tami5 I do agree with you that vscode has a lot of problems with some of these things. But I'm ok with changing the completion kind on the Calva side to 15 if it has an insert-text.

> I saw what @yyoncho mentioned about not wanting to follow neovim way, well no issues maybe $ then, the point is that lsp clients must somehow show the user that; hey, this item will expand!.

I think the difference is that vscode does this with the icon, which depends on the completion-kind. I do fully support giving extra information so neovim can show more. The easiest way for Calva + emacs would be to get `:completion-kind 15` and get something like `:snippet-kind 2` as an extra in the data object. Would that work for you?

tami5:
> The easiest way for Calva + emacs would be to get :completion-kind 15 and get something like :snippet-kind 2 as an extra in the data object. Would that work for you?

makes sense, In general any extra fields that would help clients and is not totally out of the norm is welcomed.

Cyrik:
@ericdallo Are you ok with that solution? If so I'd add that to the PR as well.

I'll try to add a test + fix for the new bug tomorrow.

ericdallo:
I think it's worth the shot @Cyrik

Cyrik:
So everything is working now, but for some reason {:data{:snipet-kind 123}} comes out as ::snipet-kind. Any idea why @ericdallo ?

cli/integration-test/integration/completion_test.clj
```diff
@@ -0,0 +1,45 @@
...
+
+(deftest completion
+  (lsp/start-process!)
+  (lsp/request! (fixture/initialize-request
+                 {:initializationOptions {:lint-project-files-after-startup? false}
+                  :capabilities
+                  {:textDocument
+                   {:completion
+                    {:contextSupport true
+                     :completionItemKind {:valueSet (vec (range 1 26))}
```
ericdallo, line 18: Why we need this line? I don't see emacs sending it

cli/integration-test/integration/completion_test.clj
```diff
@@ -0,0 +1,45 @@
...
+                     :completionItem {:snippetSupport true
+                                      :commitCharactersSupport true
+                                      :preselectSupport true
+                                      :documentationFormat ["markdown" "plaintext"]
+                                      :resolveSupport {:properties ["documentation" "detail" "additionalTextEdits"]}}}}}}))
```
ericdallo, line 23: Aren't we missing `insertTextModeSupport` under `completion completionItem`?

cli/integration-test/integration/completion_test.clj
```diff
@@ -0,0 +1,45 @@
...
+  ;; :client-capabilities :text-document :completion :completion-item :snippet-support
```
ericdallo, line 24:
```suggestion
```

cli/integration-test/integration/fixture.clj
```diff
@@ -14,11 +14,17 @@
...
-                 {:rootUri (h/file->uri (io/file h/root-project-path))
-                  :initializationOptions settings})))
+                 (merge {:rootUri (h/file->uri (io/file h/root-project-path))}
+                        params))))
+
+(defn completion-request [path row col kind]
+  (lsp-json-rpc :textDocument/completion
+                {:textDocument {:uri (h/file->uri (h/source-path->file path))}
+                 :position {:line row :character col}
+                 :context {:triggerKind kind}}))
```
ericdallo, line 27: Do we need to send this :context or triggerKind? we don't check for that any time on clojure-lsp

cli/integration-test/integration/helper.clj
```diff
@@ -50,6 +50,22 @@
...
+       (let [m-keys (keys m#)]
+         (is (some #(= m# (some-> % (select-keys m-keys))) res#)))))) 
+(comment
+  (let [maps# [{:a 1 :b 1} {:a 3}]
+        res# [{:a 1 :b 1} {:a 3 :b 5}]]
+    (doseq [m# maps#]
+      (let [m-keys (keys m#)]
+        (is (some #(= m# (some-> % (select-keys m-keys))) res#)))))
+  )
+
```
ericdallo, line 68: I'm not a big fan of comments on the middle of the file, can we remove it?

lib/src/clojure_lsp/feature/completion.clj
```diff
@@ -327,10 +332,11 @@
     (concat
       (map (fn [item]
              (if-let [snippet-item (get snippet-items-by-label (:label item))]
-               (shared/assoc-some snippet-item
-                                  :detail (:detail item)
-                                  :data (:data item)
-                                  :kind (:kind item))
+               (let [data (shared/assoc-some (:data item)
+                                             :snippet-kind (get completion-kind-enum (:kind item)))]
```
ericdallo, line 336: I'm kind of lost why we are adding this only because of vim editor. 
This field is no part of the spec and it's already the :kind but as int, right?

lib/src/clojure_lsp/feature/completion.clj
```diff
@@ -450,8 +456,14 @@
...
-    (if ns
-      (resolve-item-by-ns item db)
-      (resolve-item-by-definition item db))
-    item))
+  (let [item (shared/assoc-some item
+                                :insert-text-format (:insertTextFormat item)
+                                :text-edit (:textEdit item)
+                                :filter-text (:filterText item)
+                                :additional-text-edits (:additionalTextEdits item)
+                                :insert-text (:insertText item))]
```
ericdallo, line 464: I think is not that simple, as for example `:additionalTextEdits` will return a List<TextEdit>, not a clojure list of maps representing the text edit, right?
This will probably fail the coercer which expect clojure maps and lists but will get now a bunch of fields which are classes.

Maybe the best is to try to make the borders parse from/to java/clojure properly so the feature/completion always handle keywords, do you think that would be possible?

lib/test/clojure_lsp/features/completion_test.clj
```diff
@@ -229,7 +229,7 @@
       [{:label "comment"
         :detail "clojure.core/comment"
         :insert-text "(comment\n  ${0:body}\n  )"
-        :kind :reference
+        :kind :snippet
```
ericdallo, line 232: Why this was not failing before?

ericdallo:
@Cyrik thanks for the integration tests, although I'd like to understand better the :snppet-kind need.
Also, I know I said to try to solve the issue specifically for the completion but the way was solved will probably not work because what I commented [here](https://github.com/clojure-lsp/clojure-lsp/pull/720#discussion_r794155305). Maybe we can try on this PR to focus on fix the root cause which is missing insertText which is just a string instead of adding other complex fields? WDYT?

cli/integration-test/integration/completion_test.clj
Cyrik: This is one of the capabilities Calva sends. I wanted to stay kind of close to the actual usage, so that the test works if lsp ever starts checking that capability. I'm happy to remove it though.

cli/integration-test/integration/completion_test.clj
Cyrik: [doc](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#completionClientCapabilities)
the relevant part states:
> Indicates which properties a client can resolve lazily on a completion item.

I understand this to mean: "that's the parts you can add to the original item on a completionItem/Resolve call".
The values in here are what Calva sends.

cli/integration-test/integration/fixture.clj
Cyrik: I didn't check if we use it, just that it's getting sent. I'll remove it.

cli/integration-test/integration/helper.clj
Cyrik: Yes, I'll remove it.
Do you like them at the end or not at all?

lib/src/clojure_lsp/feature/completion.clj
Cyrik: See my [comment](https://github.com/clojure-lsp/clojure-lsp/pull/720#discussion_r794592439) for why I added it.
As for type: `:kind` will be coerced into an int by the spec + java-lsp lib. I needed to do that manually for `:snippet-kind` because `:data` does not go through the same process.

Regarding the `::`, are you sure? I don't see a spec for `::data` and couldn't find any special handling for `:ns`, `:filename`, or `:name`.

lib/src/clojure_lsp/feature/completion.clj
Cyrik: I didn't about `:additionalTextEdits`, good catch.

Yeah this whole thing is kind of ugly, but I couldn't find a good way handle camelCase <-> kebab-case with the way `lsp4j` flows into the `clojure.java.data` functions. I'd like to do the camelCase <-> kebab-case conversion by hand in the `clojure.java.data/from-java` multimethod, but `clojure-lsp.server/sync-request` pipes the `clojure-lsp.coercer/java->clj` call directly into the handler. 
I think the cleanest solution would be to also pipe this through a request-spec spec before calling the handler. That way we can change the casing or any other stuff that might come up at the "right" location.

What do you think?

Cyrik:
> @Cyrik thanks for the integration tests, although I'd like to understand better the :snppet-kind need. Also, I know I said to try to solve the issue specifically for the completion but the way was solved will probably not work because what I commented [here](https://github.com/clojure-lsp/clojure-lsp/pull/720#discussion_r794155305). Maybe we can try on this PR to focus on fix the root cause which is missing insertText which is just a string instead of adding other complex fields? WDYT?

I think it would be ok to pull out `:additionalTextEdits` for now and create a new issue for it that fixes the conversion in a cleaner way, maybe like I [suggested](https://github.com/clojure-lsp/clojure-lsp/pull/720#discussion_r794615928). That way we would have this done once I find out why `::snippet-kind` gets namespaced.

lib/test/clojure_lsp/features/completion_test.clj
tami5: hmmm I thought it was going to be an additional field not a change on the data structure, but `kind` into `:snippet-kind` feels like restructure no? I'm worried it might break existing setups but anyways whatever @ericdallo see appropriate.

cli/integration-test/integration/completion_test.clj
ericdallo: Yeah, that part makes sense, sorry I think I was not clear or my comment should be on few lines above.
I meant that we are not sending InsertTextMode, that are the ones client support, at least emacs send that, maybe this could affect the test somehow?

cli/integration-test/integration/helper.clj
ericdallo: I accept then on the end, but personal I remove it, but feel free to choose

lib/src/clojure_lsp/feature/completion.clj
ericdallo: Yes, sounds better indeed

lib/src/clojure_lsp/feature/completion.clj
Cyrik: I found it. I missed that the key in `:data` just gets to-stringed. Fixing it right now for :snipet-kind.

lib/test/clojure_lsp/features/completion_test.clj
ericdallo: Yeah, a additional field outside of data seems wrong and not following the spec, the spec has the data field for client route the data to server again, although I'm okay adding a field to the data, but yes, it looks to me clojure-lsp should return the snippet as :kind indeed

Cyrik: @tami5 It would not break the current implementation in that sense, since `:kind` still exists and still returns a completion-kind. The "enriched" information would move to `:data {:snippet-kind "kind"}` for anyone who wants to display that instead, which is "only" vim I think.

But yes @ericdallo one of the IDEs will have to do extra work since their handling of completionItem display is different.

cli/integration-test/integration/completion_test.clj
Cyrik: it probably should, but I couldn't find any code that takes that client capability into account.

Cyrik:
I fixed the extra `:` issues and removed `:additional-text-edits` handling from clompletionItem/resolve.
This should be fine to merge and we should start a second issue or PR for coercion cleanup + `:additional-text-edits` handling.

cli/integration-test/integration/completion_test.clj
ericdallo: Yeah, this is true, doesn't makes sense to add I guess

lib/src/clojure_lsp/feature/completion.clj
```diff
@@ -18,6 +18,11 @@
 
 (set! *warn-on-reflection* true)
 
+(def completion-kind-enum
+  {:text 1 :method 2 :function 3 :constructor 4 :field 5 :variable 6 :class 7 :interface 8 :module 9
+   :property 10 :unit 11 :value 12 :enum 13 :keyword 14 :snippet 15 :color 16 :file 17 :reference 18
+   :folder 19 :enummember 20 :constant 21 :struct 22 :event 23 :operator 24 :typeparameter 25})
```
ericdallo, line 24: Not sure it's a good idea having this enum here as it's related to the LSP enum which is specific for `cli`, but I can't think on a better solution now

ericdallo: Maybe it's part of fixing the borders to correctly coerce java->clj and clj->java

ericdallo (commented):
Code looks ok to me, but we should focus on fixing those border parsers to not let these cases issues grow, I'm ok for now fixing this way.

Also, thank you for the integration tests, it's a good start to help make sure the behavior is working in the future when we change the coercer again.
I'll manually test and If everything good we are good to merge

ericdallo (approved):
Works good! thanks!

mainej:
@Cyrik @kkharji @ericdallo I'm working on a project to automatically convert messages from clojure-lsp back to the clients to camelCase. One thing that's tripping me up is `:snippet-kind`. Without some special casing, my project wants to convert that to "snippetKind". But, I assume that one of the clients is relying on it being spelled "snippet-kind". I want to ask a maintainer of whichever client this is whether that would be OK. What I can't figure out from the above conversation is who to ask. :) Who's relying on "snippet-kind"?

Cyrik:
Nice work. Just for posterity, it was emacs that's relying on that.

Issue #704, opened 2022-01-20
Fix flaky tests


ericdallo:
We already had flaky tests in the past and a lot of them were already fixed, but we still have some flaky tests especially on integration tests.

ericdallo:
It seems to me that the integration-test `diagnostics_test` is the most (only?) flaky ATM

mainej:
I looked at this a bit, but didn't make much progress. Here's what I found...

I see two tests that fail intermittently. 
1. `integration-test/integration/settings_change_test.clj:16` [Ex](https://github.com/mainej/clojure-lsp/runs/4907925973#step:7:107)
2. `integration-test/integration/diagnostics_test.clj:10` [Ex](https://github.com/clojure-lsp/clojure-lsp/runs/4899541082#step:7:71)

Both tests analyze `sample-test/src/sample_test/diagnostics/unused_public_var.clj`, and in both cases they are expecting diagnostics that report two unused public vars for that file. When they fail, they see no diagnostics.

I've identified a few possible sources of flakiness, though I'm not sure this is all of them:
1. In failing tests, the `await-diagnostics` loop could be finding the wrong diagnostics message. I don't think this is the case because the test logs seem to show that the same messages are sent and received whether the test passes or fails. When the test fails a few later messages aren't sent, but I assume that's because the tests are in some sort of fail-fast mode.
2. In failing tests, the `.lsp/config.edn` could be stale, with the public-var linting turned off.
3. In failing tests, the kondo cache could be used instead of a fresh run, or vice versa.
4. In failing tests, the kondo cache could be stale, or be a mismatch with what you'd expect for the current `.lsp/config.edn`.
5. In failing tests, the kondo analysis could fail to detect unused public vars.
6. In failing tests, the kondo analysis could be running asynchronously. It _appears_ that clojure-lsp invokes kondo synchronously (i.e. not `:parallel`) for public var linting but it may be worth researching.
7. In failing tests, the :project-root could be wrong. If this is happens, the linted file is considered "external" to the project, and is therefore excluded from public var linting. This doesn't seem to be the case, because the `initialize-request` fixture sets it in both passing and failing runs.

I don't have the time to keep looking at this right now, but here are some avenues of further research:

1. Confirm whether the tests are in a fail-fast mode. If not, research why message logs are different in passing and failing tests.
2. Log the contents of`.lsp/config.edn` before each test, to ensure public var linting is on/off as expected.
[...]

PR #723, opened 2022-01-27
Ensure lsp config is restored in integration test that changes it

mainej:
As discussed in https://github.com/clojure-lsp/clojure-lsp/issues/704.

### DIFF
--- a/cli/integration-test/integration/settings_change_test.clj
+++ b/cli/integration-test/integration/settings_change_test.clj
@@ -1,15 +1,23 @@
 (ns integration.settings-change-test
   (:require
    [clojure.string :as string]
-   [clojure.test :refer [deftest testing]]
+   [clojure.test :refer [deftest is testing]]
    [integration.fixture :as fixture]
    [integration.helper :as h]
    [integration.lsp :as lsp]))
 
 (lsp/clean-after-test)
 
-(def current-lsp-config-file (h/source-path->file "../../.lsp/config.edn"))
-(def current-lsp-config-content (slurp current-lsp-config-file))
+(def lsp-config-file (h/source-path->file "../../.lsp/config.edn"))
+
+(defn with-revised-lsp-config-content [new-content f]
+  (let [old-content (slurp lsp-config-file)]
+    (is (not= new-content old-content))
+    (try
+      (spit lsp-config-file new-content)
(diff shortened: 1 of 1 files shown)


mainej:
The [failing integration test](https://github.com/clojure-lsp/clojure-lsp/runs/4961198768?check_suite_focus=true) for this PR proves that this _doesn't_ fix the flakiness, but it still seems like the right thing to do.

cli/integration-test/integration/settings_change_test.clj
```diff
@@ -37,15 +45,14 @@
         :tags [1]}]
       (lsp/await-diagnostics "diagnostics/unused_public_var.clj")))
 
-  (spit current-lsp-config-file new-lsp-config-content)
-
-  (Thread/sleep 1100)
-
-  (lsp/notify! (fixture/did-open-notification "diagnostics/unused_public_var.clj"))
+  (with-revised-lsp-config-content
```
ericdallo, line 48: Maybe we can use a feature from clojure.test for this purpose, called `clojure.test/use-fixtures`, we use it a lot on the unit-tests, check `clojure-lsp.test-helper/reset-db-after-test`

ericdallo:
Also, could you resolve the conflicts?

cli/integration-test/integration/settings_change_test.clj
mainej: `with-revised-lsp-config-content` does look like it should be a fixture. But I see two problems...
1. Fixtures are applied before or after the deftest as a whole, whereas we want to change the config halfway through a deftest
2. Fixtures are applied around every deftest in a namespace.

We could fix (1) by splitting the tests into two deftests, but because of (2) then both deftests would get the revised config.

I think the only way around both problems would be to create a separate namespace, which seems like it would obscure the intention of these tests. So, I'm inclined to keep the test as is. What do you think?

mainej:
Conflicts resolved.

cli/integration-test/integration/settings_change_test.clj
ericdallo: makes sense! no problem keeping it this way

ericdallo (approved):
Thank you!

PR #724, opened 2022-01-27
Turn off async custom linters in integration tests

mainej:
In an attempt to fix flaky tests #704.

### DIFF
--- a/cli/integration-test/integration/diagnostics_test.clj
+++ b/cli/integration-test/integration/diagnostics_test.clj
@@ -62,7 +62,8 @@
 (deftest report-duplicates-disabled
   (lsp/start-process!)
   (lsp/request! (fixture/initialize-request {:lint-project-files-after-startup? false
-                                             :linters {:clj-kondo {:report-duplicates false}}}))
+                                             :linters {:clj-kondo {:report-duplicates false
+                                                                   :async-custom-lint? false}}}))
   (lsp/notify! (fixture/initialized-notification))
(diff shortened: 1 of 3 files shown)


mainej:
This is just an experiment for now. I'm curious whether the integration tests will be green. If they are, I'll bump it several times to see if they stay green.

mainej:
See all test runs here: https://github.com/clojure-lsp/clojure-lsp/actions/workflows/ci.yml?query=branch%3Async-custom-linters-in-integration-tests

mainej:
OK, let me summarize my thinking here... Almost all of the recent integration test failures are related to linting unused-public-vars. I remembered that `:clojure-lsp/unused-public-var` is a custom clj-kondo linter implemented by clojure-lsp. And looking at the custom linter, I saw that it has an async version, which is enabled by default. On a hunch, I turned off that settting, making the custom linter synchronous in the integration tests.

Then I ran the integration tests several times.

At this point there have been six runs of the integration test matrix on this branch. Five out of six passed. The one failure was a timeout waiting for a response to [this linked-editing-range request](https://github.com/clojure-lsp/clojure-lsp/blob/c1890e53e54b8e553124fa0c9ffd935d6013f3e0/cli/integration-test/integration/linked_editing_range_test.clj#L22). There were no failures related to unused-public-vars.

So, I think this suggests that a lot of the flakiness is related to the async custom linter. I'm not sure that turning it off is the right way to fix the flakiness, but at least we have a lead.

ericdallo (approved):
I'm totally OK disabling it for integration-tests, thanks for the finding!

PR #726, opened 2022-01-28
Remove "resolve code action"

### LINKED ISSUE #722: When using the action menu, the cursor is not repositioned
**Describe the bug**

Some commands attempt to move the cursor via `:show-document-after-edit`. The cursor is not relocated correctly when the command is invoked from the action menu. Invoking the same command directly does move the cursor.

There are two parts to this bug. First the code that attempts to move the cursor was broken by 2b8b66bf8, which removed the `:client` from the db. [This line](https://github.com/clojure-lsp/clojure-lsp/blob/c1890e53e54b8e553124fa0c9ffd935d6013f3e0/lib/src/clojure_lsp/handlers.clj#L291) wasn't updated to remove the check for the `:client`.

However, even after removing that check,  the bug persists. It appears that on the (working) `executeCommand` path, the edit happens before the cursor movement, but on the `resolveCodeAction` path, they happen in the reverse order, with the movement before the edit. See logs below.

This bug probably also manifests in a different way, which I haven't researched. I suspect that related files are not focused (when a command is invoked from the action menu). It would happen for anything that uses `:take-focus? true`: creating tests, or creating functions in an aliased namespace.

**To Reproduce**

1. Create a map with several entries, each covering a different number of lines, including their comments.
2. Move cursor to one of the entries. Note this as the "starting cursor"
2. Invoke "Move coll entry down" from the action menu
3. Note where the cursor ends up as the "after action cursor"
4. Undo. Move back to the "starting cursor" position.
5. Invoke `lsp-clojure-move-coll-entry-down` directly.
6. Note where the cursor ends up as the "after command cursor".
7. Observe the "after action cursor" and "after command cursor" are not the same. You may have to retry the experiment with different "starting cursor" positions, including on the comments or between a key and value.

**Expected behavior**

The cursor is always positioned in the same place, regardless of how the command is invoked.

**Screenshots**

<details>
 <summary><b>Log - clojure-lsp</b> :executeCommand</summary>
<pre>
2022-01-27T02:41:55.509Z  DEBUG [clojure-lsp.server:?] - :didChange 1ms
2022-01-27T02:41:55.509Z  INFO [clojure-lsp.server:495] - Requesting to show on editor the document {:uri "file:///Users/jmaine/workspace/opensource/clojure-lsp/lib/src/clojure_lsp/feature/move_coll_entry.clj", :take-focus? true, :range {:row 470, :col 4, :end-row 470, :end-col 4}}
2022-01-27T02:41:55.515Z  DEBUG [clojure-lsp.server:?] - :executeCommand 40ms
</pre>
</details>

<details>
 <summary><b>Log - clojure-lsp</b> :resolveCodeAction</summary>
<pre>
2022-01-27T02:10:18.753Z  DEBUG [clojure-lsp.server:?] - :codeAction 161ms
2022-01-27T02:10:19.935Z  INFO [clojure-lsp.server:495] - Requesting to show on editor the document {:uri "file:///Users/jmaine/workspace/opensource/clojure-lsp/lib/src/clojure_lsp/feature/move_coll_entry.clj", :take-focus? true, :range {:row 469, :col 4, :end-row 469, :end-col 4}}
2022-01-27T02:10:19.940Z  DEBUG [clojure-lsp.server:?] - :resolveCodeAction 50ms
2022-01-27T02:10:20.530Z  DEBUG [clojure-lsp.server:?] - :didChange 3ms
</pre>
</details>

 - OS: MacOS
 - Editor emacs
 - Version: clojure-lsp 2022.01.22-01.31.09, clj-kondo 2022.01.15

### LINKED ISSUE #725: Do code actions need to be resolved?

There are two code paths for getting edits applied. 

First, obviously, you can execute a command. This code path returns nil but begins a background process to calculate edits, which the editor will be notified about later. 

Second, you can use the action menu. Currently, after picking an action menu item, the server is sent a "resolve code action" request. (If the client supports that capability. If the client doesn't, the server is sent an "execute command" request.) This "resolve code action" request also schedules a background process to calculate edits. (It currently fails to return a valid response—its return value doesn't match the expected clojure.spec—but that's not the bug I'm focused on here.)

Having two code paths leads to some code duplication and to bugs like #722, which stem from slight differences in the way the client is notified of edits.

I'd like to propose that the "resolve code action" path be removed. Here's my argument... 

According to the [LSP spec on code actions](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_codeAction), when the server is generating the action menu options, it has three options.

First, it can include edits along with an action menu item. The spec recommends that servers only do this if the edits are so easy to compute that they won't slow down the generation of the menu option. `clojure-lsp` does not use this option currently.

Second, the server can include a command along with an action menu item. When the menu item is selected the client is supposed to send this command back to the server, i.e. "execute" it. The spec recommends that servers do this whenever the edits would take too long to calculate but they can quickly calculate a command. `clojure-lsp` asks clients to use this option if they don't support the third option.

Third, the server can include some arbitrary data with the action menu item. When the menu item is selected, the client sends this data back to the server via the "resolve code action" path. The server then generates the edits. The spec recommends that servers do this when both the edits _and_ the command would take too long to calculate. It's a way for the server to say "I know this action menu item should be available, but figuring out the exact command would slow down the menu popup". Not all clients support this option, but when they do `clojure-lsp` always asks them to follow it.

I think `clojure-lsp` should stop generating menu items of the third type. My argument is that `clojure-lsp` never actually takes advantage of the "resolve code action" capability. That is, it never uses it as a shortcut to generate the menu items more quickly. Thus, it might as well always generate options of the second type. Then all clients would execute commands, instead of some executing commands and some resolving code actions, ensuring that all edits are applied in the same way.

The biggest question is whether all the editors will be OK with this. It's possible that some editors that support the "resolve code action" capability actually _require_ it.

One open question is whether the server should stop advertising that it supports "resolve code action". This may be necessary for some editors, who might otherwise try to follow that path regardless of the contents of the menu item.

I'll work on a PR for this, but will need help testing in various editors (I use Emacs so can test there.) There aren't currently any integration tests for action menu interaction or direct execution of commands, but there should be to validate that this refactoring works correctly. I'll write some of those tests as I work on the PR.

mainej:
Resolves #725 and fixes #722. This changes code actions (the action menu items) so that they always execute commands. This is as opposed to having them invoke "resolve code action". The rational for this change is in #725.

I have tested this in Emacs. The one strange behavior I noticed was that resolve-macro asks the same questions twice (to get the macro style and then the config file). I haven't check whether that's new behavior or old. In any case, the refactoring does work in the end.

I'd appreciate help testing this on other editors. I believe that all that is needed is a quick check that any of the action menu refactorings work, though to be thorough, all the [refactorings](https://clojure-lsp.io/features/#refactoring) could be checked. It would also be good to check the "interactive" menu items: resolve macro and create test.

### DIFF
--- a/cli/integration-test/integration/fixture.clj
+++ b/cli/integration-test/integration/fixture.clj
@@ -12,22 +12,44 @@
      :params params
      :id (lsp/inc-request-id)}))
 
+(def default-init-options {:lint-project-files-after-startup? false})
+
 (defn initialize-request
   ([]
-   (initialize-request {:lint-project-files-after-startup? false}))
+   (initialize-request default-init-options))
   ([settings]
+   (initialize-request settings nil))
+  ([settings capabilities]
    (lsp-json-rpc :initialize
-                 {:rootUri (h/file->uri (io/file h/root-project-path))
-                  :initializationOptions settings})))
+                 (cond-> {:rootUri (h/file->uri (io/file h/root-project-path))
+                          :initializationOptions settings}
+                   capabilities (assoc :capabilities capabilities)))))
+
+(defn initialize-emacs-request []
+  (initialize-request
+   default-init-options
+   ;; Emacs' advertised client capabilities
+   ;; LSP :: lsp-mode 20220115.1742, Emacs 27.2, darwin
+   ;; Derived from a printout of client capabilities that was halfway through
+   ;; the json -> java -> coerce process. It was messy. I hope this is accurate.
+   ;; Emacs advertises many other capabilities, but this includes only the ones
+   ;; that clojure-lsp cares about as of 2022-01-28.
+   {:experimental {:testTree true}
+    :textDocument {:hover      {:contentFormat ["markdown" "plaintext"]}
+                   :completion {:completionItem {:snippetSupport true}}}
+    :window       {:showDocument {:support true}}
+    :workspace    {:workspaceEdit {:documentChanges true}
+                   ;; does not advertise code lens refresh support, though clojure-lsp would use it
+                   #_#_:codeLens  {:refreshSupport true}}}))
 
 (defn definition-request [path row col]
   (lsp-json-rpc :textDocument/definition
-                {:textDocument {:uri (h/file->uri (h/source-path->file path))}
+                {:textDocument {:uri (h/source-path->uri path)}
                  :position {:line row :character col}}))
 
 (defn declaration-request [path row col]
   (lsp-json-rpc :textDocument/declaration
-                {:textDocument {:uri (h/file->uri (h/source-path->file path))}
+                {:textDocument {:uri (h/source-path->uri path)}
                  :position {:line row :character col}}))
 
 (defn formatting-full-request [path]
@@ -38,7 +60,7 @@
 
 (defn rename-request [path new-name row col]
   (lsp-json-rpc :textDocument/rename
-                {:textDocument {:uri (h/file->uri (h/source-path->file path))}
+                {:textDocument {:uri (h/source-path->uri path)}
                  :position {:line row :character col}
                  :newName new-name}))
 
@@ -52,21 +74,32 @@
 
 (defn document-symbol-request [path]
   (lsp-json-rpc :textDocument/documentSymbol
-                {:textDocument {:uri (h/file->uri (h/source-path->file path))}}))
+                {:textDocument {:uri (h/source-path->uri path)}}))
 
 (defn document-highlight-request [path row col]
   (lsp-json-rpc :textDocument/documentHighlight
-                {:textDocument {:uri (h/file->uri (h/source-path->file path))}
+                {:textDocument {:uri (h/source-path->uri path)}
                  :position {:line row :character col}}))
 
 (defn linked-editing-range-request [path row col]
   (lsp-json-rpc :textDocument/linkedEditingRange
-                {:textDocument {:uri (h/file->uri (h/source-path->file path))}
+                {:textDocument {:uri (h/source-path->uri path)}
                  :position {:line row :character col}}))
+
+(defn code-action-request [path row col]
+  (lsp-json-rpc :textDocument/codeAction
+                {:textDocument {:uri (h/source-path->uri path)}
+                 :context      {:diagnostics []}
+                 :range        {:start {:line row :character col}}}))
+
+(defn execute-command-request [command & args]
+  (lsp-json-rpc :workspace/executeCommand
+                {:command   command
+                 :arguments args}))
 
 (defn cursor-info-raw-request [path row col]
   (lsp-json-rpc "clojure/cursorInfo/raw"
-                {:textDocument {:uri (h/file->uri (h/source-path->file path))}
+                {:textDocument {:uri (h/source-path->uri path)}
                  :position {:line row :character col}}))
 
 (defn initialized-notification []
--- a/cli/integration-test/integration/fixture.clj
+++ b/cli/integration-test/integration/fixture.clj
@@ -13,20 +13,21 @@
      :id (lsp/inc-request-id)}))
 
 (def default-init-options {:lint-project-files-after-startup? false})
-(def emacs-capabilities
-  ;; Emacs' advertised client capabilities
-  ;; LSP :: lsp-mode 20220115.1742, Emacs 27.2, darwin
-  ;; Derived from a printout of client capabilities that was halfway through
-  ;; the json -> java -> coerce process. It was messy. I hope this is accurate.
-  ;; Emacs advertises many other capabilities, but this includes only the ones
-  ;; that clojure-lsp cares about as of 2022-01-28.
+(def full-client-capabilities
+  "Turn on all the client capabilities that clojure-lsp cares about.
+
+  ```
+  (intiailize-request {:initializationOptions default-init-opts
+                       :capabilities full-client-capabilities})
+  ```
+
+  Last updated 2022-01-28."
   {:experimental {:testTree true}
    :textDocument {:hover      {:contentFormat ["markdown" "plaintext"]}
                   :completion {:completionItem {:snippetSupport true}}}
    :window       {:showDocument {:support true}}
    :workspace    {:workspaceEdit {:documentChanges true}
-                  ;; does not advertise code lens refresh support, though clojure-lsp would use it
-                  #_#_:codeLens  {:refreshSupport true}}})
+                  :codeLens {:refreshSupport true}}})
 
 (defn initialize-request
   ([]
@@ -35,10 +36,6 @@
    (lsp-json-rpc :initialize
                  (merge {:rootUri (h/file->uri (io/file h/root-project-path))}
                         params))))
-
-(defn initialize-emacs-request []
-  (initialize-request {:initializationOptions default-init-options
-                       :capabilities          emacs-capabilities}))
 
 (defn completion-request [path row col]
   (lsp-json-rpc :textDocument/completion
--- a/cli/integration-test/integration/fixture.clj
+++ b/cli/integration-test/integration/fixture.clj
@@ -13,21 +13,6 @@
      :id (lsp/inc-request-id)}))
 
 (def default-init-options {:lint-project-files-after-startup? false})
-(def full-client-capabilities
-  "Turn on all the client capabilities that clojure-lsp cares about.
-
-  ```
-  (intiailize-request {:initializationOptions default-init-opts
-                       :capabilities full-client-capabilities})
-  ```
-
-  Last updated 2022-01-28."
-  {:experimental {:testTree true}
-   :textDocument {:hover      {:contentFormat ["markdown" "plaintext"]}
-                  :completion {:completionItem {:snippetSupport true}}}
-   :window       {:showDocument {:support true}}
-   :workspace    {:workspaceEdit {:documentChanges true}
-                  :codeLens {:refreshSupport true}}})
 
 (defn initialize-request
   ([]
--- a/cli/integration-test/integration/lsp.clj
+++ b/cli/integration-test/integration/lsp.clj
@@ -154,7 +154,7 @@
           (Thread/sleep 500)
           (recur))))))
 
-(defn await-request [method]
+(defn await-client-request [method]
   (loop []
     (let [method-str (keyname method)
           msg        (first (filter #(= method-str (:method %)) @server-requests))]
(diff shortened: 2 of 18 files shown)


cli/src/clojure_lsp/server.clj
```diff
@@ -488,8 +482,7 @@
     (->> edit
          (coercer/conform-or-log ::coercer/workspace-edit-or-error)
          ApplyWorkspaceEditParams.
-         (.applyEdit client)
-         .get))
```
mainej, line 492: Removing this `.get` was required to get the integration tests passing. Otherwise, they freeze, presumably waiting for this CompletableFuture to resolve, which it never does. This change doesn't seem to break anything because nothing depends on the return value of publish-workspace-edit. I didn't look into how the `client` works in the tests, nor whether there is a way to make it resolve the Future.

cli/src/clojure_lsp/server.clj
```diff
@@ -343,8 +338,7 @@
                                         (.setSignatureHelpProvider (SignatureHelpOptions. []))
                                         (.setCallHierarchyProvider true)
                                         (.setLinkedEditingRangeProvider true)
-                                        (.setCodeActionProvider (doto (CodeActionOptions. (vec (vals coercer/code-action-kind)))
-                                                                  (.setResolveProvider true)))
```
mainej, line 347: The server no longer advertises that it can resolve code actions, which I hope will ensure editors don't try that path 🤞. Of course, this could be turned back on someday, if needed.

ericdallo: I think we still need to keep the CodeActionOptions

cli/src/clojure_lsp/server.clj
ericdallo: I think this get was there to wait for this action to finish, but I think there is no usage waiting for that, so it's probably ok

cli/src/clojure_lsp/server.clj
ericdallo: exactly

cli/integration-test/integration/fixture.clj
```diff
@@ -12,27 +12,47 @@
      :params params
      :id (lsp/inc-request-id)}))
 
+(def default-init-options {:lint-project-files-after-startup? false})
+(def emacs-capabilities
```
ericdallo, line 16: I don't think we need to have capabilities specific for a client like Emacs, AFACS we just use it for code actions bring a lot of actions similar to emacs, but I think we could create different tests acvating different capabilities just like the code actions unit tests do, WDYT?

cli/integration-test/integration/lsp.clj
```diff
@@ -153,3 +153,19 @@
         (do
           (Thread/sleep 500)
           (recur))))))
+
+(defn await-request [method]
```
ericdallo, line 157: Maybe this to let it more clear?
```suggestion
(defn await-client-request [method]
```

ericdallo (commented):
Thanks for the integration tests!

Looks good overall, but we really need to test each code action indeed, I can manually test it on this weekend.

About the resolve-macro double asking, it's probably a bug with this PR, we ask it with `producer/show-message-request`, probably the command is being called twice or something, it's probably a priority check that as is one of the complex commands I mentioned before.

I can test with Calva as well as it's really simple to setup and point to a local clojure-lsp

cli/integration-test/integration/fixture.clj
mainej: I added this because I realized the code action integration tests were getting different responses than Emacs was. It took me a while to figure out that's because the tests default to the minimum client capabilities. Once I understood that it took awhile to track down all the possible capabilities, and then another while to figure out what Emacs was actually using, and then another while to figure out how to write the JSON to turn them on.

So, I sort of like being able to look in one place to see how to turn on all the client capabilities.

But, you're right, this is only used in the code action tests. Maybe it'd be better to inline it into that test? Or maybe change it from `emacs-capabilities` to `full-client-capabilities`, so that it's easy to write other tests for fully featured clients? What do you prefer?

cli/integration-test/integration/lsp.clj
mainej: Agreed, these names are really confusing. I find it hard to keep track of who initiates communications. Anyway, I like your proposed name better. I'll update that.

mainej:
> we really need to test each code action

I ran all the code actions in Emacs, and they all look good now. I found one bug while doing that—the order of the arguments to the `add-require-suggestion` command was wrong, so that was broken for clients that didn't support codeAction/resolve. But, it's fixed now for all clients.

> probably a bug with this PR, we ask it with `producer/show-message-request`

It's funny... `create-test`, also uses `producer/show-message-request` to ask which directory to put the test in, but it doesn't have this problem.

> it's probably a priority check

Will you elaborate?

> I can test with Calva

Thank you so much! What other editors use clojure-lsp? Perhaps I can test some of them.

cli/integration-test/integration/fixture.clj
mainej: Or to put it another way, when I'm writing integration tests, I want to pretend I don't know anything about the code. I want to say "pretend we're in an editor..." not, "pretend we're in an editor which has a very specific set of capabilities that I know will affect the outcome of this test..." That's a job for unit tests, IMO.

mainej:
Answered one of my own questions: [these](https://clojure-lsp.io/clients/) are popular clojure-lsp editors.

cli/integration-test/integration/fixture.clj
ericdallo: I see your point, I think most of the actions just rely on `[:workspace :workspace-edit]` which is a pretty common capability almost (all?) editors have nowadays, I'd still go with just enabling it for the actions we would like to get testing both ways, but having it only on a code actions test is a way too.

ericdallo:
> it's probably a priority

I meant this was one of the reasons I migrated to resolve, during the resolve-macro-as implementation IIRC, but I later changed that refactor a lot and probably should work with your PR, we need to understand why it's not working correctly and if this is a block, I can help with that later if you don't find it before.

> Thank you so much! What other editors use clojure-lsp? Perhaps I can test some of them.

To be honest, the most used ones and important for us to worry are Emacs(lsp-mode), VSCode(Calva) and vim(coc and neovim).

cli/integration-test/integration/fixture.clj
mainej: Makes sense, I'll update it.

mainej:
> I meant this was one of the reasons I migrated to resolve, during the resolve-macro-as implementation IIRC... we need to understand why it's not working correctly

Gotcha. OK, I'll look into why it's repeating the questions.

>  vim(coc and neovim)

I've been meaning to try out one of the modern vim forks, so I'll see if I can get at least one of these installed. If you already have them installed, it may be faster for you to do this test, but I'll try too.

ericdallo:
@mainej I found and [fixed](https://github.com/emacs-lsp/lsp-mode/commit/d3bc47bde5ffc1bace40122a6ec0c6d8b9e84500) the double questions on resolve-macro-as, it turns out we used to rely on clients handling the questions, and that's were one of the reasons this didn't work with commands, but, recently I managed to make it work on server side, I just forgot to remove the question on client side 😅

mainej:
> I found and fixed the double questions

Nice find!

mainej:
> suggest you try neovim

I'll work on getting that running.

mainej:
> we used to rely on clients handling the questions

Do other clients need to be updated too?

ericdallo:
It seems to be working with Calva and Emacs :)

ericdallo:
> Do other clients need to be updated too?

Calva was already updated some weeks ago, there is only emacs and Calva that were implementing this feature, not it's not necessary as it's a command just like others for clients

lib/src/clojure_lsp/feature/code_actions.clj
```diff
@@ -318,182 +111,159 @@
          {:title      (format "Add require '[%s %s %s]'" ns (if alias ":as" ":refer") (or alias (str "[" refer "]")))
           :kind       :quick-fix
           :preferred? true
-          :data       {:id        "add-missing-require"
-                       :uri       uri
-                       :line      (:line position)
-                       :character (:character position)}})
+          :command    {:title     "Add missing require"
+                       :command   "add-missing-require"
```
ericdallo, line 115: This one is not working, it's tricky, the correct should be `add-missing-libspec` for legacy reasons

mainej: Good catch. I thought I tested that, but I must have been testing `add-require-suggestion`, which looks the same in the action menu. Two questions...

1. What's the difference between a "missing require" and a "require suggestion"?
2. Originally, resolve-action called `add-missing-libspec` with `:version 0`. Do you know why and whether that's necessary to maintain?

ericdallo: 1. Clojure-lsp used to have only `add-missing-libspec`, which would add a require that already exists on the project somewhere, the `add-require-suggestion` as the name say suggest some namespaces know via kondo analysis, the code actions are similar to make it more user friendly, and that's why I always suggest relying on the code actions than manual commands.

2. the version is something we use to track how much times the file was changed, via user or lsp, this helps know if the file we are changing now has a version outdated during some async stuff, but for code actions this usually is not a issue, I think any number here would work

mainej: OK, fixed. Thanks again for catching that.

ericdallo (approved):

ericdallo:
I think we can merge this, it's working quite good to me both on calva and emacs, vim follows pretty much emacs so I'm ok merging it since emacs and calva is working

mainej:
Great, thanks @ericdallo! I just did a quick test in neovim. Looks OK there too. 🚀

lib/src/clojure_lsp/feature/code_actions.clj
mainej: I've been thinking about this... Each missing-require action calculates what the alias/refer should be but then throws that away, forcing the server to recalculate it when the action is selected. If those actions also used the `add-require-suggestion` command, which includes the alias/refer in the arguments, we could avoid that recalculation. It'd be a small performance improvement, but maybe nice anyway. We probably couldn't remove the `add-missing-libspec` command entirely—I'm sure some editor or user calls it directly—but it'd become an infrequently used code path.

ericdallo: I don't think we should bother with performance during code action execution unless a huge performance issue, but if you have a way to make all add-require logic in a single command, sounds a good idea.
I know a lot of vim users use that command, so less breaking changes would be better

PR #753, opened 2022-02-09
Exclude vars marked ^:export from unused var check

Deraen:
If a var (defn or def) is marked with `^:export` metadata in Cljs it is probably going to be used from the JS console or from other JS code, and should be considered used.


ericdallo (approved):

Issue #750, opened 2022-02-05
Improve completion performance on comments


ericdallo:
As discussed on this [thread](https://clojurians.slack.com/archives/CPABC1H61/p1643959935463449), we should probably do not return completion items (or just what makes sense?) on comments.

mainej:
I have a lot of context from the thread discussion, so I'll take this. To summarize, when completion is invoked in comments, whitespace, or anywhere it can't find a symbol/keyword zloc, clojure-lsp currently returns every symbol defined in clojure.core, java.lang and java.util. (Note that this is true even before the changes in #716.)

I'm concerned that finding and serializing that large result set causes the performance problems mentioned in the thread. (Note that these problems are probably exacerbated by the particular client mentioned in the thread, nvim+coc.vim, which requests completion for single characters, without much or any delay.)

We think the client provides the text it is trying to complete. But it looks like we ignore that data. So, the proposal is to research that more. If possible, we'll use that text to filter the completions. @ericdallo, should the server _always_ use that text for completion? Or only when it can't find a valid zloc? What's the benefit of finding a zloc at all, if the client provides the text it's trying to complete? It looks like there's special handling of completion inside ns definitions and other scenarios, so maybe it's related to that? I'll have to understand that code better.

Also we'll short-circuit when no prefix string can be found. This will avoid running completion when the cursor is on whitespace, or on forms that aren't symbols or keywords. A variation of this idea is that we could have a setting for the minimum prefix length. If we defaulted that to 2, the result sets would be much smaller. Users that wanted more results could configure it to 1 (or even 0) at their own risk of performance problems. @ericdallo do you think that's worth it? I don't know how hard it is to add configuration options.

As a reminder to myself, it'd be good to benchmark the existing code a little, and keep an eye on how these changes and any other refactorings affect performance.

And finally, we're still exploring the performance issues mentioned in the original thread. If we decide that completion isn't the problem, it may make sense to abandon this line of work.

ericdallo:
@mainej My bad, double checking the SPEC [here](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#completionParams), client only pass uri, row and col, so we don't have any extra text, we do need to parse the URI with that row/col and get the zipper to know what is being completed.

ericdallo:
[...]

PR #756, opened 2022-02-10
Abort completion in comment

mainej:
This PR skips completion when inside of a comment, as discussed in #750.

@emilaasa if you urgently need to turn completion back on, but want to avoid the heavy CPU usage in comments, you can `make` this branch and use the resulting `clojure-lsp` executable. If it's not urgent, you can wait until this is merged. Note that this doesn't address any further performance issues with completion. If we decide to research that more, we should make a follow-up Issue.

@ericdallo this PR is built off #716. Only commit 4b5c5841e3b10b2ce345ad9233707f9090b4283f is relevant to skipping completion in comments. I had to base it off #716 because otherwise the code can't distinguish between comments (where we don't want completion) and whitespace (where we do).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
   - Fix exception during code actions calculation when in a invalid code of a map with not even key-pairs.
   - Don't return diagnostics for external files like files on jar dependencies, avoiding noise on lint when opening dependencies.
   - Support finding implementations of defprotocol and references of defrecord/deftype, implementing LSP method `textDocument/implementation`. #656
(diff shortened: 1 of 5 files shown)


ericdallo:
Thanks @mainej, I didn't have the time to review #716 but I hope I can take a look today before we check this PR

mainej:
This is rebased, so it now contains only the relevant commit. It's ready for review any time.

mainej:
Changelog added.

ericdallo:
Thank you!

PR #757, opened 2022-02-11
Print tail of log when integration tests fail or time out

mainej:
This prints a tail of the log when an integration tests fails, has an error, or times out. The need for this has been discussed in #704.

While implementing this, I noticed that the CLI throws an NPE when shutting down from the "listen" command. This PR also fixes that.

I don't love hard-coding the log file, but it seemed easier than trying to extract it from clojure-lsp, which is running in a separate process and which may have died. If you have other ideas, I'd be happy to try them.

### DIFF
--- a/cli/integration-test/entrypoint.clj
+++ b/cli/integration-test/entrypoint.clj
@@ -50,7 +50,7 @@
     (print-log-tail!))
   (original-report data))
 
-(defmacro report-log-tail
+(defmacro with-log-tail-report
   "Execute body with modified test reporting functions that prints log tail on failure."
   [& body]
   `(binding [original-report t/report
@@ -66,7 +66,7 @@
   (apply require namespaces)
 
   (let [test-results (timeout 600000
-                              #(report-log-tail
+                              #(with-log-tail-report
                                  (apply t/run-tests namespaces)))]
 
     (when (= test-results :timed-out)
--- a/cli/integration-test/entrypoint.clj
+++ b/cli/integration-test/entrypoint.clj
@@ -35,7 +35,7 @@
     ret))
 
 (defn log-tail []
-  (:out (sh/sh "tail" "-n" "300" "/tmp/clojure-lsp.integration.out")))
+  (:out (sh/sh "tail" "-n" "300" "clojure-lsp.integration-test.out" :dir "integration-test/sample-test/")))
 
 (defn print-log-tail! []
   (binding [*out* *err*]
--- a/cli/integration-test/entrypoint.clj
+++ b/cli/integration-test/entrypoint.clj
@@ -35,7 +35,7 @@
     ret))
 
 (defn log-tail []
-  (:out (sh/sh "tail" "-n" "100" "/tmp/clojure-lsp.integration.out")))
+  (:out (sh/sh "tail" "-n" "300" "/tmp/clojure-lsp.integration.out")))
 
 (defn print-log-tail! []
   (binding [*out* *err*]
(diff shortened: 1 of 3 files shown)


cli/integration-test/sample-test/.lsp/config.edn
```diff
@@ -1,3 +1,4 @@
 {:linters {:clj-kondo {:ns-exclude-regex ""
                        :async-custom-lint? false}}
+ :log-path "/tmp/clojure-lsp.integration.out"
```
ericdallo, line 3: Hum, we may have issues with some downstream that build clojure-lsp and run integration tests like [nix](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/tools/misc/clojure-lsp/default.nix) that doesn't accept that kind of absolute path

ericdallo: Maybe we can pass a relative path? like:
```suggestion
 :log-path "clojure-lsp.integration-test.out"
```

cli/integration-test/entrypoint.clj
```diff
@@ -31,6 +32,29 @@
       (future-cancel fut))
     ret))
 
+(defn log-tail []
+  (:out (sh/sh "tail" "-n" "100" "/tmp/clojure-lsp.integration.out")))
```
ericdallo, line 36: I already saw exceptions longer than 100, especially graalvm ones, WDYT about increasing it?
```suggestion
  (:out (sh/sh "tail" "-n" "300" "/tmp/clojure-lsp.integration.out")))
```

cli/integration-test/sample-test/.lsp/config.edn
mainej: 👍 I'll try that to see if it works.

cli/integration-test/entrypoint.clj
mainej: Sure, MORE LOGS!

cli/integration-test/sample-test/.lsp/config.edn
ericdallo: Yes! I hope that works with nix 😅 , I don't see why that wouldn't work now

mainej:
Coincidentally, one of the integration tests for this PR failed: https://github.com/clojure-lsp/clojure-lsp/runs/5150490400?check_suite_focus=true. So, we get to see our first log output. I don't have time to review it right now, but will tomorrow.

It timed out in the integration.formatting-test, much like a failure from a few weeks ago https://github.com/clojure-lsp/clojure-lsp/runs/4988454455?check_suite_focus=true. Perhaps we'll be able to see a pattern there.

ericdallo:
Yeah, taking a quick look, it seems the issue is on babashka side (client side) not on server one, maybe that loop has some point of failure

mainej:
You mean [this loop](https://github.com/clojure-lsp/clojure-lsp/blob/23d1fb8c48d30aae73073de7f7455738ae930bdb/cli/integration-test/integration/lsp.clj#L49-L68)? That's what I've been looking at too. The server seems to think it did everything correctly, but the client never sees a response. There could be a problem in the loop, or maybe there's a bug in lsp4j, where it doesn't flush the server's response.

ericdallo:
This PR looks OK, do you think we should merge this PR already @mainej ?

mainej:
I just renamed a function and rebased on master, so the PR is building again. But otherwise, yes, it's ready to go.

ericdallo (approved):

ericdallo:
Thank you!

PR #758, opened 2022-02-11
Properly position cursor after moving multiline elements 

mainej:
Before this change, when swapping elements the cursor sometimes ended up in the wrong place. This happened when moving down and when one element had more lines than the other. Now the cursor is always moved with the element, ending up at its beginning.

This is particularly important when invoking move up or down repeatedly. If the cursor jumps around between or within elements, repeated invocations can end up moving the wrong thing.


lib/src/clojure_lsp/feature/add_missing_libspec.clj
```diff
@@ -285,29 +284,19 @@
...
-      (concat chosen-require
-             ;; TODO: the intention of the following code appears to be to
-             ;; convert un-aliased symbols to aliased symbols, e.g.
-             ;; `clojure.string/split` -> `my-str/split`. However, when on a
-             ;; `clojure.string/split` node, the UI will only ever suggest
-             ;; 'clojure.string' as an alias. So, in practice, there's never any
-             ;; meaningful change. Proposal: delete this code and rely on
-             ;; "rename" to change aliases. Alternatively, make the UI accept
-             ;; custom aliases.
-              (when chosen-alias
```
mainej, line 305: I've finally understood—in fact, the code actions WILL suggest renames. If you're on `clojure.string/split`, the code actions will suggest "require clojure.string :as string". If you choose that, this code will convert the usage to `string/split`. That's very common, and very useful. I think the comment is more accurate now.

ericdallo: Yes, I forgot I added that some time ago 😅

mainej:
@ericdallo this is ready for your review. The main part of it is the fix for the bug you noticed about moving multiline elements. Now the cursor is reliably positioned in those cases.

I've also included a few unrelated changes. There's a revision of the comments around add-require-suggestion and a changelog entry that should have been part of an earlier PR. If you'd like, I can separate those into their own PRs.

ericdallo:
Code-wise looks good to me, I'll manually test it later tonight

PR #764, opened 2022-02-14
Improve performance of core completions

mainej:
This commit improves completion performance when a large number of symbols in `clojure.core` or `cljs.core` are valid completions.

The fix is to change a quadratic time query into a linear time query. In particular, let the number of elements in the analysis db be `M` and the number of symbols defined in `clojure.core` be `N`. Creating completions for all `N` symbols took `O(M*N)` time, because we ran the analysis query once for each symbol. Now we query the analysis only once, changing run time to `O(M+N)`.

On my machine, when in `clojure-lsp/feature/completion.clj` (i.e. a 500 line file in a medium-to-large project), completing from a whitespace node used to take about 1100ms. Now it takes about 110ms.

This wasn't noticed before because usually `N` is small—i.e. usually there aren't many symbols in `clojure.core` that are valid completions. However, when completion is invoked when the cursor isn't on a symbol, every symbol in `clojure.core` is a valid completion.

This performance problem is probably the true underlying cause of [#750][750]. That bug was patched by [#756][756] because we didn't know the underlying cause at the time. After this patch is merged, it may be possible to revert 756 without any detriment to performance. On the other hand, while completing within a comment may no longer be _slow_, it probably still isn't very _useful_. So, I'd vote for leaving 756 in place.

This bug was diagnosed by wrapping the main parts of `clojure-lsp.feature.completion/completion` in [metrics-clojure][metrics-clojure] timers, and bisecting down to slow parts of the code. That process hasn't unearthed any other major performance problems. I suspect that snippets are slower than they need to be because we unnecessarily re-parse the file. I have a fix for that, but it's in a different branch with several other changes. I'll submit that separately. All that said, the completion code handles several contexts, and I didn't performance test them exhaustively. There may be other improvements to be found.

[750]: https://github.com/clojure-lsp/clojure-lsp/issues/750
[756]: https://github.com/clojure-lsp/clojure-lsp/pull/756
[metrics-clojure]: https://metrics-clojure.readthedocs.io/en/latest/

- [ ] ~I updated documentation if applicable (`docs` folder)~

### DIFF
--- a/lib/src/clojure_lsp/feature/completion.clj
+++ b/lib/src/clojure_lsp/feature/completion.clj
@@ -88,13 +88,16 @@
     :else
     :reference))
 
-(defn ^:private resolve-item-kind [name ns analysis]
-  (->> (mapcat val analysis)
-       (filter #(and (= name (:name %))
-                     (= ns (:ns %))
-                     (= :var-definitions (:bucket %))))
-       first
-       element->completion-item-kind))
+(defn ^:private var-defs-in-ns-named [analysis ns names]
+  ;; TODO: maybe better to use queries/find-all-var-definitions, or to refactor
+  ;; queries so that they are composable as transducer transformations.
+  (let [name-set (set names)]
+    (into []
+          (comp (mapcat val)
+                (filter #(= ns (:ns %)))
+                (filter #(identical? :var-definitions (:bucket %)))
+                (filter #(contains? name-set (:name %))))
+          analysis)))
 
 (defn ^:private element->label [{:keys [alias bucket] :as element} cursor-alias priority]
   (cond
@@ -280,26 +283,30 @@
          (mapv #(element->completion-item % alias :alias-keyword)))))
 
 (defn ^:private with-clojure-core-items [matches-fn analysis]
-  (->> common-sym/core-syms
-       (filter (comp matches-fn str))
-       (map (fn [sym] {:label (str sym)
-                       :kind (resolve-item-kind sym 'clojure.core analysis)
-                       :data (walk/stringify-keys {:filename "/clojure.core.clj"
-                                                   :name (str sym)
-                                                   :ns "clojure.core"})
-                       :detail (str "clojure.core/" sym)
-                       :priority :clojure-core}))))
+  (let [matches (->> common-sym/core-syms (filter (comp matches-fn str)))
+        elem-by-name (->> (var-defs-in-ns-named analysis 'clojure.core matches)
+                          (medley/index-by :name))]
+    (map (fn [sym] {:label (str sym)
+                    :kind (element->completion-item-kind (elem-by-name sym))
+                    :data (walk/stringify-keys {:filename "/clojure.core.clj"
+                                                :name (str sym)
+                                                :ns "clojure.core"})
+                    :detail (str "clojure.core/" sym)
+                    :priority :clojure-core})
+         matches)))
 
 (defn ^:private with-clojurescript-items [matches-fn analysis]
-  (->> common-sym/cljs-syms
-       (filter (comp matches-fn str))
-       (map (fn [sym] {:label (str sym)
-                       :kind (resolve-item-kind sym 'cljs.core analysis)
-                       :data (walk/stringify-keys {:filename "/cljs.core.cljs"
-                                                   :name (str sym)
-                                                   :ns "cljs.core"})
-                       :detail (str "cljs.core/" sym)
-                       :priority :clojurescript-core}))))
+  (let [matches (->> common-sym/cljs-syms (filter (comp matches-fn str)))
+        elem-by-name (->> (var-defs-in-ns-named analysis 'cljs.core matches)
+                          (medley/index-by :name))]
+    (map (fn [sym] {:label (str sym)
+                    :kind (element->completion-item-kind (elem-by-name sym))
+                    :data (walk/stringify-keys {:filename "/cljs.core.cljs"
+                                                :name (str sym)
+                                                :ns "cljs.core"})
+                    :detail (str "cljs.core/" sym)
+                    :priority :clojurescript-core})
+         matches)))
 
 (defn ^:private with-java-items [matches-fn]
   (concat
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
   - Small performance improvement during startup on unused-public-var calculation parallelizing calculations.
   - Small performance improvement on code actions calcullation.
   - Add `:use-source-paths-from-classpath` setting defaulting to true, which makes clojure-lsp do not manually discovery source-paths but get from classpath excluding jar files and paths outside project-root. #752 #551
+  - Improve completion performance when all clojure.core or cljs.core symbols are valid completions. #764 @mainej
   
 - Editor
   - Fix exception during code actions calculation when in a invalid code of a map with not even key-pairs.
--- a/cli/build.clj
+++ b/cli/build.clj
@@ -17,7 +17,7 @@
 
 (defn javac [opts]
   (clean opts)
-  (println "Compiing java classes...")
+  (println "Compiling java classes...")
   (b/javac {:src-dirs ["src-java"]
             :class-dir lsp-class-dir
             :basis (b/create-basis basis)})


lib/src/clojure_lsp/feature/completion.clj
```diff
@@ -279,27 +271,46 @@
                            (string/starts-with? (:name %) name))))
          (mapv #(element->completion-item % alias :alias-keyword)))))
 
+(defn ^:private var-defs-in-ns [analysis ns]
+  ;; TODO: maybe better to use queries/find-all-var-definitions, or to refactor
+  ;; queries so that they are composable as transducer transformations.
+  (->> analysis
+       (mapcat val)
+       (filter #(= ns (:ns %)))
+       (filter #(= :var-definitions (:bucket %)))))
```
ericdallo, line 280: Oh, this one is probably a query should be on queries.clj and we missed

Can we use `identical?` here instead of `=`, move this as the first predicate and convert this to use transducers just like we are using on queries.clj? It should increase even more the performance

lib/src/clojure_lsp/feature/completion.clj
```diff
@@ -279,27 +271,46 @@
                            (string/starts-with? (:name %) name))))
          (mapv #(element->completion-item % alias :alias-keyword)))))
 
+(defn ^:private var-defs-in-ns [analysis ns]
+  ;; TODO: maybe better to use queries/find-all-var-definitions, or to refactor
+  ;; queries so that they are composable as transducer transformations.
+  (->> analysis
+       (mapcat val)
+       (filter #(= ns (:ns %)))
```
ericdallo, line 279: Also, use safe-equal here if moving to queries.clj

ericdallo (commented):
Amazing finding!
I can test this soon, but looks all good to me

lib/src/clojure_lsp/feature/completion.clj
mainej: Yeah, using a transducer is a good idea.

Is `identical?` any different from `=` when comparing keywords? 

I think moving this to queries.clj should be part of a bigger discussion though. queries.clj already has `find-var-definitions`, which is similar. But, it can't be reused because it only works on a single file, which isn't applicable in this case. There's also `find-all-var-definitions`, which doesn't let me filter by namespace soon enough. I'm hesitant to introduce a third variation that's only used in one place. 

I think it'd be nice to separate the pieces of queries.clj into several filter functions that can be recombined in transducers. That is, instead of a function like `find-var-definitions`, which says "find analysis for file AND filter to var definitions", create two functions "find analysis for file" and "filter to var definitions". Then they could be recombined in transducers with `comp` as needed. For example, this completion code could first filter by ns, then comp that with the shared "filter to var definitions" query.

I'd rather not do all of that as part of this PR, but it's something to think about. What do you think?

ericdallo: > Is identical? any different from = when comparing keywords?

Yes, [it has a better peformance](https://groups.google.com/g/clojure/c/cqYUG4b9uWE)

> I'd rather not do all of that as part of this PR, but it's something to think about. What do you think?

Yes, makes sense, it's not that easy since comparing a ns could be different if a var-definition or a var-usage for example.
I'm ok not changing this on this PR

ericdallo:
BTW, I never used metrics-clojure, usually when I want to measure performance I use `criterium`, good to know of an alternative

lib/src/clojure_lsp/feature/completion.clj
mainej: > it has a better peformance

Good to know.

> could be different if a var-definition or a var-usage

Because of aliases? Aliases introduce so much subtlety!

ericdallo: not only alias, but there is `:to`, `:from`, `:ns`, multiple ways to check a namespace depending on the kondo analysis you are iterating

ericdallo (approved):
Thank you!

mainej:
> I never used metrics-clojure, usually when I want to measure performance I use `criterium`, good to know of an alternative

I often reach for `criterium` too. It might have been better for this case. `metrics-clojure` is more useful if you want to monitor performance and throughput in production. It's meant to be long-lived so it doesn't use too much memory, and doesn't introduce too much overhead on hot code. It might be a nice way to monitor things like `f.file-management/analyze-changes` that are executed asynchronously. You could wrap that in a timer, and then periodically [print stats to the logs](https://metrics-clojure.readthedocs.io/en/latest/reporting.html#console).

The underlying [`metrics`](https://metrics.dropwizard.io/4.2.0/) library is really beautiful. I'd recommend the [talk](https://www.youtube.com/watch?v=czes-oa0yik) if you're interested in the subject.

mainej:
@ericdallo In the last few days I heard about [`clj-async-profiler`](https://github.com/clojure-goes-fast/clj-async-profiler) from two different sources. I thought I'd try it out on the completions code. It would have been useful while improving the completion performance–more useful than either `criterium` or `metrics-clojure`.

I haven't gone back to see how much improvement this PR made, though that would be interesting.

To get a feeling for it, here's a flamegraph of the completions, as of a recent build.

15-cpu-flamegraph-2022-02-24-17-52-24

The original image is an interactive SVG, where you can drill into the individual bars. If you'd like I'll Slack it to you.

I got this by profiling the invocations of `completion` from the first test in `clojure-lsp.features.completion-test`. After stripping out all the clojure.test related stuff, I ran them in a loop 400 times. That test exercises the completions code in several different ways, but it's not a perfect sample. So, we shouldn't draw too many conclusions from it. But here's my analysis anyway: 

1. There's a lot of time spent in Java land, compiling. I'm not sure what that's about. Some sort of JIT thing? Might be worth asking around.
2. Drilling down into completion, there's no single thing that dominates the time, which is probably good.
3. Working left to right, only 1% in `sorting-and-distincting-items`. I commented out the "complete from whitespace" invocation, but including it makes the code spend more time in this function, which isn't surprising—it adds a ton of items to sort.
4. 7.6% in `with-clojure-core-items`. That's still quite a lot of time. I think this is mostly because it filters through all of the analysis, regardless of how many symbols match. Still, seems a little high. Some other functions that use big chunks of the analysis also show up in the flamegraph, so maybe there's a theme there.
5. 14.5% in `with-elements-from-alias`. It spends a big chunk of time creating a change set, editing nodes with `add-missing-libspec/add-to-namespace`. I wonder—is that necessary? Is it something that can be postponed, maybe to `completionItem/resolve`?
6. 14.6% and 10% respectively, for parsing and positioning. At 25% of the total time, this is a big chunk. I bet avoiding extra parses helped completion performance a lot too.
7. A smattering of smaller things.
8. 13.4% in `into`. This one's harder to follow, but if you drill into it, it's dominated once again by calls to `add-missing-libspec/add-to-namespace`, invoked from `with-elements-from-alias`.

So, based on this one analysis, I'd focus on `with-elements-from-alias` next, in particular on its call to `add-missing-libspec/add-to-namespace`.

But, using one analysis is pretty naive. I'd want to spend more than an hour analyzing before starting any work. Still, pretty interesting, right? I can definitely see how I'd use this for perf work in the future. I'd like to try out the [differential flamegraphs](http://clojure-goes-fast.com/blog/clj-async-profiler-040/) to measure the impact of changes. Cool stuff!

ericdallo:
Awesome @mainej, It'd be great to use this analysis on other features that I know that have a lot of performance issues like `formatting`/`range-formatting`, this one is the most expensive that affect users ATM, making a lot of them disabling because of paredit actions that trigger range-formatting and lock the file.
Thanks for the analysis!

> 14.5% in with-elements-from-alias. It spends a big chunk of time creating a change set, editing nodes with add-missing-libspec/add-to-namespace. I wonder—is that necessary? Is it something that can be postponed, maybe to completionItem/resolve?

Good point, I think this is something we could move to resolveItem indeed.

mainej:
> `range-formatting`, this one is the most expensive

Yeah, formatting is a great target! I'll plan to profile it when I have a chance.

> move to resolveItem

Looks like you have to check whether the client supports additionalTextEdits in resolveItem. emacs-lsp [does](https://github.com/emacs-lsp/lsp-mode/blob/c30b4b5393b5922d80e2bbbf4bfbfddbf7b00a75/lsp-mode.el#L3495). I haven't looked at the other major clients.

ericdallo:
Yes, most clients support, but we need to check to avoid errors.

> Yeah, formatting is a great target! I'll plan to profile it when I have a chance.

The issue with formatting is that probably the performance cost is on cljfmt itself

mainej:
A bit more research... coc.nvim [supports](https://github.com/neoclide/coc.nvim/blob/e657ec91d853db3e8941df4da883f06e4a4eeb94/src/language-client/client.ts#L1872) additionalTextEdits in completionItem/resolve too. AFAICS the built-in neovim lsp [does not](https://github.com/neovim/neovim/blob/c096561041840826245a82699dfb49064d326c52/runtime/lua/vim/lsp/protocol.lua#L653-L678). I'm not sure about Calva. It doesn't [look](https://github.com/BetterThanTomorrow/calva/blob/9fddb149aba4ec12f600aa96e0bc35e9025f023a/src/providers/completion.ts) like. But it also doesn't look like Calva supports additionalTextEdits on the primary completion endpoint either. The LSP spec seems to assume that all clients will support at least that. Anyway, it'll definitely be necessary to check.

Yeah, my expectation is that most of the cost is in `cljfmt`, but doesn't hurt to check. It's even possible we could find optimizations to submit upstream to `cljfmt`.

ericdallo:
@mainej even so, it's an optional capability, any client that wants to integrate with clojure-lsp can decide to not support that, so us as a service should handle that as optional indeed IMHO

ericdallo:
> but doesn't hurt to check. It's even possible we could find optimizations to submit upstream to cljfmt.

Yes, there were some optimizations done in the past on cljfmt that helped a lot, maybe could be more to improve

mainej:
After researching a bit more, I don't think we'd improve performance much by moving additionalTextEdits to completionItem/resolve.

The flamegraph I made analyzed a few test invocations in what is essentially a tiny project—a small handful of files with only a few lines each. In that world creating the additionalTextEdits is a large contributor to completion time. But in a real project, the additionalTextEdits are lost in the noise of all the other stuff going on.

I haven't had a chance to create flamegraphs for a big project, yet, but my expectation is that there will be much more time spent filtering the analysis. Perhaps that will reveal other optimization targets.

Anyway, I'm going to set aside this part of my performance research, for now.

PR #765, opened 2022-02-16
Ignore kondo config nested under the project root

mainej:
As discussed in this [thread][slack-thread], sometimes users would see
lint findings that didn't match their .clj-kondo config. This happened
both on the command line and in editors.

It was determined that if a user had, for example, both a .clj-kondo
directory and a lib/.clj-kondo directory, the settings in the second
would take precedence when clojure-lsp linted an individual file in
lib/.

The root config is used when invoking `clj-kondo` on the command line,
leading to different results depending on which tool linted the file.

We think that users never intend to have a nested config directory. If
they do, they should delete it, perhaps after merging its settings into
the root config directory.

But, to prevent confusion, this patch makes clojure-lsp ignore nested
config dirs when linting individual files.

- ~I created a issue to discuss the problem I am trying to solve or there is already a open issue.~ See [Slack thread][slack-thread].

[slack-thread]: https://app.slack.com/client/T03RZGPFR/CPABC1H61/thread/CPABC1H61-1644946235.338589


ericdallo:
Thank you!

PR #769, opened 2022-02-16
Try to expose errors on client side of integration tests

mainej:
While researching [flaky integration tests][flaky-tests] we noticed that sometimes a test will [time out][time-out] because the server has sent a response, but the client fails to read it. No exceptions are logged, the tests just hang waiting for the client to do something.

The client runs in a separate thread, looping over and parsing the server's output.

This commit removes an exception handler in the client loop. The handler was using an IOException as a signal that the server's output stream had closed, and that therefore the client could close too. We now use a different mechanism of detecting that the server has closed.

Though this is speculative, we hope is that this will expose unexpected IOExceptions, which could have caused the client to stop responding.

[flaky-tests]: https://github.com/clojure-lsp/clojure-lsp/issues/704
[time-out]: https://github.com/clojure-lsp/clojure-lsp/pull/757#issuecomment-1035760361

### DIFF
--- a/cli/integration-test/integration/initialize_test.clj
+++ b/cli/integration-test/integration/initialize_test.clj
@@ -94,4 +94,7 @@
     (lsp/await-notification :$/progress))
 
   (testing "initialized notification"
-    (lsp/notify! (fixture/initialized-notification))))
+    (lsp/notify! (fixture/initialized-notification))
+    (h/assert-submaps
+      [{:registerOptions {:watchers [{:globPattern "**/*.{clj,cljs,cljc,edn}"}]}}]
+      (:registrations (lsp/await-client-request :client/registerCapability)))))
--- a/cli/integration-test/integration/lsp.clj
+++ b/cli/integration-test/integration/lsp.clj
@@ -43,6 +43,12 @@
 
 (defn ^:private keyname [key] (str (namespace key) "/" (name key)))
 
+(defn client-log
+  ([color msg params]
+   (client-log @client-id color msg params))
+  ([client-id color msg params]
+   (println (colored color (str "Client " client-id ": " msg)) (colored :yellow params))))
+
 (defn ^:private listen-output! []
   (let [client-id (swap! client-id inc)]
     (future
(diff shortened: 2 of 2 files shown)


mainej:
After a bit of experimentation, this includes some additional changes:
1. Tries a different approach to shutting down the client thread. The prior approach with `async/close!` left many IOExceptions in the logs, perhaps because `async/close!` doesn't actually stop an `async/thread`. This approach, using a future and `future-cancel`, has cleared up the IOExceptions, or at least so far.
2. Logs the client shutdown process, because this happens in another thread and can be confusing to track while other clients are starting up
3. Logs which client process is sending or receiving messages, again, because the clients live in many threads. Most of the time there's only one client, but they can overlap during startup and shutdown.

mainej:
I think the changes in this PR are reasonable, but sadly, we're still getting [timeouts](https://github.com/clojure-lsp/clojure-lsp/runs/5223893108?check_suite_focus=true), with no additional information. It doesn't appear that the client thread is crashing, it's just stuck waiting for a message that the server thinks it's sent and which the client never receives.

So, I think we have to move on to looking at other parts of the code. My best bet now is that it's some sort of i/o buffering problem. Perhaps `clojure-lsp.server/run-server!` which connects the input and output streams to the server? Or perhaps `integration.lsp/start-process!` which connects the server's i/o to the client's i/o? For that second path, we might find inspiration in [lsp4j's tests](https://github.com/eclipse/lsp4j/blob/main/org.eclipse.lsp4j/src/test/java/org/eclipse/lsp4j/test/launch/LauncherTest.java) which sets up the client and server using Piped streams, using their [`.connect`](https://docs.oracle.com/javase/8/docs/api/java/io/PipedInputStream.html#connect-java.io.PipedOutputStream-) method.

Amazingly, while testing this, I actually saw it happen locally. On one test run, right after the first request, suddenly everything froze. I wasn't sure how to take advantage of the situation, and ten minutes elapses quickly when you're scrambling for ideas. All I was able to do was sample the processes. Looking at those samples, there isn't much of interest. You can see what I assume is the client blocking on `readBytes  (in libjava.dylib) + 340`, but I don't see anything else illuminating. The stacktraces mostly look like `???  (in <unknown binary>)`. I think all the Clojure code appears as `<unknown binary>`, which isn't very helpful. I can provide the samples, but I'm not sure they'd be very interesting. If it happens again, I'll also try to run `lsof` for the processes. I'd love guidance on other debugging tools for these scenarios.

mainej:
I take back what I said... Perhaps we are getting errors in the client loop. Originally I saw that they were being printed, and so assumed that errors in threads are always printed. But this isn't true. Something I changed—perhaps switching from async/thread to future—suppressed the error printing. That would certainly explain why the errors stopped appearing in the logs after that switch. Anyway, error printing is back on. We'll see what appears in the logs.

mainej:
I finally had a test [fail](https://github.com/clojure-lsp/clojure-lsp/runs/5242985842?check_suite_focus=true#step:7:111) with the additional client logging. For reference, here's a [passing](https://github.com/clojure-lsp/clojure-lsp/runs/5242985864?check_suite_focus=true#step:7:111) version of the same test, on the same commit.

AFAICS, neither the client's nor the server's logs explain what's going on. So, I'm at a dead-end researching the client loop. The changes in this PR could expose issues in the future, so it could be merged @ericdallo. But right now it doesn't explain what's making the tests flaky.

My hypothesis is back to what I said [before](https://github.com/clojure-lsp/clojure-lsp/pull/769#issuecomment-1042488016)... perhaps there's some I/O buffering problem elsewhere in the code. I don't really know how to research that, so I'm going to put this aside for now.

As one last observation, earlier I said I saw the problem happen locally. At the time I thought it was a "random" occurrence, as we see on CI. But actually it wasn't random at all. Just before it happened I had intentionally inserted a `throw` inside the client loop. Nothing was printed, and so I assumed I must have hit the flakiness. The behavior was exactly as we see in CI—everything frozen, but no errors logged. But, in fact, another change I had recently made meant that the client errors had stopped being printed, a problem I fixed in a53a764. With the printing back in place, and with an intentional throw in the client loop, a client error is logged, different from CI.

All that is to say, there is still a small chance that the problem is in the client loop. If the loop stops, we get the exact symptoms we see in CI. I don't know how it could stop without throwing an error, but perhaps somehow it does.

ericdallo:
That's is becoming hard to debug indeed, I think we can try merge this PR and keep an eye on other PRs when the flakiness happen

ericdallo (approved):

PR #771, opened 2022-02-18
Fix and accelerate clojure.core completion

mainej:
Make completion of clojure.core more consistent by using queries. Returns slightly fewer results, because of de-duplication. Also, change filtering order to slightly improve performance of completion, especially for completions with more given characters.

This is an extension of #764.

### DIFF
--- a/lib/src/clojure_lsp/queries.clj
+++ b/lib/src/clojure_lsp/queries.clj
@@ -412,32 +412,28 @@
     (catch Throwable e
       (log/error e "can't find references"))))
 
+(defn xf-var-defs [include-private?]
+  (comp
+    (filter #(and (identical? :var-definitions (:bucket %))
+                  (or include-private?
+                      (not (get % :private)))))
+    (medley/distinct-by (juxt :ns :name :row :col))
+    (remove #(or (and (safe-equal? 'clojure.core/defrecord (:defined-by %))
+                      (or (string/starts-with? (str (:name %)) "->")
+                          (string/starts-with? (str (:name %)) "map->")))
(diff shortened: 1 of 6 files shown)


lib/src/clojure_lsp/queries.clj
```diff
@@ -412,32 +412,28 @@
     (catch Throwable e
       (log/error e "can't find references"))))
 
+(defn xf-var-defs [include-private?]
```
ericdallo, line 415:
```suggestion
(defn ^:private xf-var-defs [include-private?]
```

mainej: It's used by another ns.

ericdallo: yeah, I just saw it

ericdallo:
Code looks good!
Would you mind adding integration tests for the clojure.core/cljs.core completion, the integration tests are simple ATM?

mainej:
> Would you mind adding integration tests
Yep, happy to.

mainej:
@ericdallo I added integration tests. They dump a TON of data into the test output. The entire response from the server is logged. I can't think of a way to suppress that logging. Do you still want them? IMO, the unit tests cover the same thing. 

On that subject, I noticed the unit tests were missing a test for cljs.core completion, so I added that too.

ericdallo:
I see... Well, I'm ok adding only unit for this since these changes affect only completion data and not something a integration test would catch but a unit not.

ericdallo (approved):
Thank you!

PR #772, opened 2022-02-18
Tidy require/import code actions and refactorings

mainej:
The aim of this commit is not to change any behavior but to simplify and clarify several pieces of code, especially around the generation of actions. I was inspired to try to clean up the actions by a [comment](https://github.com/clojure-lsp/clojure-lsp/pull/716#discussion_r804280479) @ericdallo made in a previous PR.

First, in `add-missing-libspec` and `code-actions` it reorganizes the code related to generating actions. To me, this code is much more succinct now, though of course that's subjective.

Second, in the `parser`, it separates parsing from cursor positioning. When these two responsibilities are separate, they are easier to compose.

Related to this separation, there's a small performance improvement for code action generation. We now avoid parsing the code once per combination of diagnostic and possible resolution, and instead parse once, repositioning the cursor as necessary.

There's also a small performance improvement in snippet generation, where we avoid parsing the file a second time. This fix was mentioned in [#764][764].

The parser refactoring exposed the fact that `parser/safe-zloc-of-string` was actually only _mostly_ safe. It recovered from several errors related to parsing incomplete Clojure, but would still throw if the string was invalid Clojure. Its name has been updated, and a new version added that will return nil for invalid Clojure. This safe version should be used in more places, with the calling code handling nils instead of relying on exceptions to be thrown. This additional cleanup hasn't been completed.

[764]: https://github.com/clojure-lsp/clojure-lsp/pull/764

### DIFF
--- a/lib/src/clojure_lsp/feature/completion.clj
+++ b/lib/src/clojure_lsp/feature/completion.clj
@@ -323,9 +323,8 @@
                          :detail (str "java.util." sym)
                          :priority :java})))))
 
-(defn ^:private merging-snippets [items cursor-loc matches-fn text row col settings]
-  (let [next-loc (parser/safe-loc-at-pos text row (inc col))
-        snippet-items-by-label (->> (concat
+(defn ^:private merging-snippets [items cursor-loc next-loc matches-fn settings]
+  (let [snippet-items-by-label (->> (concat
                                       (f.completion-snippet/known-snippets settings)
                                       (f.completion-snippet/build-additional-snippets cursor-loc next-loc settings))
                                     (map #(assoc %
@@ -356,11 +355,13 @@
        not-empty))
 
 (defn completion [uri row col db]
-  (let [{:keys [text]} (get-in @db [:documents uri])
-        cursor-loc (when-let [loc (parser/safe-loc-at-pos text row col)]
+  (let [root-zloc (parser/safe-zloc-of-file @db uri)
+        ;; (dec col) because we're completing what's behind the cursor
+        cursor-loc (when-let [loc (some-> root-zloc (parser/to-pos row (dec col)))]
                      (when (or (not (-> loc z/node meta))
                                (= row (-> loc z/node meta :row)))
-                       loc))]
+                       loc))
+        next-loc (some-> root-zloc (parser/to-pos row col))]
     ;; When not on a symbol or keyword, we want to return all valid completions
     ;; (almost 1000 in an empty file), even though it's expensive to compute.
     ;; The one exception is in comments. Some editors (nvim + coc.nvim) request
@@ -443,7 +444,7 @@
 
                       (and support-snippets?
                            simple-cursor?)
-                      (merging-snippets cursor-loc matches-fn text row col settings)))]
+                      (merging-snippets cursor-loc next-loc matches-fn settings)))]
         (sorting-and-distincting-items items)))))
 
 (defn ^:private resolve-item-by-ns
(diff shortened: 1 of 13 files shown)


lib/src/clojure_lsp/feature/completion.clj
```diff
@@ -356,11 +355,13 @@
        not-empty))
 
 (defn completion [uri row col db]
-  (let [{:keys [text]} (get-in @db [:documents uri])
-        cursor-loc (when-let [loc (parser/safe-loc-at-pos text row col)]
+  (let [root-zloc (parser/safe-zloc-of-file @db uri)
+        ;; (dec col) because we're completing what's behind the cursor
+        cursor-loc (when-let [loc (some-> root-zloc (parser/to-pos row (dec col)))]
```
ericdallo, line 360: Coud we have any different behaviors we are not aware because of this `(dec col)` that we were not doing before, I wonder what issues we could solve with that `dec` or any issues we could cause

lib/src/clojure_lsp/parser.clj
```diff
@@ -76,36 +76,36 @@
...
-      (if-let [node (handle-end-slash-code text e)]
-        node
-        (if-let [node (handle-keyword-with-end-slash-code text e)]
-          node
-          (if-let [node (handle-single-colon-code text e)]
-            node
-            (throw e)))))))
+      (or (handle-end-slash-code text e)
+          (handle-keyword-with-end-slash-code text e)
+          (handle-single-colon-code text e)
```
ericdallo, line 85: Way better 💯

lib/src/clojure_lsp/feature/completion.clj
mainej: It's hard to follow, but this positioning code hasn't changed. 

There used to be two functions, `parser/loc-at-pos` and `parser/safe-loc-at-pos`. From their names, I would expect them to do the same thing. But actually, the first used `col` as is, while the second one did a `(dec col)`. That's confusing by itself, but even more confusing was that elsewhere some code did an `(inc col)` before calling `parser/safe-loc-at-pos` to counteract the `dec`.

Since this is the only spot that needs a `dec`, I decided to move it here, out of the parser.

The old function definition is [here](https://github.com/clojure-lsp/clojure-lsp/pull/772/files#diff-3e2731e79f3889675b6b04cd07fffb3c2b2cce1baa0a60e66b481ed22344e11cL98) if you want to look at it. Originally there were only two places that called `parser/safe-loc-at-pos`, here and the one that did the `inc` to counteract, so it seemed simple to refactor.

ericdallo: 🤯  I really forgot about that, yeah, there is a issue open to add a coerced layer to input just like we have for output to avoid those crazy inc/decs

ideally, the incs/decs should happen at coercer level, both on borders, like `server.clj`, but while we don't have input coercer, we have most at handlers.clj.
Also, that's why we have the naming `line` `character` to represent lsp lines/cols that are `inc`ed and `row` `col` to represent both clojure-lsp/clj-kondo lines/cols  `dec`ed

ericdallo: I'm ok keeping it that way for now until we have input coercers

mainej: > input coercers

Nice! I was wondering if input coercion was in the works. There's definitely a cognitive overhead to keeping track of both ways of numbering. This PR also has a [fix](https://github.com/clojure-lsp/clojure-lsp/pull/772/files#diff-c4f2705d9ea712285fa2f27dcfacf16733b610505fc7d60229ffd9896f5d5a23R425) where we were using the name "row" when we really had a "line". They're easy to mix up. There are also still lots of places that use "column" when they mean "character", which is easy to confuse with "col".

ericdallo (approved):
Thank you!

PR #780, opened 2022-02-22
Move entries intuitively in `case`, `cond` and similar functions

mainej:
Some functions are known to take a set of clauses which are binary expressions. For example, `case` has clauses made up of a test-constant and a result-expr. `move-coll-entry` should move these clauses together.

```clojure
(case a
  |1 :one
  2 :two)

;; should become

(case a
  2 :two
  |1 :one)
```

This patch makes clauses move together for `assoc`, `assoc!`, `case`, `cond`, `condp`, `cond->`, and `cond->>`. It handles the optional default expression in `case` and `condp`.

It's common to find some of these functions within threading macros, and so this patch handles those situations too. I don't use that syntax often so I'm not sure I've covered all the cases.

It also handles the [ternary expression syntax](https://clojuredocs.org/clojure.core/condp) of `condp`. I believe that usage is rare in the wild, but this code understands it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
   - Fix cljfmt settings merge during refresh/classpath configs merge to avoid multiple config vectors on same symbol.
 
 - Editor
-  - Fix wrong args on extract function from multi-arity fn. #683
+  - extract-function: Fix wrong args when extracting from multi-arity fn. #683
+  - move-coll-entry: clauses move intuitively in `assoc`, `case`, `cond`, and similar functions. #780 @mainej
 
 ## 2022.02.23-12.12.12
 
--- a/lib/src/clojure_lsp/feature/move_coll_entry.clj
+++ b/lib/src/clojure_lsp/feature/move_coll_entry.clj
@@ -9,25 +9,33 @@
 
 (set! *warn-on-reflection* true)
 
-(defn- count-siblings-left
-  "Count the number of sibling nodes to the left of the child node `zloc`."
-  [zloc]
-  (->> zloc
-       (iterate z/left)
-       (take-while (complement z/leftmost?))
-       count))
+(defn ^:private tag-p [tag-set]
+  (comp tag-set z/tag))
 
-(defn- count-siblings-right
-  "Count the number of sibling nodes to the right of the child node `zloc`."
-  [zloc]
-  (->> zloc
-       (iterate z/right)
-       (take-while (complement z/rightmost?))
-       count))
+;;;; Redefine core helpers to treat uneval as comment. See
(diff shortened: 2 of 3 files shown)


ericdallo:
Amazing! I'll leave it to review soon after the next release

mainej:
That's good @ericdallo, because this PR is not quite ready. I submitted it and then started discovering lots of little edge cases. I have one more to fix. I'll leave a comment when it's ready for review.

mainej:
Added: pairwise movement in `assoc` and `assoc!`.

mainej:
clj-rewrite doesn't treat uneval nodes, what it calls `#_`, as whitespace. 🤯 That means none of the "whitespace aware" movement helpers like `z/right` behave like I expect them to.

ericdallo:
@mainej maybe you can ask on #rewrite-clj Clojurians slack, usually @lread helps a lot with those doubts

mainej:
> ask on #rewrite-clj

Good idea. Question [asked](https://clojurians.slack.com/archives/CHB5Q2XUJ/p1645662848468789).

mainej:
@lread clarified: there's an [existing clj-rewrite issue](https://github.com/clj-commons/rewrite-clj/issues/70) about skipping uneval nodes. He said that—as a workaround—copying the navigation helpers is OK. I updated the comments with a link to that issue, but otherwise I don't think there are any other changes that belong in this PR. So, it's ready for review.

My sense is that a lot—or maybe all—of the clojure-lsp code expects uneval nodes to be skipped when navigating. I'd be nice to configure the skipping logic at zipper construction time, as suggested in https://github.com/clj-commons/rewrite-clj/issues/70. We might benefit by offering patch for that issue, or at least by moving the helpers in move-coll-entry into a shared namespace. But, those fixes can be explored in a separate Issue.

ericdallo (approved):
Looks good, thanks for the detailed tests!

ericdallo:
We can merge after solving the changelog conflict

mainej:
I'm not seeing any conflicts. That is, I can cleanly merge this branch with master. But for some reason it looks like Github can't do an automatic rebase. Would you like me to do a manual rebase and force push that?

mainej:
Oh... I bet I know what happened. There _was_ a merge conflict momentarily, but the `master` commit that caused the conflict was reverted. So now there's no conflict again.

ericdallo:
Yes, it seems, maybe a force push will solve it

mainej:
@ericdallo force-pushed and ready for merge, I believe.

ericdallo:
Maybe we could add a gif on features.md later?
Thank you!

PR #789, opened 2022-03-02
Fix wrong args on extract function from multi-arity fn

mainej:
This bumps clj-kondo to include a fix that adds scope-end to the args of multi-arity fns. This fix ensures that these args are included as params when a new function is extracted from their body, fixing #683.

Also includes some test changes to avoid false negatives from a new diagnostic in clj-kondo.


ericdallo (approved):
Thank you!

PR #790, opened 2022-03-02
Add alias and build for profiling and benchmarking

mainej:


cli/deps.edn
```diff
@@ -32,6 +32,11 @@
                  :jvm-opts ["-Xmx2g" "-server"]}
            :debug {:extra-deps {nrepl/nrepl {:mvn/version "0.9.0"}
                                 cider/cider-nrepl {:mvn/version "0.28.3"}}}
+           :perf {:extra-deps {criterium/criterium {:mvn/version "0.4.6"}
```
ericdallo, line 35: Maybe just rename to a better name consistency?
```suggestion
           :performance {:extra-deps {criterium/criterium {:mvn/version "0.4.6"}
```

ericdallo (approved):
Awesome!

cli/deps.edn
mainej: 👍  done.

Issue #332, opened 2021-02-21
Completion is not available before reader conditional


zane:
**Describe the bug**
In a `.cljc` file that contains a reader conditional before the reader conditional, but is available after it.

**To Reproduce**
Create a directory containing the following files:
`deps.edn`:
```
{}
```
`src/repro.cljc`:
```
(ns repro)

#?(:clj "clojure" :cljs "clojurescript")
```
1. In Emacs open `repro.cljc`.
2. Enable `clojurec-mode` and `lsp-mode`.
3. Insert a new line below the reader conditional in `repro.cljc`.
4. Type `t r a` and observe that completion is available at this point in the file.
5. Insert a line above the reader conditional.
6. Type `t r a`.

**Expected behavior**
I would expect to get the same completion options as I did in step 4.

**Screenshots**

![image](https://user-images.githubusercontent.com/26162/108611572-b483b900-73ad-11eb-918f-0657b00990cc.png)

![image](https://user-images.githubusercontent.com/26162/108611581-d54c0e80-73ad-11eb-859d-7a803681d83b.png)

<details>
<summary><b>Log - client <-> server</b></summary>
<p>Here is the contents of <tt>*lsp-log*</tt></p>
<pre>
Found the following clients for /Users/zane/projects/035a1e4d6d58b5e007e6854fe48f54a9/src/repro.cljc: (server-id clojure-lsp, priority 0)
The following clients were selected based on priority: (server-id clojure-lsp, priority 0)
Creating watch for /Users/zane/projects/035a1e4d6d58b5e007e6854fe48f54a9
Creating watch for /Users/zane/projects/035a1e4d6d58b5e007e6854fe48f54a9/src
Cancelling textDocument/codeAction(18) in hook after-change-functions
Cancelling textDocument/codeAction(27) in hook after-change-functions
Cancelling textDocument/codeAction(32) in hook after-change-functions
Cancelling textDocument/codeAction(47) in hook after-change-functions
Cancelling textDocument/codeAction(51) in hook after-change-functions
</pre>
<p>Interaction:</p>
<pre>
[Trace - 07:08:46 PM] Received response 'initialize - (1)' in 427ms.
Result: {
  "capabilities": {
    "semanticTokensProvider": {
      "full": true,
      "range": true,
      "legend": {
        "tokenModifiers": [],
        "tokenTypes": [
          "type",
          "function",
          "macro"
        ]
      }
    },
    "callHierarchyProvider": true,
[...]

PR #793, opened 2022-03-03
Economy of motion

mainej:
Almost every clojure-lsp command begins by navigating to a cursor position. The current algorithm uses `z/next*` to find that position. This is very inefficient, especially for large files.

Suppose, for example, we want to navigate to `b-a-c` within the following code:

```clojure
(a
 (a-a
  (a-a-a)
  (a-a-b)
  (a-a-c))
 (a-b
  (a-b-a)))

(b
 (b-a
  (b-a-a)
  (b-a-b)
  (b-a-c)) ;; <-- we want to get here
 (b-b
  (b-b-a)))
```

If we use `z/next*`, we look depth first through the whole tree:
```
a
a-a
a-a-a
a-a-b
...
b-a-a
b-a-b
b-a-c
```

But we can be more economical...  we know that `a` spans lines 1 through 7. But we're trying to get to `b-a-c` on line 13. Neither `a` **nor any of its children** will be `b-a-c`, because `a` doesn't span `b-a-c`'s position.

This PR changes the navigation algorithm to skip `z/right*` over nodes that can't contain the cursor:
```
a ;; skip a's children
b
b-a
b-a-a
b-a-b
b-a-c
```

On an smallish (56 line) file, this new algorithm is a bit faster (1.2ms 95% 1.15-1.23ms vs 8.1ms 95% 7.9-8.3ms) when searching to a position near the end of the file.

On a long file (~2000 lines), the effect is much more prominent. The original algorithm takes about 50ms regardless of where in the file you search to. (Early in the file takes 50.0ms 95% 49.3-50.6ms, middle 50.1ms 95% 49.5-50.9ms, late 51.1ms 95% 49.2-53.5ms.) The new algorithm takes essentially 0 time to search to the beginning of the file (4.2 µs, 95% 4.18-4.29µs) and follows a linear growth from there, based on how far into the file you need to search (5.9ms 95% 5.7-6.2ms halfway through the file and 11.5ms 95% 11.4-11.7ms near the end of the file).

There are a several things that need to be done before this PR is ready:
1. The new algorithm doesn't have any special handling of reader macros, which the original navigation algorithm did. The tests all pass, so perhaps this is OK.
2. The new algorithm searches right from the leftmost position. Even better would be a binary search, starting in the approximate middle of the file and moving left, right or down as appropriate. I'd expect this to make the new algorithm logarithmic in the row of the cursor, instead of linear.
3. Perhaps this style of position-aware search belongs in rewrite-clj. And, if it can be made to work, perhaps binary search belongs in a general zipper library.
4. There's cleanup to do. The code is split up into more functions than necessary to facilitate benchmarking. The function names also need more thought.
5. The helper named `by-heritability` doesn't feel very "zipper-like". Perhaps there's a more elegant way to express the algorithm.


ericdallo:
Thank you! that looks like a good tuning indeed.

> 1. The new algorithm doesn't have any special handling of reader macros, which the original navigation algorithm did. The tests all pass, so perhaps this is OK.

AFAIK we don't have any logic applied to reader macros, actually, I found even clj-kondo ignore everything tagged with a reader macro, maybe this is a issue we should report later.

> 3. if it can be made to work, perhaps binary search belongs in a general zipper library.

Yes, this really sounds like a capability/function rewrite-clj could offer to clients IMO

lib/src/clojure_lsp/refactor/edit.clj
```diff
@@ -42,6 +42,42 @@
...
+
+(defn find-last-by-pos-next [zloc f pos]
+  (let [forms (z-filter zloc f #(liberal-in-range? % pos))
+        disconsider-reader-macro? (and (some #(= "?" (z/string %)) forms)
+                                       (> (count forms) 1))]
+    (last
+      (cond->> forms
+        disconsider-reader-macro? (remove (comp #(= "?" %) z/string))))))
+
 ;; From rewrite-cljs
```
ericdallo, line 81: this function was included before rewrite-cljs was merged into rewrite-clj, maybe we should check if there is already a function that does this on rewrite-clj itself

ericdallo: Otherwise, maybe we should a test for that function even knowing other tests would fail, a test testing the corner cases here seems to make sense, WDYT?

mainej: rewrite-clj has `find-last-by-pos`, which perhaps it got in the merge with rewrite-cljs. But that function uses the same depth-first approach that we're trying to avoid here. I looked around rewrite-clj pretty carefully to see if there was another faster algorithm, but didn't find anything.

mainej: > testing the corner cases here

I'm sorry, I don't follow. What should be tested?

mainej:
>> position-aware search belongs in rewrite-clj

> Yes, this really sounds like a capability/function rewrite-clj could offer

One reason rewrite-clj might not want this code is that they already have `find-last-by-pos`. That covers the case where you need to find the last node in a range, which is more general than our case, where you need to find _the_ node at a specific position. Because our use-case is narrower, we can make this kind of optimization. Still, I may submit it to rewrite-clj anyway. If they merge it, we can change this code to use their function.

>> binary search

I decided to skip binary search for now... It would make more sense if the zipper was laid out differently, with fast access to children near the middle. But after descending `z/down*`, the locs to the right are all connected as Cons cells. So, to get to the middle you have to traverse that list anyway. It could still be beneficial in really huge files, but we're talking about avoiding a small handful of numerical comparisons, which seems like overkill.

mainej:
Aside from my questions above, this is ready for review and merge.

lib/src/clojure_lsp/refactor/edit.clj
ericdallo: no problem, so there is no function that does what we need there, so my point is even if we know there are other unit tests that end up using this function, maybe we should have tests for this function especially too?

ericdallo:
> One reason rewrite-clj might not want this code is that they already have find-last-by-pos. That covers the case where you need to find the last node in a range, which is more general than our case, where you need to find the node at a specific position. Because our use-case is narrower, we can make this kind of optimization. Still, I may submit it to rewrite-clj anyway. If they merge it, we can change this code to use their function.

Great

I decided to skip binary search for now... It would make more sense if the zipper was laid out differently, with fast access to children near the middle. But after descending z/down*, the locs to the right are all connected as Cons cells. So, to get to the middle you have to traverse that list anyway. It could still be beneficial in really huge files, but we're talking about avoiding a small handful of numerical comparisons, which seems like overkill.

Sounds good, ok for me

lib/src/clojure_lsp/refactor/edit.clj
mainej: Ah, yes, I understand. And it's a good idea!

I've added several tests for `edit/find-at-pos` that confirm that it navigates to the correct node, accounting for as many edge cases as I can think of.

`edit/find-at-pos` was originally called `edit/find-last-by-pos`. That function had special handling for two corner cases: incomplete nodes and the `#?` reader macro.

First let's talk about the reader macro, i.e. code like `#?(:clj (+ 1 2))` or `#?@(:clj (+ 1 2))`. I don't think we need special handling for it anymore. It looks like the special handling was added as part of fixing #332, which was about completion in files with reader macros. The tests that were added as part of #332 still pass, even without any special handling. I suspect that the special handling was no longer necessary as soon as rewrite-clj learned to parse reader macro nodes, around v1, which was soon after the fix for #332 was applied. So, I think we're safe to remove the special handling. Does that sound right @ericdallo?

I've extended the tests around incomplete nodes, code like `foo/`. Those tests exposed some problems with how my changes handled things. The fix was to ensure that all nodes, even ones that initially appear incomplete, have position metadata. (See `parser/replace-preserving-meta`.)  That moves all the special handling out of `edit` and into `parser`, so I added tests around `parser/safe-zloc-of-string` and `parser/to-loc`. In the process, I also fixed a long-standing bug where nodes that followed an incomplete node would have incorrect column numbers. (See `parser/zero-width-space` and the way that it is used to keep the column numbers correct.)

ericdallo:
I'll try to take a look today or tomorrow! sorry for the delay

mainej:
Sounds good... I'm getting the lint tests passing, but otherwise it's ready for review.

lib/src/clojure_lsp/refactor/edit.clj
ericdallo: > Does that sound right @ericdallo?

Yes! Since the tests I added are still passing looks good to me

lib/src/clojure_lsp/parser.clj
```diff
@@ -27,54 +27,65 @@
...
     (when (or (re-matches #"\[line (\d+), col (\d+)\] A single colon is not a valid keyword." cause)
               (re-matches #"\[line (\d+), col (\d+)\] Invalid keyword: ." cause))
-      (let [colon-pattern (re-pattern ":(\\s|\\n|\\))")]
-        (when-let [[_ extra-token] (re-find colon-pattern text)]
-          (-> text
-              (string/replace-first colon-pattern (str ":___" extra-token))
-              z/of-string
-              (z/edit->
-                (z/find-next-value z/next :___)
-                (z/replace (n/token-node (symbol ":"))))
```
ericdallo, line 56: One of the biggest workarounds I did on clojure-lsp I think, glad it's going away 🎉

ericdallo: Oh, it's not going away, but you improved the code, nice

lib/src/clojure_lsp/parser.clj
```diff
@@ -27,54 +27,65 @@
        (mapv (fn [loc] (z/find loc z/up edit/top?)))
        (distinct)))
 
+(def ^:private zero-width-space
+  "A unicode character that is incredibly unlikely to be used in regular code.
+  During parsing, used as a valid and easily identified subsitute for what would
+  otherwise be an invalid character."
+  "\u200b")
```
ericdallo, line 34: not sure a unicode code is the best choice here, not sure this could bring other issues, but it may worth the try

ericdallo (approved):
Looks good codewie, I'll merge it so we can keep testing this on master until next release, thank you!

ericdallo:
Oh, could you add a entry later on general changelog @mainej ?

lib/src/clojure_lsp/parser.clj
mainej: Yeah, I hear you. I'd be OK using another character if you have a suggestion. I picked this one because:
1. Code rarely uses any kind of whitespace besides tabs, spaces and linebreaks.
2. I can't think of a reason code would use it anyway, since text editors are supposed to give this character "zero-width", i.e. make it invisible. In typesetting there's probably some use for it, but not in code.
3. rewrite-clj parses it as a symbol, or part of a symbol, not as whitespace.

The fact that rewrite-clj is doing what we want suggests that it (and Java) assume that the code uses unicode characters, which is probably necessary because people put emojis and other stuff in code. I've never looked into how Java tracks how files and strings are encoded—if it works, best not to touch it I guess.

lib/src/clojure_lsp/parser.clj
mainej: Yeah, not going away, just a refactoring to make sure that incomplete nodes and their neighbors get position metadata. Actually, it's a pretty elegant hack. The [tree-sitter docs](https://tree-sitter.github.io/tree-sitter/#underlying-research) talk about how to make parsers that can recover from errors, which use a similar technique.

Issue #266, opened 2021-01-21
slow formatting of fairly trivial file


gdanov:
while triaging [this issue](https://github.com/emacs-lsp/lsp-mode/issues/2537#issuecomment-762843091) we saw clojure-lsp needs half a second for formatting 2k lines of identical & very trivial expression. This is slow according to my *expectations* ... I'm not aware of benchmark or something that sets the baseline what's 'normal'. If this is the best performance achievable maybe something must be done on lsp-mode side to mitigate.

I'm creating this issue as requested by @ericdallo to track the potential performance problem.

ericdallo:
Thanks for the report!
Clojure-lsp just leave the format to cljfmt, do you have a sample file where you notice the slowliness?
I suspect this happen with huge forms/functions, if you have the same file with multiple not big functions this will be fast

ericdallo:
@gdanov, I may found the issue, but have no idea how we could improve that, the issue is that when clojure-lsp format a range, we can't format only the range as we don't know if that range is inside a function or any form, so clojure-lsp format the whole form, example:
```clojure
(def a 1)

(defn foo []
   (+ 1 2))

(def b 2)
```
with that code, if you try to range format only `(+ 1 2)`, clojure-lsp will format the whole `foo` function.
In your case, as `(+ (+ 1 2 3))` is not inside a form, we try to format the whole document

gdanov:
How about not formatting it after paredit change?

ericdallo:
this is not clojure-lsp responsibility, probably would need to be done on lsp-mode if possible

snoe:
we may be able to index form ranges with rewrite-clj when files open instead of when formatting. Alternatively it might be a useful function to clj-format to (format-form-at line col)

ericdallo:
I made some tests and what seems to be the issue is the need to format the whole parent form instead of the selected one.
For example, formatting:
```clojure
(a 
  (b
   |(c
     (+ 2 3))|))
```

With the | as the selected region, it will format the whole `a` macro, since formatting only the selected region would result in wrong spaces since cljfmt doesn't know the spacing before the `c` macro.

This behavior on a big function/form/test will result sometimes in a slowness on cljfmt side even if user selected a small region of code to format or in this case, using paredit, where call lsp-mode ident to format after barf/slurp operations.  So not sure how to fix that

snoe:
We could probably hack something here.

[...]

PR #795, opened 2022-03-04
Range formatting performance

mainej:
This code investigates performance in range-formatting. Historically there have been reports (#266) of poor range formatting performance, particularly in large files for users who have auto-formatting turned on and who use paredit.

I haven't been able to reproduce really terrible performance. The worst I can do is a little over half a second to re-format large chunks of long files. For example when re-formatting the first half of a 2000 line file, which is mostly repeats of the line `(= (+ 1 2) 3)`, I get the following performance data:

```
original algorithm
614ms, std-dev 14ms
7% parsing
52% finding top forms
21% formatting
```

So that's odd... Why does it spend so much time finding the top-level forms? Let's look at `parser/find-top-forms-in-range`. First it finds every node in the range—every token, every sexp, everything. Then for _each one_ it searches upwards, to find a top-level node. Then it distincts those nodes. This is ... inefficient.

It would work equally well to search upward from the first node in the range, and the last node, and then take all the intervening top-level nodes. Let's try that.

```
first-and-last top-level
358ms, std-dev 5ms
10% parsing
9% finding top forms
43% formattting
```

This is better. (I think it also fixes some off-by-one bugs where the full range isn't formatted, but that's tangential.) Can we improve formatting time too? The original algorithm loops over the top level forms that cover the range, calling `cljfmt` for each one. But, perhaps it'd be faster to pass a single node that contains the top-level forms to `cljfmt`, so we only have to call it once.

```
first-and-last top-level, with merged formatting
301ms, std-dev 9ms
14% parsing
10% finding top forms
42% formatting
```

Not a huge improvement, but still better. And one last attempt... What if we use the more efficient cursor navigation from the `economy-of-motion` branch, which this branch is built on, to find the first and last top-level forms?

```
efficient first-and-last top-level, with merged formatting
271ms, std-dev 7ms
15% parsing
0.6% finding top forms
48% formatting
```

All together, that's about 50% faster, just by improving the `clojure-lsp` side. I still wouldn't call 271ms _fast_. I imagine it would be too slow for paredit users. But at least now most of the time is spent outside of `clojure-lsp`—in `rewrite-clj` or `cljfmt`. I think that's a win.

I'll clean this up and turn it into a real PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - General
   - Fix cljfmt settings merge during refresh/classpath configs merge to avoid multiple config vectors on same symbol.
   - Fix install script for aarch64. #794
+  - parser: more efficiently seek to cursor position, improving performance especially in large files. #793 @mainej
 
 - Editor
   - extract-function: Fix wrong args when extracting from multi-arity fn. #683
@@ -14,6 +15,7 @@
   - Improve and add lots of new snippets following practicalli config. #797
   - Improve how watched new files are analyzed avoiding infinite loops and performance issues. #796
   - Fix "incoming call hierarchy" not considering usages inside defmethods. #808
+  - range-formatting: more efficiently locate extent of range and reduce number of calls to cljfmt, improving performance especially when formatting large ranges. #795 @mainej
 
 ## 2022.02.23-12.12.12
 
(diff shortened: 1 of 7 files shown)


mainej:
Github "helpfully" converts these interactive SVGs into static SVGs. I can provide the original interactive files if requested.

01-range-formatting-orig
02-range-formatting-first-last
03-range-formatting-first-last-merge-fmt
04-range-formatting-to-loc-merge-fmt

ericdallo:
Nice! the flamegraphs seems to help a lot on those performance overviews, I didn't think we could improve that much on clojure-lsp side, thank you!

ericdallo (approved):
Looks great!

PR #805, opened 2022-03-07
Modular server

### LINKED ISSUE #807: Modular LSP lib
I'd like to be able to use the LSP bit of clojure-lsp as a library to build my own lsp server.
Currently, the cli part has a hard dependency on the lib part. This dependency would have to be pulled apart so that it's possible to depend on the lsp server and provide your own handlers.

After trying to pull it a part, I realize we probably need 4 parts:
- common: has shared utility bits and protocols
- lib: mostly stays the same as it is. implements all the clojure features of the lsp
- lsp lib: provides a layer on top of lsp4j and a coercer that handles translating lsp4j objects <-> clojure data
- cli/server: is "just" the entry point that brings it all together. provides the cli and lsp server main functions

@ericdallo what do you think about the split?

Cyrik:

- [] I added a new entry to [CHANGELOG.md](https://github.com/clojure-lsp/clojure-lsp/blob/master/CHANGELOG.md)

Closes #807 

The idea is to remove any references from cli -> lib, so that the cli part can be used as a lib. This way other LSPs can be written in Clojure without reimplementing the coercer / lsp4j interop.


ericdallo:
I hardly think we could use the cli part for other LSPs, there is probably a lot of things clojure-lsp cli does that other LSPs would like to customize, specially the capabilities supported

cli/src/clojure_lsp/server.clj
```diff
@@ -181,6 +183,37 @@
...
+    (go
+      (when (some-> @db/db :client-capabilities :experimental j/from-java :testTree)
+        (shared/logging-time
+          "Refreshing testTree took %s secs"
+          (doseq [uri uris]
+            (when-let [test-tree (f.test-tree/tree uri db)]
+              (->> test-tree
+                   (coercer/conform-or-log ::coercer/publish-test-tree-params)
+                   (.publishTestTree client)))))))))
+
```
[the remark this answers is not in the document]
ericdallo: I'm ok with that, but what's the difference between IClojureProducer and IProducer?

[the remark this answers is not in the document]
Cyrik: IProducer has methods that can have a generic implementation in lsp4clj, while IClojureProducer holds Clojure specific methods.
I'll have to do the same thing with clojure-lsp.handler/IHandler.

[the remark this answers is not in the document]
ericdallo: Got it, so the idea is that one could use lsp4clj for other servers than clojure ones right?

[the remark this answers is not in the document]
Cyrik: Yeah, like borkdudes zen or in my case an internal language at work.
Right now it would be kind of a wrapper around lsp4j, but maybe we could throw out lsp4j at some point.
If we're crazy enough it might even become a cljc project so that it can run in the browser, but that's a ways off 😄

cli/src/clojure_lsp/clojure_coercer.clj
```diff
@@ -0,0 +1,39 @@
+(ns clojure-lsp.clojure-coercer
+  (:require
+   [clojure-lsp.coercer :as coercer]
+   [clojure.spec.alpha :as s])
+  (:import
+   (clojure_lsp.feature.test_tree
+     TestTreeParams
+     TestTreeNode
+     TestTreeKind)))
+
```
ericdallo, line 10: Missing a `(set! *warn-on-reflection* true)` here

lib/src/clojure_lsp/feature/diagnostics.clj
```diff
@@ -143,9 +144,15 @@
         (when (not= :unknown (shared/uri->file-type uri))
           (sync-lint-file! uri db))))))
 
+(defn pmap-light
+  "Call claypoole pmap with less threads than pmap to avoid topping cpu."
+  [f coll]
+  (let [threadpool-size (int (Math/ceil (/ (.. Runtime getRuntime availableProcessors) 3)))]
+    (cp/upmap threadpool-size f coll)))
```
ericdallo, line 151: looks better having this function here indeed.
```suggestion
(defn ^:private pmap-light
  "Call claypoole pmap with less threads than pmap to avoid topping cpu."
  [f coll]
  (let [threadpool-size (int (Math/ceil (/ (.. Runtime getRuntime availableProcessors) 3)))]
    (cp/upmap threadpool-size f coll)))
```

lib/src/clojure_lsp/internal_api.clj
```diff
@@ -56,7 +56,10 @@
                            :type type
                            :extra extra}]
       (show-message-cli options message-content)))
-  (register-capability [_this _capability]))
+  (register-capability [_this _capability])
+
+  clojure-producer/IClojureProducer
+  (refresh-test-tree [_this _uris]))
```
ericdallo, line 62: Couldn't we don't implement `IClojureProducer` at all here since we don't have test-tree on API? 
Or would that break something, since somewhere will try to call this function from a protocol that this record doesn't implement it?

lsp4clj/.gitignore
```diff
@@ -0,0 +1,5 @@
+*.jar
+*.class
+/.cpcache
+/.clj-kondo
+/.lsp
```
ericdallo, line 5: The only LSP/clj-kondo folders that always should be included on gitignore are these
```suggestion
/.clj-kondo/.cache
/.lsp/.cache
```

lsp4clj/deps.edn
```diff
@@ -0,0 +1,10 @@
+{:deps {com.taoensso/timbre {:mvn/version "5.1.2"}
+        org.eclipse.xtend/org.eclipse.xtend.lib {:mvn/version "2.25.0"}
```
ericdallo, line 2: timbre is already in common, right?
```suggestion
{:deps {org.eclipse.xtend/org.eclipse.xtend.lib {:mvn/version "2.25.0"}
```

lsp4clj/deps.edn
```diff
@@ -0,0 +1,10 @@
...
+        org.eclipse.lsp4j/org.eclipse.lsp4j {:mvn/version "0.12.0"  :exclusions [org.eclipse.xtend/org.eclipse.xtend.lib
+                                                                                 com.google.code.gson/gson]}
+        org.clojure/data.json {:mvn/version "2.4.0"}
+        com.google.code.gson/gson {:mvn/version "2.9.0"}
```
ericdallo, line 6: Shouldn't we add `org.clojure/clojure` as well? I never know if I should add org.clojure/clojure or not on deps.edn, I think it's important to know which clojure rely when building jar/uberjar

lsp4clj/src/clojure_lsp/lsp.clj
```diff
@@ -0,0 +1,436 @@
+(ns clojure-lsp.lsp
```
ericdallo, line 1: IMO we could rename all ns inside lsp4clj to `lsp4clj`, which makes sense and helps a lot understand the boundaries between lsp4clj and the rest of clojure-lsp, WDYT?
```suggestion
(ns lsp4clj.core
```

lsp4clj/src/clojure_lsp/lsp.clj
```diff
@@ -0,0 +1,436 @@
...
+  (go-loop []
+    (<! (timeout 5000))
+    (if (shared/process-alive? ppid)
+      (recur)
+      (do
+        (log/info "Parent process" ppid "is not running - exiting server")
+        (.exit ^LanguageServer server)))))
+
+#_{:clj-kondo/ignore [:clojure-lsp/unused-public-var]}
+(deftype LSPServer [handler db capabilities client-settings]
```
ericdallo, line 294: We are probably having reflection issues here that will affect graalvm since this code is kind of generic on some parts (the macro ones), so it's a good idea to type hint `handler` here and on the other deftypes.

```suggestion
(deftype LSPServer [^IHandler handler db capabilities client-settings]
```

ericdallo (commented):
Looks good! thank you for this!

I added minor comments, main points are:

- Since we are separating most generic LSP things from clojure-lsp to lsp4clj, we could rename its namespaces to `lsp4clj` right? In the future we could even move to another repo or separated clojars lib.
- It seems a little bit confusing to me having to work with both `clojure-producer`/`producer`, `clojure-handler`/`handler`, `clojure-coercer`/`coercer`, but I know it's necessary, maybe the point above would help with that?
- I think in the future we should add tests for coercers, this should help catch bugs on that, usually integration tests test that, but not heavily.
- Maybe we should rename `lib` to `core`? I never liked `lib`, and `core` doesn't look that better but it's at least following clojure convention, but I can do that another PR.

common/src/clojure_lsp/shared_config.clj
```diff
@@ -0,0 +1,14 @@
...
+   [taoensso.timbre :as log]))
+
+(set! *warn-on-reflection* true)
+
+(def diagnostics-debounce-ms 100)
+(def change-debounce-ms 300)
+(def created-watched-files-debounce-ms 500)
+
+(defn clojure-lsp-version []
+  (string/trim (slurp (io/resource "CLOJURE_LSP_VERSION"))))
```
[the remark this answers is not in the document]
Cyrik: done and done

lib/src/clojure_lsp/internal_api.clj
Cyrik: Yeah, the problem is that the server needs to implement the function and the lib needs to call it. And we do have it on the API, `ClojureLspProducer` in server.clj implements it. I'm also not happy with the way this part is working.

The good thing is that this is all purely Clojure, with no lsp4j or java in the way. So I think we can come up with something better here.
I do like the producer design, so maybe we should just have two of those. One is provided by lsp4clj with a default implementation for built-in LSP things and a second one for custom messages. The custom one can look however you want since the implementation and the calling side is fully outside of lsp4clj. lsp4clj would just handle construction and clean up for you.
WDYT @ericdallo?

lsp4clj/deps.edn
Cyrik: Yeah, but I think it is common to still depend on the stuff your libs depend on if you use it directly, right? I think we should treat common as "just a lib", right?

lsp4clj/deps.edn
Cyrik: Yeah, I don't know either, let's be safe though 😄

lsp4clj/src/clojure_lsp/lsp.clj
Cyrik: I named it lsp4clj.lsp. I really really dislike the core naming convention. I always have 10 core.clj files open and requiring them is also a pain since you can't just `:as` the top-level namespace. I like the way @seancorfield does it in all his libs, e.g. https://github.com/seancorfield/next-jdbc/tree/develop/src/next.
So have your main entry ns (usually this would have been core) a level higher than the implementation ns's and give it the name of the lib. So I probably should change the ns to lsp4clj.lsp4clj. I'd also be happy switching to the more modern clojars format of something like `org.clojure-lsp.lsp4j` or what ever you'd want to call the org

lsp4clj/deps.edn
ericdallo: Ok, sounds good

lsp4clj/src/clojure_lsp/lsp.clj
ericdallo: the org is `com.github.io` a long name IMO, I still prefer something that indicate that this is the entrypoint of the lsp4j without repeating the name, I already used Sean suggestion on deps-bin but for this one IMO I still think  lsp4clj.core looks better

lib/src/clojure_lsp/internal_api.clj
Cyrik: I've created a new issue to do this in the next version: #818

cli/integration-test/entrypoint.clj
```diff
@@ -65,7 +65,7 @@
 
   (apply require namespaces)
 
-  (let [test-results (timeout 600000
+  (let [test-results (timeout 800000
```
ericdallo, line 68: I'm ok changing to that, but probably is not what is failing the tests but flakyness, check #704 for more details

lsp4clj/src/clojure_lsp/lsp.clj
Cyrik: done

cli/integration-test/entrypoint.clj
Cyrik: Yeah, it seems to sometimes take forever randomly on the diagnostics change integration test. I'll change it back then, no need to burn extra time if it's not going to finish anyway.

ericdallo (approved):
Thank you for the huge help!

PR #773, opened 2022-02-18
ci security step

mthbernardes:
add [clj-watson](https://github.com/clj-holmes/clj-watson) and [clj-holmes](https://github.com/clj-holmes/clj-holmes) tools to perfom security checks

PR #813, opened 2022-03-10
Match cljfmt for regex & .clj config file reading

snoe:
https://github.com/weavejester/cljfmt/issues/190
cljfmt suggests ` {:indents ^:replace {#".*" [[:inner 0]]}}` this line in their readme. Regex is not readable by `edn/read-string` so following the above issue, if the `cljfmt-config-path` ends in `.clj` use `clojure.core/read-string` instead. Otherwise, the lsp and cljfmt configs would be incompatible.

- [] I created a issue to discuss the problem I am trying to solve or there is already a open issue.


ericdallo (approved):
Thank you!

ericdallo:
Could you fix conflicts so we can merge @snoe ?

mainej:
@snoe and @ericdallo this works, but ... is it a good idea? As the [read-string docs](https://clojuredocs.org/clojure.core/read-string) clearly state, `read-string` can execute arbitrary code—write to remote URLs, etc. `(read-string "#=(prn \"pwned\")")` This is how CVEs happen. The caveat is that you can use `read-string` if the string comes from a "trusted source". But do we know that nobody but the user has touched their cljfmt config file?

I don't know... maybe I'm over reacting.

I think at a very minimum, this should be wrapped in `binding [*read-eval* false]`. But even then, it still scares me.

Maybe this is a complaint for cljfmt?

ericdallo:
Yeah, good point @mainej , IMO we should try to avoid that or think in another solution, WDYT @snoe ?

snoe:
I think the `*read-eval* false` is a good idea; added that. But I think it's important to match these tools we use so if we feel it's important we should upstream the issue and come to a consensus there.

My view is that your format file probably lives along with scripts or source code you execute anyways; not coming through untrusted input.

ericdallo (approved):
Agreed, read-eval looks enough and matching the upstream seems important indeed.

ericdallo:
Talking with @mthbernardes, he suggested using edamame from borkdude to parse both edn and clj and we wouldn't t have any security issues with that :)
Related issue #773

PR #814, opened 2022-03-10
Add feature to cycle-fn-literal

### LINKED ISSUE #774: Literal lambda to fn refactor
When I've written `#(something %1)` it's "hard" to get to a "full" lambda if I need it for some reason.

#(something %1) <-> (fn [x1] (something x1)) refactor like thread-all.

mainej:
Adds a command and code action that cycle between `(fn ...)` and `#(...)`.

- [x] Resolves #774. 


lib/src/clojure_lsp/refactor/transform.clj
```diff
@@ -379,17 +382,152 @@
...
+             (fn [zloc]
+               (when-let [replacement (get replacements (z/sexpr zloc))]
+                 (z/replace zloc replacement)))))
+
+(defn ^:private convert-literal-to-fn-params [zloc]
+  ;; only function literals #(,,,)
+  (when-let [literal-node (z/find-tag zloc z/up :fn)]
+    [literal-node]))
+
+(defn can-convert-literal-to-fn? [zloc]
```
ericdallo, line 404: could this be private?

lib/test/clojure_lsp/refactor/transform_test.clj
```diff
@@ -303,6 +303,82 @@
...
+      "(fn [element] (+ 1 element))"                                 "|#(+ 1 %)"
+      "(fn [element1] (+ 1 element1))"                               "|#(+ 1 %1)"
+      "(fn [element1 element2] (+ 1 element1 element2))"             "|#(+ 1 %1 %2)"
+      ;; mixed numbering styles
+      "(fn [element] (+ 1 element element))"                         "|#(+ 1 % %1)"
+      "(fn [element element2] (+ 1 element element2))"               "|#(+ 1 % %2)"
+      ;; vararg
+      "(fn [& args] (+ 1 args))"                                     "|#(+ 1 %&)"
+      "(fn [element & args] (+ 1 element args))"                     "|#(+ 1 % %&)"
+      "(fn [element1 element2 & args] (+ 1 element1 element2 args))" "|#(+ 1 %1 %2 %&)"
```
ericdallo, line 324: should we have a test for fn with restructuring? we will probably not be able to cycle for those cases right? maybe just don't return the code action for those cases? WDYT?

lib/src/clojure_lsp/refactor/transform.clj
mainej: Yes, good idea... Initially I was thinking `convert-fn-to-literal` and `convert-literal-to-fn` would both be public commands, and would need their own public `can-*?` helpers. But I think it's better the way it is—one public function that does what you want. I'll make these private.

lib/test/clojure_lsp/refactor/transform_test.clj
mainej: > not be able to cycle for those cases

Correct. I think there's already a test for that. See `can-cycle-fn-literal-test`.

ericdallo (approved):
Tested manually and everything looks good! Thank you!

Issue #796, opened 2022-03-04
Infinite analyze loop after changing lots of files outside editor


ericdallo:
@borkdude reported with more details [here](https://clojurians.slack.com/archives/CPABC1H61/p1646227041285309), it seems clojure-lsp gets stuck on a inifinte loop scanning the file changes watched, causing high CPU easily.

We should try to make a repro, probably 2 different branches with lots of different files should cause this

ericdallo:
For now, restarting with `lsp-workspace-restart` should reset this state

ericdallo:
@borkdude I couldn't repro the infinite loop exactly even on clj-kondo repro changing branches that have multiple changed files, but I improved the code a lot to avoid possible infinite loops/performance issues, before we were calling clj-kondo to analyze each new file, now we call one time passing all new files.

I'm not sure this will solve your issue though, could you test the latest master? you can find a nightly build on #clojure-lsp-builds

borkdude:
It happened to me several times in the last few days on different projects (clj-kondo, SCI). Every time by a git change (reset or branch switch).

borkdude:
Will test.

ericdallo:
@borkdude mentioned he still faced the issue on git branch switch, it seems not related with didChangeWachedFiles at all, but related with some looping where we analyze with clj-kondo, which could be:
- didOpen: when opening a new file
- notifyReferences: when we change a function definition, we analyze the references of it to update analysis + references count + project linting

mainej:
Here's the research I've been working on today. Sorry for taking so long to put it together. As I've been writing I've been testing things and my understanding keeps evolving. Here's where I am...

You can get into a bad state in [`file-management.analyze-changes`](https://github.com/clojure-lsp/clojure-lsp/blob/master/lib/src/clojure_lsp/feature/file_management.clj#L141). Here's a simplification of that function, with some numbers added to make it easier to talk about.

```clojure
(defn analyze-changes [text db]
  (loop [old-db @db] ;; 1
    (let [kondo-result (lsp.kondo/run-kondo-on-text! text db) ;; 2
          new-db (merge-kondo-result old-db kondo-result)] ;; 3
      (if (compare-and-set! db old-db new-db) ;; 4
        (follow-up) ;; 5
        (recur @db) ;; 6
        ))))

```

[...]

PR #821, opened 2022-03-13
Try to avoid infinite loop when analyzing several files

mainej:
This is another attempt at fixing #796. The basic idea is that this avoids a situation where when two files are updated simultaneously, the first one finishes and sets :processing-changes to false before the second one is actually done processing changes. This creates conditions where it is more likely that the app will get into an infinite loop. See #796 for details about what those conditions are.

### DIFF
--- a/lib/src/clojure_lsp/handlers.clj
+++ b/lib/src/clojure_lsp/handlers.clj
@@ -40,7 +40,7 @@
      (loop [backoff# 1]
        (if (> (quot (- (System/nanoTime) ~'_time) 1000000) 60000) ; one minute timeout
          (log/warn "Timeout waiting for changes for body")
-         (if (:processing-changes @db/db)
+         (if (seq (:processing-changes @db/db))
            (do
              (Thread/sleep backoff#)
              (recur (min 200 (* 2 backoff#)))) ; 2^0, 2^1, ..., up to 200ms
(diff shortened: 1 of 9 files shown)


lib/src/clojure_lsp/handlers.clj
```diff
@@ -40,7 +40,7 @@
      (loop [backoff# 1]
        (if (> (quot (- (System/nanoTime) ~'_time) 1000000) 60000) ; one minute timeout
          (log/warn "Timeout waiting for changes for body")
-         (if (:processing-changes @db/db)
+         (if (seq (:processing-changes @db/db))
```
mainej, line 43: I have a commit locally where this is `(contains? (:processing-changes @db/db) ~uri)`, i.e., the handlers wait for changes to the file they are updating, instead of waiting for changes to any file. I'm not sure though... it'd be a very minor speed up because it's rare that there are changes to several files simultaneously. WDYT?

ericdallo: sounds good to me, the only issue would be if analysis from a file would affect another file, but I don't see that an issue for now.

ericdallo (approved):
Looks good, LMK if you want to commit that local commit

mainej:
Thanks @ericdallo. I added that local commit, so this is finalized.

mainej:
Oh wait, one more thing... line numbers in the macro.

mainej:
OK, fixed.

PR #835, opened 2022-03-18
add Components conecpt

### LINKED ISSUE #831: Remove timbre logger and its dependencies from lib
We have the timbre dependency really coupled to almost all namespaces, timbre has some dependencies and it would be nice to remove it from `lib` to provide a API with less dependencies.

ericdallo:
Fixes #831


ericdallo:
Integration tests are failling because of timeout but the reason is that is happening a StackOverflow when logging with logger, need to debug that later

snoe:
I suggested slf4j elsewhere, and looking at this pr reinforces my belief it would better handle the issue of a library/app split. Is it not graal-able? I think it'd be worth a look before merging this.

Looking at this changeset, and we can see the `logger` needs to be passed through functions to get deep into the call hierarchies where it's needed. If another function needs to log something but doesn't have it passed through another large changeset would need to be created updating five or six functions to get the logger for the single log line. Cross cutting concerns like this are much better handled by a global singleton for this reason. 

Logging is also a dev tool, what if you want to quickly `spy` a value and `logger` isn't around? `println` doesn't work since writing to stdout breaks clients.

ericdallo:
Yeah, I was thinking the same for logger, but was not sure would be a good idea, but maybe makes sense indeed to have a reference to a atom just for logger.

The issue with slf4j is that is not Graal compatible

ericdallo:
@snoe good news, I followed your suggestion and it seems better indeed, we now have a dynamic var for logger in a way it doesn't seem that bad, and we can still remove timbre deps from `lib` and have it only for `cli` :)

snoe:
Yup, much better. 

It looks like there's no `spy` in the protocol, it's an important part of my workflow. Does it make sense to require it in implementations (timbre and clojure.tools.logging have it but java libs don't) or do we add our own in the `protocols` ns that delegates to `debug`? 

Something like this would do (untested)
```clojure
(defmacro [expr]
   `(let [result# ~expr] (debug (str '~expr " => " result#)))
```

ericdallo:
@snoe for debugging tests you can require timbre manually and use spy normally, we can't just push timbre in lib since it won't compile anymore

snoe:
I see, still haven't my head around where all the seams between pieces are now.

snoe (approved):

PR #836, opened 2022-03-18
Fixes #830 - change find-let-form to consider locals within zloc

### LINKED ISSUE #830: Move to let doesn't handle other binding forms like if-let/doseq/for/etc..
**Describe the bug**

Move to let doesn't consider other functions with binding forms 

**To Reproduce**
With the cursor at the `|`

```clojure
(let [x 1]
  (doseq [n nums]
    |(+ x n)))
```
this transforms to the invalid

```clojure
(let [x 1
      m (+ x n)]
  (doseq [n nums]
    m))
```

I believe anything other than `let` that introduces bindings will run into this.

**Expected behavior**
instead, it should noop since the move is illegal.

after https://github.com/clojure-lsp/clojure-lsp/issues/829 it should introduce a `:let [m (+ x n)]` in the doseq.
I think the order of these two issues can be tackled independent but they should consider each other.

```
clojure-lsp 2022.02.23-12.12.12
clj-kondo 2022.02.09
```

snoe:

### DIFF
--- a/lib/src/clojure_lsp/refactor/transform.clj
+++ b/lib/src/clojure_lsp/refactor/transform.clj
@@ -37,6 +37,19 @@
                 (dissoc :loc))))
         zip-edits))
 
+(defn- unify-to-one-language
+  "Drop the clojurescript analysis info when there is also clojure info"
+  [analysis]
+  (let [langs-present (->> analysis
+                           vals
+                           flatten
+                           (map :lang)
+                           (into #{}))
+        non-clj-lang? (fn [{:keys [lang]}] (and lang (not (= :clj lang))))]
+    (if (= #{:clj :cljs} langs-present)
+      (medley/map-vals (partial remove non-clj-lang?) analysis)
+      analysis)))
+
 (defn- coll-tag [zloc]
   (get #{:vector :set :list :map} (z/tag zloc)))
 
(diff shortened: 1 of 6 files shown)


lib/src/clojure_lsp/refactor/transform.clj
```diff
@@ -37,6 +37,19 @@
                 (dissoc :loc))))
         zip-edits))
 
+(defn- unify-to-one-language
```
ericdallo, line 40: Should this go to queries.clj or do you think it's a overhead?

lib/src/clojure_lsp/refactor/transform.clj
```diff
@@ -228,16 +241,40 @@
...
+  "Finds a let-form that would be valid to move zloc to"
+  [zloc uri db]
+  (let [let-loc (some-> zloc
+                        (edit/find-ops-up "let")
+                        z/up)]
+    (when let-loc
+      (let [{:keys [col row end-row end-col]} (meta (z/node zloc))
+            clj-analysis (unify-to-one-language (:analysis @db))
+            bindings-loc (z/right (z/down (zsub/subzip let-loc)))
+            local-defs (->> (q/find-local-usages-under-form
```
ericdallo, line 254: Usually this is not a problem but this function that is called all the time on code actions has increased it's complexity and it's probably slower than before, with a chance to slow code actions process as a whole, right?
Should we check performance of this new function?
Usually we just manually test both old and new code on a code snippet that should be relevant, WDYT?

lib/src/clojure_lsp/refactor/transform.clj
snoe: I just moved this function up in the ns, so I don't know. Probably outside scope tho of this tho.

lib/src/clojure_lsp/refactor/transform.clj
snoe: Good catch, unify to one language is expensive, and upon reflection I think we don't want to call it either way. We would want to invalidate the move if the local escapes in any language.

The other calls are pretty quick it seems to pop up immediately in large functions and in large files. But my client doesn't keep the actions up like I think emacs does. So I dropped the unify but I think the rest is safe since it operates on local-analysis.

ericdallo: looks good, I was afraid regarding `q/find-local-usages-under-form` and especially `q/find-definition` since it maps over all var-definitions.

BTW emacs calculate the code actions each time user change a line but with smart caches.

snoe: yup, fyi since the only bucket going into `find-definition` will be locals, the multi-method will only look at the file's local analysis.

PR #838, opened 2022-03-19
Fixes #837

### LINKED ISSUE #837: Add missing changes

There are two places where `add-missing-libspec` fails to add to the `:require`
- Namespaces that have never been aliased in the project.
- If you do not wish to alias namespace (think top level packages) but still require it.

As well, including a namespace that you just created is cumbersome, if you could use completion to find the new ns, and choosing it requires it it would be much better.

`add-missing-libspec` right now asks "is there an alias?"
If we split it to ask "is this a namespace?" and "what alias should I use?" then both these issues would be handled.

If the symbol under cursor's namespace is either an alias or namespace, then it should be added to the require. If it is an alias it should be added to the require.

This could also introduce a client-setting like `:auto-alias-add-missing?` that if a namespace is found instead of an alias, then an alias could be added automatically by taking splitting the namespace by `.` and progressively adding the tail until a unique project alias is created. 
i.e. if `[foo.bar.models.thing :as thing]` exists, and you are adding to `foo.bar.common.thing/name` it would require `[foo.bar.common.thing :as common.thing]` and change the cursor to `common.thing/name`

This should also work with completions that add `:additional-text-edits` for adding missing from selecting the completion.

snoe:
A number of logic changes to what gets suggested, the principle is not
to introduce conflicting aliases and to reuse existing ones as much as
possible.

Combines the logic for add missing libspec and find suggestions
Removed add missing libspec from code actions because it's redundant
when a user chooses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
   - Fix handling cljfmt config files that end in `.clj` https://github.com/weavejester/cljfmt/issues/190
   - parser: more efficiently seek to cursor position, improving performance especially in large files. #793 @mainej
   - Fix clean-ns not sorting properly node requires for cljs. #815
+  - Improve logic around require suggestions. #837
 
 - Editor
   - extract-function: Fix wrong args when extracting from multi-arity fn. #683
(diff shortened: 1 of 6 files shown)


ericdallo:
@snoe I tested manually and it looks good! I hope this won't affect negatively anyone, but as we discussed probably makes sense 99% of time to follow that idea indeed.
Could you fix lint? there is a `make lint-fix`

snoe:
Thanks for testing it @ericdallo. I think now that there's an ordering of suggestions here, it's possible to add to if people want it.

PR #839, opened 2022-03-20
Avoid changing DB identity

mainej:
This is another tool to avoid the infinite loop described in https://github.com/clojure-lsp/clojure-lsp/issues/796. It avoids accidentally changing the db's identity when the linting data is `=` but not `identical?` to what was there before.

### DIFF
--- a/lib/src/clojure_lsp/kondo.clj
+++ b/lib/src/clojure_lsp/kondo.clj
@@ -152,8 +152,21 @@
               (do
                 (Thread/sleep 50)
                 (recur (inc tries)))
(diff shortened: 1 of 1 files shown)


ericdallo (approved):
Thank you!

mainej:
@ericdallo when I was testing this change, I couldn't get the infinite loop to happen at all, even if I removed the check for `:processing-changes` entirely. I think `:processing-changes` is an extra layer of security, so it's still useful. 

At this point I think the infinite loop is _very_ unlikely, or at least, unlikely from all the ways I know it can happen. You might want to consider closing #796. Of course if anyone sees the infinite loop, it should be re-opened.

Issue #840, opened 2022-03-20
Vars that are declared but never used are not marked as unused


mainej:
**Describe the bug**
I want to check... is this actually a bug or desired behavior? In the below code, I expect `foo` to get the unused-public-var diagnostic, but it doesn't:
```clojure
(declare foo) ;; should be marked as unused, but isn't
```
If the declared var is later defined, but is still unused, both should get the unused-public-var diagnostic, but neither does:
```clojure
(declare foo) ;; should be marked as unused, but isn't
(def foo 1) ;; should be marked as unused, but isn't
```

**Expected behavior**
I expect `foo` to be marked as unused in all three declarations above.

When `foo` is finally used, I expect the diagnostic to be removed from both of them
```clojure
(declare foo) ;; should be used
(def foo 1) ;; should be used
(inc foo)
```

If it's a bug, the fix is to remove [this line](https://github.com/clojure-lsp/clojure-lsp/blob/f4ee01df134988a7d8191e9de973672244d9784b/lib/src/clojure_lsp/queries.clj#L354). I did a quick test—it gives me the expected behavior and all the tests pass.

ericdallo:
It seems to be your expected result makes sense, but I can't recall that was on purpose to solve some kind of edge case.
Anyway, I'm ok fixing it since all tests pass adding a new test and waiting to see if anyone complains

Issue #841, opened 2022-03-20
defmultis are never marked as unused


mainej:
**Describe the bug**
This is similar to #840... In the below code, I expect `foo` to get the unused-public-var diagnostic, but it doesn't:
```clojure
(defmulti foo :bar) ;; should be marked as unused, but isn't
(defmethod foo :baz [_] 1)
```

**Expected behavior**
I expect `foo` to be marked as unused above.

When `foo` is used, I expect the diagnostic to be removed
```clojure
(defmulti foo :bar) ;; should be used
(defmethod foo :baz [_] 1)
(foo {:bar :baz})
```

If it's a bug, I have a fix.

ericdallo:
Thanks for the report, the issue is that a defmethod is a var-usage of a defmulti I think, making it a usage.

mainej:
> defmethod is a var-usage

That's exactly right, that's why it isn't marked as unused. So, my proposal is to make `find-references` exclude defmulti when `include-declaration?` is false. With that change, and with the change in #840, the implementations of `find-references` for var-usages and var-definitions will match, which I think makes sense.

I have a patch that will fix both issues at once, so I'll submit that.

ericdallo:
not sure it's a good idea exclude defmulti from find-references, most of the time references should include both implementations and usages, but I guess this won't be a problem since users could call find-implementation now for those cases

mainej:
Yeah, I'd expect to use find-implementation for that case. Here's how things work currently, and my proposed changes. What do you think? Are there other cases to consider?

| highlighting | from defmulti | from defmethod | from usage |
|--------------|---------------|----------------|------------|
| to defmulti  | yes           | yes            | yes        |
| to defmethod | yes           | yes            | yes        |
| to usage     | yes           | yes            | yes        |

| unused public var                   | warning on defmulti   | warning on defmethod |
|-------------------------------------|-----------------------|----------------------|
| only defmulti defined               | yes                   | n/a                  |
| only defmulti and defmethod defined | no **(proposed yes)** | n/a                  |
| at least one other usage            | no                    | n/a                  |

| code lens                           | on defmulti                   | on defmethod |
|-------------------------------------|-------------------------------|--------------|
| only defmulti defined               | 0 references                  | n/a          |
| only defmulti and defmethod defined | 1 reference **(proposed 0)**  | n/a          |
| at least one other usage            | 2 references **(proposed 1)** | n/a          |


[...]

Issue #808, opened 2022-03-08
lsp-treemacs-show-call-hierarchy does not report call sites within defmethod bodies


lassemaatta:
**Describe the bug**

`lsp-treemacs-show-call-hierarchy` does not report call sites which are inside `defmethod` bodies.

**To Reproduce**

```clojure
(defn foobar
  [])
​
(defn zed
  []
  (foobar))  ;; shows up in call hierarchy
​
(defmulti my-multi :some-key)
​
(defmethod my-multi :some-value
  [_]
  (foobar)) ;; does not show up in hierarchy
```

**Expected behavior**

`lsp-treemacs-show-call-hierarchy` should show all call sites.

**Screenshots**

![Screenshot_20220308_150915](https://user-images.githubusercontent.com/18479268/157245198-93b07689-b745-446f-99df-ed04c38a6946.png)

**User details (please complete the following information):**
 - OS: ArchLinux
 - Editor: emacs
 - Version: `clojure-lsp 2022.02.23-12.12.12`

**Additional context**

- [Slack discussion](https://clojurians.slack.com/archives/CPABC1H61/p1646745006768989)

ericdallo:
I just realized this issue depends on https://github.com/clj-kondo/clj-kondo/issues/1570 :(

ericdallo:
My PR was merged so I can work on that fix !

ericdallo:
Fixed on master @lassemaatta! you can try the nightly release on #clojure-lsp-builds on Clojurians slack

lassemaatta:
Great, thanks! I tested with the latest nightly and the issue is now fixed :+1:

PR #843, opened 2022-03-21
Exclude declare when calculating unused-public-vars

mainej:
Edit: this used to address #840 and #841. Now it only addresses #840. 

This resolves https://github.com/clojure-lsp/clojure-lsp/issues/840 ~and https://github.com/clojure-lsp/clojure-lsp/issues/841.~

Before this, `declare` and `def/defn` counted in each other's usages, and so they weren't marked as unused-public-vars, even if they had no real usages. ~Similarly, `defmethod` counted for `defmulti`'s usages, so a `defmulti` with at least one `defmethod` was never marked as an unused-public-var, even if it didn't have any real usages.~

~This also changes the code lens reference counts.~

~In the case of defmulti, it means that defmethods are no longer included when invoking find-references (a.k.a. go to references). You can still get from a defmulti to its defmethods with find-implementations. We may add defmethods back to find-references at some point, at the same time as we add protocol method implementations.~

### DIFF
--- a/lib/src/clojure_lsp/queries.clj
+++ b/lib/src/clojure_lsp/queries.clj
@@ -340,8 +340,7 @@
                             ;; usage from own definition
                             (and (:from-var %)
                                  (= (:from-var %) (:name element))
-                                 (= (:from %) (:ns element)))
-                            (:defmethod %))))
+                                 (= (:from %) (:ns element))))))
             (medley/distinct-by (juxt :filename :name :row :col)))
           analysis)))
 
--- a/lib/src/clojure_lsp/queries.clj
+++ b/lib/src/clojure_lsp/queries.clj
@@ -52,13 +52,13 @@
                             (shared/uri->available-langs (:filename check-element)))]
     (seq (set/intersection match-file-lang check-file-lang))))
 
-(defn ^:private defrecord-names-for [{:keys [name]}]
-  #{name
-    (symbol (str "->" name))
-    (symbol (str "map->" name))})
-
-(defn ^:private deftype-names-for [{:keys [name]}]
-  #{name (symbol (str "->" name))})
+(defn ^:private var-definition-names [{:keys [defined-by name]}]
+  (case defined-by
+    clojure.core/defrecord
+    , #{name (symbol (str "->" name)) (symbol (str "map->" name))}
+    clojure.core/deftype
+    , #{name (symbol (str "->" name))}
+    , #{name}))
 
 (defn find-local-usages-under-form
   [analysis filename line column end-line end-column]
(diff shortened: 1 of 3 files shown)


lib/src/clojure_lsp/queries.clj
```diff
@@ -316,45 +316,32 @@
...
-                         (and (or (not (:from-var %))
-                                  (not= (:from-var %) (:name element))
-                                  (not= (:from %) (:to element)))
-                              (not (identical? :var-definitions (:bucket %)))
-                              (not (:defmethod %)))))
-            (medley/distinct-by (juxt :filename :name :row :col)))
-          analysis)))
+    (let [var-definition {:ns (:to element)
+                          :name (:name element)
+                          :bucket :var-definitions}]
```
ericdallo, line 324: Kind of dangerous, if we start to check other var-definitions fields we need to remember to update here, for example, this is missing the defined-by which is used there

lib/src/clojure_lsp/queries.clj
```diff
@@ -316,45 +316,32 @@
...
+        exclude-declaration? (not include-declaration?)]
     (into []
           (comp
             (mapcat val)
-            (filter #(not (identical? :keywords (:bucket %))))
-            (filter #(or (safe-equal? (:name element) (:name %))
-                         (and (or defrecord? deftype?)
-                              (contains? names (:name %)))))
+            (remove #(identical? :keywords (:bucket %)))
+            (filter #(contains? names (:name %)))
```
ericdallo, line 335: way easier indeed

lib/src/clojure_lsp/queries.clj
```diff
@@ -316,45 +316,32 @@
...
             (filter #(safe-equal? (:ns element) (or (:ns %) (:to %))))
-            (filter #(or include-declaration?
-                         (and (or (not (identical? :var-definitions (:bucket %)))
-                                  (= (:defined-by %) 'clojure.core/declare))
-                              (or (not (:from-var %))
-                                  (not= (:from-var %) (:name element))
-                                  (not= (:from %) (:ns element))))))
+            (remove #(and exclude-declaration?
```
ericdallo, line 337: Shouldn't this be a `or`?
```suggestion
            (remove #(or exclude-declaration?
```

lib/src/clojure_lsp/queries.clj
ericdallo: Yeah, agreed, I could only think on calling find-definition which I don't want to add right now for that until we confirm performance is not a issue there anymore, So I'm ok with that for now

lib/src/clojure_lsp/queries.clj
ericdallo: I don't think so it would be clearer, sounds the same to me 🤷🏻‍♂️  but I'm ok too, I think this was `exclude-declaration` before I try to make it clearer with `include-declaration` but we can see no matter how, it's hard indeed 😂

ericdallo (approved):
Looks good to me

ericdallo:
Thank you!

mainej:
@ericdallo I apologize, but this may have impacted something you worked on recently, #808. It's not exactly broken—you still see the defmethod, as requested in that bug report—but it does something I wouldn't expect:

Screen Shot 2022-03-22 at 6 51 20 PM

Because the defmethod is now included in its own references, you end up with cyclic references in the call hierarchy.

I'm not sure how to fix this. Maybe prevent cycles [here](https://github.com/clojure-lsp/clojure-lsp/blob/9b759b345bc4c26088dd21bd8bbb4c5477501e8a/lib/src/clojure_lsp/feature/call_hierarchy.clj#L69)? Would that break the call hierarchy for true recursive calls?

ericdallo:
@mainej the call hierarchy code is generic and I'd like to keep this way, this probably is showing that we are not considering `include-declaratoon?` properly or something like this I guess

ericdallo:
@mainej I took a look and the reason is that when calling `find-references` on a defmehotd, it returns itself, not sure if that makes sense.
here's a code example:
```code
(defn foobar
  [])

(defn zed
  []
  (foobar))  ;; shows up in call hierarchy

(defmulti my-multi :some-key)

(defmethod my-multi :some-value
  [_]
  (foobar)) ;; does not show up in hierarchy

(defn foo []
  (my-multi :some-value))
```

image

It'd be nice to fix this behavior, although not a huge bug IMO

mainej:
> when calling find-references on a defmethod, it returns itself

That's my interpretation too. That's a change that was introduced in #843.

> It'd be nice to fix this behavior, although not a huge bug IMO

I agree, it makes the call hierarchy a little confusing, but it's not that painful. If we include implementations in references, as we've been talking about, this behavior will start showing up in more cases. But I think we can ignore it for awhile.

PR #846, opened 2022-03-21
Fixes #829 - Move to let improvements

### LINKED ISSUE #829: Move to let should create a let at the highest reasonable level if missing
Calling move to let without a `let` present in the sexp tree does a noop, so I always wrap the body of my fn in a let [] and the call move to let on my sexp.

If move to let is called and an existing `let` is not found, a new `let` should be added at the highest "reasonable" scope. Make a best guess for the user where the let should go.

## What's reasonable?
Here's my thoughts but this probably needs some testing for what feels natural or if differs to other's ideas of idiomatic clojure, and possibly introduce settings for some of these choices or differences.

- if the sexp being moved contains a local then the let must be added below the definition of that local.
- if the local is defined in a `let`, it should move into the same one, if it is defined by a `for` or `doseq` a `:let [... ...]` should be added.
- if extracting through an anonymous function only escape the function if the locals are all closures, otherwise wrap the fn body.
- if you are extracting through a branch (`cond/if/when/etc`) we should wrap the let around the branch body
- otherwise the `let` should wrap the top level `let/defn/def/comment` body

This essentially combines introduce let/expand let/move to let but in my experience that three step is far more cumbersome to do than manually putting the let where you want it and doing it yourself, especially since you may be deep in the tree and want the let up top.
However I do think this is different than introduce-let/expand-let as those are less about putting a let in an idiomatic position than moving code around.

In cursive and intellij the extract variable operation asks the user where to put the let after calling the command, I'm unsure if that's possible for us to introduce the question, but if it was that might be preferable to these rules (with options still being legal moves).

snoe:
Move to let will now fallback to introducing a let and expanding it if an existing one can't be found

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
   - Fix clean-ns not sorting properly node requires for cljs. #815
   - Fix move-to-let to ensure locals don't move out of scope. #830
   - Improve logic around require suggestions. #837
+  - Enhance move-to-let to introduce and expand let if an existing one doesn't exist. #829
 
 - Editor
   - extract-function: Fix wrong args when extracting from multi-arity fn. #683
--- a/cli/integration-test/integration/code_action_test.clj
+++ b/cli/integration-test/integration/code_action_test.clj
@@ -20,7 +20,9 @@
 
   (testing "When a code action is available"
     (h/assert-submaps
-      [{:title "Cycle privacy"
+      [{:title "Move to let"
+        :kind "refactor.extract"}
+       {:title "Cycle privacy"
         :kind "refactor.rewrite"}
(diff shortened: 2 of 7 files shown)


ericdallo:
Could you fix lint please @snoe ?

ericdallo:
@snoe I'm getting a NPE when trying to move for example:
lib/src/clojure_lsp/kondo.clj:219 to the let, any thoughts?

ericdallo:
I just realized recent logger refactor messed up Exceptions string on the logger, logging just the exception name, I'll try to fix it separately

ericdallo (approved):
Looks good!
I tested and it's working properly, thank you!

PR #861, opened 2022-03-23
Optimize unused public var

mainej:
The baseline algorithm for finding unused public vars has the following performance characteristics.

|  clojure-lsp             |   time | std-dev |
|---------------|-------:|--------:|
| queries.clj (large file)    |   `98` ms |     `1` ms |
| coercer.clj (kw-heavy file) |  `249` ms |    `11` ms |
| whole project | `1220` ms |    `11` ms |

This commit changes algorithms to improve performance:

|  clojure-lsp             | time | std-dev |
|---------------|-----:|--------:|
| queries.clj (large file)    | `13` ms |  `0.20` ms |
| coercer.clj (kw-heavy file) | `13` ms |  `0.20` ms |
| whole project | `24` ms |  `0.16` ms |

Larger projects see a more pronounced improvement:

| clj-kondo                 |  baseline |      new |
|---------------------------|----------:|--------:|
| analyzer.clj (large file) |  `294` ms | `18` ms |
| whole project             | `2330` ms | `32` ms |

| clojure.core          |  baseline |      new |
|-----------------------|----------:|--------:|
| core.clj (large file) | `3270` ms | `44` ms |
| whole project         | `6570` ms | `63` ms |

Impressively, and unlike the baseline algorithm, this performance is achieved without any parallelization.

To compare the algorithms, first consider the baseline algorithm. First it loops through the project analysis once to find var definitions and once to find keyword definitions. Then, for each definition, it loops through the analysis once again, looking for any usages. This has quadratic performance, based on the size of the project analysis (# of definitions * # of usages).

The new algorithm is linear in the size of the project analysis (# of definitions + # of usages).  Instead of looking for usages of one definition at a time, first it looks for all usages. It loops once through the project analysis to find all var usages, and once for all keyword usages. Then it loops once for all var definitions, using set logic to remove any that have usages, and another time for keyword definitions.

This is the biggest optimization, converting an O(N*N) algorithm to O(N). There are also some smaller optimizations to constrain the number of var usages that are inspected, reducing the size of N. This is particularly helpful during single-file analysis, when we can ignore usages of other namespaces.

Since whole-project linting is not much slower than single-file linting, perhaps it would be simpler to re-lint the full project after every file change. Then we wouldn't need to re-lint the files that reference the first file. This work is not done in this PR, but should be researched afterward.

See https://github.com/mainej/clojure-lsp/pull/1 for more discussion.

- [ ] I created a issue to discuss the problem I am trying to solve or there is already a open issue.

### DIFF
--- a/lib/src/clojure_lsp/feature/diagnostics.clj
+++ b/lib/src/clojure_lsp/feature/diagnostics.clj
@@ -6,7 +6,6 @@
    [clojure-lsp.shared :as shared]
    [clojure.core.async :as async]
    [clojure.java.io :as io]
-   [com.climate.claypoole :as cp]
    [lsp4clj.protocols.logger :as logger]
    [medley.core :as medley]))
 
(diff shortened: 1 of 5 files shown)


lib/src/clojure_lsp/kondo.clj
```diff
@@ -121,14 +121,10 @@
       (if (:api? @db)
         (do
           (logger/info (format "Starting to lint whole project files..."))
-          (shared/logging-time
-            "Linting whole project files took %s secs"
-            (f.diagnostic/lint-project-diagnostics! new-analysis kondo-ctx db)))
+          (f.diagnostic/lint-project-diagnostics! new-analysis kondo-ctx db))
         (when (settings/get db [:lint-project-files-after-startup?] true)
           (async/go
```
ericdallo, line 126: which the huge performance improvement, I'll consider remove the async of this function, making sync for both cli and editor would result in a way easier code to understand and probably fix some concurrencies issues

lib/src/clojure_lsp/feature/diagnostics.clj
```diff
@@ -144,57 +142,73 @@
...
-(defn ^:private file-definitions [project-analysis filename]
-  (concat (q/find-var-definitions project-analysis filename false)
-          (q/find-keyword-definitions project-analysis filename)))
+  [var-defs kw-defs project-analysis {:keys [config reg-finding!]}]
+  (let [var-definitions (remove (partial exclude-public-diagnostic-definition? config) var-defs)
+        var-nses (set (map :ns var-definitions)) ;; optimization to limit usages to internal namespaces, or in the case of a single file, to its namespaces
+        var-usage-signature (juxt :to :name)
+        var-usages (into #{}
+                         (comp
+                           (q/xf-all-var-usages-to-namespaces var-nses)
```
ericdallo, line 152: I wonder if just get all var-usages is faster than filter for those nses, usually all namespaces have var-definitions right? so why do filter those `var-nses` which probably we will get all var-usages in the end?

ericdallo:
I tested on a Nubank huge project (> 180K loc) and takes 3s where before it took at least 5-10 minutes...

I'll merge this as a huge addition to clojure-lsp performance and if you think you wanna change anything related to my comment of var-usages feel free to open another PR.
This looks fantastic... thank you very much for that!

lib/src/clojure_lsp/feature/diagnostics.clj
mainej: > probably we will get all var-usages in the end?

No, we get many fewer var usages. Remember that though the project analysis has been filtered to internal files, those files use both internal vars _and_ external vars, vars defined in clojure.core and in libraries. Usages of external vars are the majority, so, it's valuable to filter them out.

I tried it both ways as I was developing it, and it's about 25% faster to filter to internal vars, for both single files and for the whole project.

I'd like to document these things, but optimizations are hard to describe and the comments often get out of date as other code changes. Let me know if you'd like to see specific things documented.

ericdallo: oh, yes, I agree, but couldn't we do `(->> analysis q/filter-project-only :var-usages)`, we would get all var-usages from project-only too right?

mainej: > couldn't we do `(->> analysis q/filter-project-only :var-usages)`

First, let's make sure we're on the same page. I think this is what you meant:

```clojure
(->> analysis                               ;; 1
     q/filter-project-analysis              ;; 2
     (mapcat val)                           ;; 3
     (filter #(= :var-usages (:bucket %)))) ;; 4
```

Is that right? You wrote `q/filter-project-only`. Is that something I don't know about, or did you mean `q/filter-project-analysis`?

Again, the answer is no, that's not the same. Let's go line-by-line. Suppose we start with this `analysis`, line 1:

```clojure
{"clojure_lsp/shared.clj"  '[{:bucket :var-definitions
                              :ns clojure-lsp.shared
                              :name inside?}
                             {:bucket :var-usages  ;; <<< we don't care about this var-usage
                              :to clojure.core
                              :name inc}]
 "clojure_lsp/queries.clj" '[{:bucket :var-definitions
                              :ns clojure-lsp.queries
                              :name find-local-usages-under-form}
                             {:bucket :var-usages
                              :to clojure-lsp.shared
                              :name inside?}]
 "clojure/core.clj"        '[{:bucket :var-definitions
                              :ns clojure.core
                              :name inc}]}
```
Then `q/filter-project-analysis`, line 2:
```clojure
{"clojure_lsp/shared.clj"  '[{:bucket :var-definitions
                              :ns clojure-lsp.shared
                              :name inside?}
                             {:bucket :var-usages  ;; even though we've filtered to project analysis, we still have this var-usage
                              :to clojure.core
                              :name inc}]
 "clojure_lsp/queries.clj" '[{:bucket :var-definitions
                              :ns clojure-lsp.queries
                              :name find-local-usages-under-form}
                             {:bucket :var-usages
                              :to clojure-lsp.shared
                              :name inside?}]}
```
"clojure/core.clj" has been removed, but we still have **usages** of clojure.core. Then `(mapcat val)`, line 3:
```clojure
'[{:bucket :var-definitions
   :ns clojure-lsp.shared
   :name inside?}
  {:bucket :var-usages
   :to clojure.core
   :name inc}
  {:bucket :var-definitions
   :ns clojure-lsp.queries
   :name find-local-usages-under-form}
  {:bucket :var-usages
   :to clojure-lsp.shared
   :name inside?}]
```
Then `(filter #(= :var-usages (:bucket %)))`, line 4:
```clojure
'[{:bucket :var-usages ;; still here
   :to clojure.core
   :name inc}
  {:bucket :var-usages
   :to clojure-lsp.shared
   :name inside?}]
```
At this point we know that `clojure.core/inc` and `clojure-lsp.shared/inside?` are both used. But it's not important that `clojure.core/inc` is used! My code filters **usages to** internal namespaces:
```clojure
'[{:bucket :var-usages
   :to clojure-lsp.shared
   :name inside?}]
```

This list is smaller, and makes the rest of the code faster.

ericdallo: Got it, makes total sense, thanks for the detailed explanation, even having var-usages from project they could be usages of external definitions which we don't care, I can see that now

PR #863, opened 2022-03-23
862 don't select after move-coll-entry

snoe:


ericdallo (approved):
Looks good, I tested and it keeps working for emacs

PR #869, opened 2022-03-24
Log meaningful durations

### LINKED ISSUE #824: Request: report accurate times in logs
The logs are useful for monitoring the performance of clojure-lsp features. But currently when you change a file, the logs report several misleading times:
```
2022-03-13T17:35:01.439Z  DEBUG [lsp4clj.core:?] - :didChange 0ms
2022-03-13T17:35:01.844Z  INFO [clojure-lsp.feature.file-management:?] - changes analyzed by clj-kondo took 0.10 secs
2022-03-13T17:35:01.917Z  DEBUG [lsp4clj.core:?] - :documentHighlight 477ms
2022-03-13T17:35:01.932Z  DEBUG [lsp4clj.core:?] - :semanticTokensRange 492ms
2022-03-13T17:35:02.044Z  DEBUG [lsp4clj.core:?] - :semanticTokensFull 603ms
```
 In reality, didChange take several hundred milliseconds—it reports 0ms because it does its work asynchronously. While didChange is working documentHighlight, semanticTokensRange and semanticTokensFull all wait. The logs report their wait time plus their processing time, even though the actual work usually takes only tens of milliseconds.

I'd like the logs to report the true processing time for all of these methods. (For easy comparison, I'd also like the logs to report times in milliseconds, instead of a mix of seconds and milliseconds.)

I think the easiest way to report true processing time would be to move the logging out of lsp4clj.core and in many cases even out of clojure-lsp.server, all the way down to the handlers. In most cases whichever layer does things synchronously has the right information to log the true times.

At the moment, lsp4clj.core contains macros that facilitate logging of asynchronous things, most often CompletableFutures. The macros create closures around the async thing, capturing the time that the async process was started, and then writing a log line when its body finishes. Though this doesn't work when you have several layers of asynchronicity (consider, for example, didChange), I think we may want to keep a facility like it for other users of lsp4clj. When possible, though, I'd like clojure-lsp to log at the synchronous layer.

@Cyrik since you were part of the recent extraction of lsp4clj, I wanted to ask—will you miss the logging at that layer? Would it help to keep async logging tools in lsp4clj, even if they aren't used in lsp4clj.core?

* We could log from both lsp4clj.core and lower layers. However, this would make the logs noisier.

mainej:
This resolves #824. Before, logs looked something like the following, which under-reports didChange duration and over-reports documentHighlight.

```
2022-03-13T17:35:01.439Z  DEBUG [lsp4clj.core:?] - :didChange 0ms
2022-03-13T17:35:01.844Z  INFO [clojure-lsp.feature.file-management:?] - changes analyzed by clj-kondo took 0.10 secs
2022-03-13T17:35:01.917Z  DEBUG [lsp4clj.core:?] - :documentHighlight 477ms
```
Now the logs look like this:
```
2022-03-24T18:16:47.494Z  INFO [clojure-lsp.feature.file-management:148] - changes analyzed by clj-kondo took 93ms
2022-03-24T18:16:47.495Z  INFO [clojure-lsp.feature.file-management:144] - :analyze-file 94ms
2022-03-24T18:16:47.504Z  INFO [clojure-lsp.handlers:184] - :document-highlight 9ms - waited 399ms
```

At a lower level, these are the changes:

* For easier comparison, durations are logged in milliseconds instead of a mix of milliseconds and seconds
* For processes that wait on analysis, logs the time waiting separately from the time running
  * This will help identify processes that need (or do not need) performance improvement
* ~Decreases time waiting by blocking on analysis to be finished, instead of sleeping with a back off~ (reverted)
* Restores ns and line number to logs /cc @Cyrik see change to the logger protocol.
* Moves duration logging out of lsp4clj, down into the clojure-lsp handlers or even lower, to a level that has sufficient information to correctly log the duration of async or partially async processes. /cc @Cyrik 

With these changes, it's now obvious that wait time is essentially 300ms (debounce time) + analysis time. 300ms is a long time for a UX to respond. Perhaps we could switch to a [leading and trailing debounce](https://docs-lodash.com/v4/debounce/). 

leading and trailing debouncing


ericdallo:
Amazing changes reading the summary!
I'll review the code and then test it soon

>  300ms is a long time for a UX to respond

Perhaps 300ms is a little too much indeed, we can think on decrease that to a good default, it used to be 500ms since the first time I implemented and then I decreased a little bit to 300, but I had no idea what is the ideal number, if you have a way to know that, that would be great.

mainej:
> decrease that to a good default

I'm proposing something different... Right now whether we receive a single did-change event, or a burst of them, we wait until 300ms after the last one, then start the analysis. My proposal is to start an analysis immediately after the first event is received. If we receive more events soon after that, we wait until 300ms after the last one, then start a second analysis. If the client sends only one event (e.g. the user deletes a line or an unused var) then we'll run only one analysis. That analysis will start immediately, so we'll get feedback 300ms faster. But, if the client sends a burst of events (e.g. the user types a bunch of characters), we'll run two analyses, once immediately and once 300ms after the last event.

So, the tradeoff would be:
* less wait on single changes (or some bursts, see below)
* same wait on bursts, at the cost of running two analyses instead of one

FWIW emacs-lsp does some of its own buffering of did-change messages. If you quickly type a bunch of code, it won't send a did-change until you've paused. So, this proposal would be especially nice for it and other "smart" clients, because they already try to favor single changes.

Here's another picture. The top row is the raw did-change events coming in. The middle row is the analysis events we trigger currently–we wait until after the last in a cluster. The bottom row is my proposal—no wait, and then wait until after the last. Notice especially the middle cluster, which is just a single raw event.

Screen Shot 2022-03-24 at 1 56 26 PM

lib/src/clojure_lsp/kondo.clj
```diff
@@ -135,13 +134,10 @@
...
           updated-analysis (assoc (:analysis @db) filename (normalize-analysis analysis))]
       (if (settings/get db [:linters :clj-kondo :async-custom-lint?] true)
-        (async/go-loop [tries 1]
-          (if (>= tries 200)
-            (logger/info "Max tries reached when async custom linting" uri)
-            (if (contains? (:processing-changes @db) uri)
-              (do
-                (Thread/sleep 50)
-                (recur (inc tries)))
+        (future
```
ericdallo, line 137: why a future and not a go/block which is less expensive?

ericdallo:
Sounds like a plan, looking forward to testing that!

> So, this proposal would be especially nice for it and other "smart" clients, because they already try to favor single changes.

Yeah, but heads up, there are clients that doesn't do that, we had issues in the past on clients that send too many didChanges.

cli/src/clojure_lsp/server.clj
```diff
@@ -43,6 +43,11 @@
 (def change-debounce-ms 300)
 (def created-watched-files-debounce-ms 500)
 
+(defn log! [level args fmeta]
+  (timbre/log! level :p args {:?line (:line fmeta)
+                              :?file (:file fmeta)
+                              :?ns-str (:ns-str fmeta)}))
```
ericdallo, line 49: good hack!

lib/src/clojure_lsp/feature/file_management.clj
```diff
@@ -138,51 +140,58 @@
...
+                (when (settings/get db [:notify-references-on-file-change] true)
+                  (notify-references filename old-local-analysis (get-in @db [:analysis filename]) components))
+                (clojure-producer/refresh-test-tree producer [uri]))
+              (recur @db))))))))
 
 (defn did-change [uri changes version db]
   (let [old-text (get-in @db [:documents uri :text])
-        final-text (reduce handle-change old-text changes)]
+        final-text (reduce handle-change old-text changes)
+        done (promise)]
```
ericdallo, line 169: promise is a good fit for this code indeed

lib/src/clojure_lsp/handlers.clj
```diff
@@ -89,70 +94,90 @@
   (f.file-management/did-change-watched-files changes db/db))
 
 (defn completion [{:keys [textDocument position]}]
-  (let [row (-> position :line inc)
-        col (-> position :character inc)]
-    (f.completion/completion textDocument row col db/db)))
+  (shared/logging-results
+    ":completion %s - total items: %s"
```
ericdallo, line 98: glad we managed to move from server to here

lib/src/clojure_lsp/kondo.clj
ericdallo: Also, why we are removing the loop here?

ericdallo (commented):
Looks good!

I wonder if we should move the shared/logging things outside of lsp4clj, I'm not sure what's the best here, having the logger specific for clojure-lsp or generic in lsp4clj, but I guess having it in clojure-lsp is the best decision for now

lib/src/clojure_lsp/kondo.clj
mainej: I think they're equally as expensive, but if you know differently, I'd love to learn. From what I understand `async/go` (or probably `async/thread` in this case) is necessary if you need to pass the result of the body to other `async` code, but otherwise isn't needed. Aside from that, the only real difference is that `future` and `async/thread` get dispatched on different thread pools.

mainej:
> heads up, there are clients that doesn't do that, we had issues in the past on clients that send too many didChanges

Understood... I don't want to remove the debouncing entirely.

mainej:
> I wonder if we should move the shared/logging things outside of lsp4clj

I've been wondering that too. But, even though this PR removes a lot of logging, lsp4clj still does a lot of logging, most of it hidden in `coercer/conform-or-log`. So maybe not yet.

I also wonder about hooking in to slf4j, so lsp4clj and clojure-lsp each could pick their own logging library, but with results going to the same place. I always get confused when I read about slf4j, so maybe that's something someone else can tackle. :)

snoe:
> debounce

In my experience, initial and trailing can get weird. `coc` seems to send `didChange` with each character, so I probably feel this more than `lsp-mode` users. It's probably useful to test it out yourself, I would maintain emacs, vim, and vscode configs when I did the initial work.

If a client is sending per character, when you type the first letter of a long string of characters, you are probably resetting the debounce window with each subsequent letter but you'll get a diagnostic based on that first character. For a user, it feels like you are about to get as-you-type updates, but then that initial diagnostic hangs around until you stop and then you get the remainder all pop in at once.

I used to list the contents of the queue in these logs because it was very easy to process them slower than they came in and end up in a situation where we couldn't keep up. I believe the long debounce was primarily to stop analysis from backing up. This backup made the pop-in effect even worse as it could be seconds before the queue would be processed. When analysis went off thread it became harder to track in this way like the compare-and-set issues you've uncovered but did help commands being serviced on time. 

So I think we're sort of trying to handle two things. User input and analysis. Ideally our window could be very small like 16ms to handle the single change clients but also not over analyze if they happen to type at a fast 32ms cadence.

mainej:
@ericdallo please hold off on merging this... I'm having second thoughts about the `promise`.

mainej:
So, I thought the `promise` was clever, but there's a problem with it... Whenever we receive a `did-change`, we put the promise in the DB and on the `current-changes-chan`. In `analyze-changes`, we get a promise off a chan and resolve it, after analyzing the file. Any task that wants to wait for analysis plucks the promise out of the DB and waits for it to resolve. 

Sounds good at first, but the problem is that not every promise makes it through to `analyze-changes`. The channel that `analyze-changes` reads off of is a different, debounced version of `current-changes-chan`. So, any task that was waiting on a debounced promise will eventually time out.

I think I'll revert to the original implementation, with the backoff loop, and think about this another time. That's what I get for trying to make two changes in one PR. :)

ericdallo:
Makes sense, doing that in another PR sounds better

mainej:
Thanks for the context @snoe. I experimented with leading+trailing debounce a bit locally. In `lsp-mode` it's nice to get feedback a little faster, but the difference in experience is not as dramatic as I had hoped. I can see why it'd be annoying in a client that sends did-change on every character. I'll leave the code here, in case either of you want to experiment with it.

```clojure
(defn leading-debounce-by
  "Debounces `in` channel. Dispatches the first message in a burst, and if more
  arrive separated by `ms` milliseconds or less, dispatches the last `ms`
  milliseconds after it arrived. If a different message, as determined by
  `by-fn` arrives, dispatches immediately."
  [in ms by-fn]
  (let [out (chan)]
    (go-loop [state :unoccupied
              holding-val nil]
      (if (= :unoccupied state)
        (let [val (<! in)]
          #_(logger/debug "leading")
          (>! out val)
          (recur :leading val))
        (let [timer (timeout ms)
              [new-val ch] (alts! [in timer])
              different? (and new-val
                              (not (= (by-fn holding-val) (by-fn new-val))))]
          (cond
            different? (do
                         #_(logger/debug "conflicting")
                         (when (= :trailing state)
                           (>! out holding-val))
                         (>! out new-val)
                         (recur :unoccupied nil))
            (= ch timer) (do
                           (when (= :trailing state)
                             #_(logger/debug "trailing")
                             (>! out holding-val))
                           (recur :unoccupied nil))
            (= ch in) (do #_(logger/debug "debouncing")
                          (recur :trailing new-val))))))
    out))
```

lib/src/clojure_lsp/kondo.clj
mainej: I reverted this, so best to ignore.

lib/src/clojure_lsp/feature/file_management.clj
mainej: Reverted, best to ignore.

mainej:
> It's probably useful to test it out yourself, I would maintain emacs, vim, and vscode configs when I did the initial work.

Yeah, that's a good habit @snoe. I have all three (`emacs+lsp-mode` and `nvim+coc`) but don't switch over to nvim or vscode as often as I should. I did actually try the leading debouncing thing in nvim and it wasn't so bad, or at least not in my limited experiments. It doesn't spam did-change as much as I expected. Still, I'm going to wait for another day, and another PR. :)

ericdallo:
@mainej I won't include this PR on today's release so we can have more time to find any bugs after merging, ok?

mainej:
> won't include this PR on today's release

@ericdallo, yep, that's fine. It's ready whenever you are.

ericdallo:
I did need to revert the removal of that setting lint-project-files-after-startup? because it breaks integration tests for some reason, I'll let for next release, probably related with the conflict on this PR

mainej:
> I did need to revert the removal of that setting lint-project-files-after-startup? because it breaks integration tests 

I saw that. Does it break them, or make them flaky?

ericdallo:
It breaks indeed, probably should not be hard to understand and fix it

Cyrik:
Thanks for keeping me updated on this @mainej

ericdallo:
@mainej sorry for the delay, would you mind solving conflicts so I can review this?

mainej:
Conflicts resolved @ericdallo

ericdallo (approved):
Thank you!

Issue #566, opened 2021-09-16
Moving vars/function to a different namespace


martinklepsch:
After a coding spree I often realize that I'd like to rearrange some of my code. For instance I defined a utility function and would now like to also move that to the "utilities" namespace. 

In these situations it would be create to trigger a command and basically rename a var including it's namespace portion.  For example I'd like to move `my.app.core/find-first` to `my.app.util/find-first`. Besides just moving the var it would also be great if all call sites are updated accordingly and requires to `my.app.util` are added as needed.

I can't think of many alternatives other than just moving the code manually, which is tedious and error-prone. That said there probably are lots of IDEs that support this type of functionality and it might be good to take a look their UX for inspiration.

ericdallo:
Thank you for the suggestion, it makes sense indeed, I think we can rely on the existing `rename` feature, where the user can just append the new namespace and clojure-lsp would take care of the rest

mynomoto:
> Thank you for the suggestion, it makes sense indeed, I think we can rely on the existing rename feature, where the user can just append the new namespace and clojure-lsp would take care of the rest

Does this mean that I would find a usage of the var and rename it? Some questions:
1. Will it understand aliases? How would be the behavior?
2. Would it change all usages in the project?

If 2 is true it should be enough to be helpful however 1 works. I think something to make aliases consistent could be missing but that should be other issue anyway.

ericdallo:
yeah, maybe rename wouldn't work for alias cases but for 2 yes, maybe we can start with a code action that would be available when hovering over function definition?

mynomoto:
One example (not trivial and only a little tricky):
```clojure
; src/app/a.clj
(ns app.a)

(defn foo
  []
  :foo)

(defn another
  []
  :another)

; src/app/b.clj
(ns app.b
  (:require
    [app.a :as a]))

(defn bar
  []
  (a/foo))


; src/app/c.clj
(ns app.c
  (:require
    [app.a :as a.a]))

(defn cde
  []
  (a.a/foo))


; src/app/d.clj
(ns app.d
  (:require
    [app.a :as a.a]
    [app.b :as b]))

(defn fgh
  []
  (= (a.a/foo) (b/bar)))
```

The user would hover over `foo` on `app.a` and select "move to other namespace" (improvising here, details may vary). It would open an input for the user to choose the namespace. They will pick `app.b`. What I would expect to happen:
```clojure
; src/app/a.clj
(ns app.a)
; foo was removed
[...]

Issue #828, opened 2022-03-16
Documenting new refactors, naming, and mnemonics.


snoe:
I wanted to make a note of a higher level principal that I've never articulated that should be maintained as new functionality is added.

[clj-refactor.el](https://github.com/clojure-emacs/clj-refactor.el#usage) describes it:
> All functions in clj-refactor have a two-letter mnemonic shortcut. E.g. `rs` for `cljr-rename-symbol`

Their list of two-letter commands is here: https://github.com/clojure-emacs/clj-refactor.el/wiki#available-refactorings

I believe that so far, two-letter mnemonic shortcuts in clojure-lsp have either been copied from clj-refactor or are still compatible with this scheme. But, as new commands are added, we should assign them a mnemonic name and two-letter code (e.g. `promote-anonymous` : `pa`) and document these all somewhere like https://github.com/clojure-lsp/clojure-lsp/blob/master/docs/features.md.

Keeping the same mnemonics as clj-refactor, helps muscle memory for anyone who switches over and can help drive a community standard if others want to write refactor tools. Publishing our list of functions and mnemonic codes can also help editors/plugins like helm or spacemacs have consistent shortcuts to our refactorings so that people who switch editors don't have to relearn shortcuts.

snoe:
This is unrelated to code actions they aren't and shouldn't be the only way to execute things. For instance, I much prefer set bindings as code-actions jump around (clients seem to shortcut to numbers but `1` could be cycle-privacy in one place or rename in another). 

For instance, in vim it's much more natural to me to have a chord bound like `<leader>crcp` (clojure refactor+cycle-privacy). 

So I don't think the server code changes (unless code-actions can suggest shortcut keys), but documentation and conversation about existing and new functionality should specify the suggested mnemonics.

ericdallo:
No problem, it makes sense to still make available, it's just that mention mnemonics on clojure-lsp side doesn't makes sense to me, this sounds to me like a thing clients should do not the server

ericdallo:
I see, In my experience, I doubt clients will follow that as it's pretty uncommon the bindings from Vscode match with vim or emacs for example, but we can do our best to suggest what we think it's best, so a section on features.md sounds reasonable

snoe:
[...]

PR #871, opened 2022-03-25
feature: move-form

snoe:
Mostly working:
Reworked add-to-namespace add a seam where a more explicit libspec could
be used.
Added a nice test helper to turn a set of changes into a document as a
client would.

Move form incomplete pieces:
- needs to add missing lib for all namespaces used within form
- needs to handle whitespace better
- removing the source require erroneously

- [x] I created a issue to discuss the problem I am trying to solve or there is already a open issue. #566 


snoe:
```
FAIL in clojure-lsp.feature.add-missing-libspec-test/add-import-to-namespace-test (add_missing_libspec_test.clj:354)
when on an invalid location
expected: (nil? (-> (h/code "(ns foo.bar) |;; comment") (add-import-to-namespace "java.util.Date")))
  actual: (not (nil? ({:range {:row 1, :col 1, :end-row 1, :end-col 13}, :loc [<list:
  (ns foo.bar
    (:import
     java.util.Date))
> {:l [], :pnodes [<forms:
  (ns foo.bar
    (:import
      java.util.Date))
>], :ppath nil, :r nil, :changed? true}]})))
1 tests, 8 assertions, 1 failures.
```

@mainej I don't understand the purpose of this test, afaict the position of the cursor is inconsequential to the addition.

snoe:
@ericdallo I kind of want to add this sooner than later to avoid merge hell.
I believe the foundation is in place and it can be iterated on & more tests can be added in future PRs. So maybe don't advertise this initial pr or mark it as alpha or something.


here's how I'm calling it from vim: 
`nnoremap <silent> crmf :call CocRequest('clojure-lsp', 'workspace/executeCommand', {'command': 'move-form', 'arguments': [Expand('%:p'), line('.') - 1, col('.') - 1, input("File: ", "", "file")]})<CR>`

I want to spend another day on getting the checklist mostly done, especially the last point before taking this out of draft.

ericdallo:
Thanks @snoe! I'm releasing a new version right now and thought would be good to keep this PR and the logging one outside of the release to have more time to test if we find bugs soon, but we can merge this on master as soon as you finish/we review and test :)

mainej:
@snoe at one point I wrote a bunch of tests about invoking the action menu and/or commands from a top-level comment or whitespace. This caught a ton of problems, some of them new (from #716) and some of them old. As I was adding the tests, honestly, I wasn't concerned with _what_ the command did, just that it didn't do anything completely nonsensical or throw any exceptions. I'd prefer to keep some version of these tests to prevent regressions, but feel free to make them specify different behavior.

snoe:
@ericdallo This should be ready for review, I marked it as ALPHA in the changelog, but it seems pretty solid, I dogfooded quite a bit.  

### Future work considerations.
It's very strict and some feedback to the user would be very helpful.
- the destination file must exist with an ns form, 
- the cursor must be at the start paren of the form to move. 
- the form being moved cannot use any vars from the source namespace.

It does not check for cyclic dependencies, I believe that should be added somewhere around `add-to-namespace` no matter the command.

### Other stuff

My vim command for invoking is here, I chose a relative filename to the project-root as the arg because I assume most editors should be able to provide completion here.
`nnoremap <silent> crmf :call CocRequest('clojure-lsp', 'workspace/executeCommand', {'command': 'move-form', 'arguments': [Expand('%:p'), line('.') - 1, col('.') - 1, input("File: ", "", "file")]})<CR>`

I addressed much of #828 in features.md as I added the new command doc.

ericdallo:
Sounds great @snoe , I'll take a look today and test it!

> the destination file must exist with an ns form,
the cursor must be at the start paren of the form to move.
the form being moved cannot use any vars from the source namespace.

I think for now this is ok, in the future this should not be hard to make work, right? at least the first and the second

snoe:
> Sounds great @snoe , I'll take a look today and test it!
> 
> > the destination file must exist with an ns form,
> > the cursor must be at the start paren of the form to move.
> > the form being moved cannot use any vars from the source namespace.
> 
> I think for now this is ok, in the future this should not be hard to make work, right? at least the first and the second

Yup, should be pretty straightforwad

docs/features.md
```diff
@@ -388,38 +374,51 @@ For more information check the [stubs settings section](../settings.md#stub-gene
...
+|----------|-------------------------|----------------------------------------|--------------------------------------------------|---------------------------|----------------------------------------------------------------------|
+| ai       | add-import-to-namespace | Add import to namespace                | `[file-uri,row,col[,name]]`                      | ✅                         | <img src="../images/features/add-common-import.gif" width=360>       |
+| am       | add-missing-libspec     | Add missing require                    | `[file-uri,row,col]`                             | ✅                         | <img src="../images/features/add-missing-require.gif" width=360>     |
+| as       | add-require-suggestion  | Add require suggestion                 | `[file-uri,row,col,ns,alias,refer]`              | ✅                         | <img src="../images/features/add-alias-suggestion.gif" width=360>    |
+| cc       | cycle-coll              | Cycle collection `(#{}, {}, [], ())`   | `[file-uri,row,col]`                             |                           |                                                                      |
```
ericdallo, line 395:
```suggestion
| cc       | cycle-coll              | Cycle collection `(#{}, {}, [], ())`   | `[file-uri,row,col]`                             |      ✅                   |                                                                      |
```

lib/src/clojure_lsp/parser.clj
```diff
@@ -11,7 +11,7 @@
 #_{:clj-kondo/ignore [:clojure-lsp/unused-public-var]}
 (defmacro zspy [loc]
   `(do
-     (log/warn '~loc (pr-str (z/sexpr ~loc)))
+     (taoensso.timbre/warn '~loc (pr-str (z/sexpr ~loc)))
      ~loc))
```
ericdallo, line 15: I wonder if we really need to add timbre to test alias just for this function, I don't use this at all as it just replace a str or a expr

docs/features.md
```diff
@@ -300,27 +306,7 @@ User can register additional custom snippets, for more information on how to con
...
-| Move to let                        | <img src="../images/features/move-to-let.gif" width=720>             |
-| Cycle privacy                      |                                                                      |
-| Cycle fn literal                   | <img src="../images/features/cycle-fn-literal.gif" width=480>        |
-| Extract function                   | <img src="../images/features/extract-function.gif" width=360>        |
-| Thread first/all last              | <img src="../images/features/thread-first-all.gif" width=360>        |
-| Unwind thread                      |                                                                      |
-| Sort map                           | <img src="../images/features/sort-map.gif" width=720>                |
-| Suppress diagnostic                |                                                                      |
-| Create test for function           |                                                                      |
+See [below](#clojuser-lsp-extra-commands) for screenshots.
```
ericdallo, line 309: IMO we should mention both commands and code actions but give more highlights for code actions, as they are smarter than commands, as they are available when they make sense be available.

lib/src/clojure_lsp/feature/move_form.clj
```diff
@@ -0,0 +1,203 @@
...
+        ;; You could have multiple in a top form (let [x 1] (def y x) (def z y))
+        ;; we don't support that
+        multiple-defs? (= 1 (count defs))
+        can-move? (and dest-ns
+                       (empty? local-inner-usages)
+                       on-top-level-form?
+                       multiple-defs?)]
+    (when can-move?
+      (let [def-to-move (first defs)
+            refs (q/find-references analysis def-to-move false true)
```
ericdallo, line 124: What is this `true`? shouldn't we pass db there?

docs/features.md
```diff
@@ -388,38 +374,51 @@ For more information check the [stubs settings section](../settings.md#stub-gene
...
+| il       | introduce-let           | Introduce let                          | `[file-uri,row,col,name]`                        |                           |                                                                      |
+| is       | inline-symbol           | Inline Symbol                          | `[file-uri,row,col]`                             | ✅                         | <img src="../images/features/inline-symbol.gif" width=360>           |
+| ma       | resolve-macro-as        | Resolve macro as                       | `[file-uri,row,col]`                             | ✅                         | <img src="../images/features/resolve-macro-as.gif" width=720>        |
+| md       | move-coll-entry-down    | Move coll entry down                   | `[file-uri,row,col]`                             | ✅                         | <img src="../images/features/move-coll-entry.gif" width=720>         |
+| mf       | move-form               | Move form                              | `[file-uri,row,col,filename]`                    | ✅                         |                                                                      |
```
ericdallo, line 407: I was wondering, maybe we could receive `file-uri, row, col, dest-namespace`, that would be a better UX?
Also, I though about not receiving that at all, and asking it so a future code action could use the same command, similar to how we do with `resolve-macro-as` and `create-test`, WDYT?

ericdallo:
@snoe could you solve conflicts one last time so I can finish the review and test it manually?

docs/features.md
mainej: I agree. Or rather, I'd say that @snoe, the wording of the new section strongly emphasizes binding commands, before offhandedly mentioning code actions. I think it'd be a barrier to entry if everyone thought they had to set up their own keybindings to use clojure-lsp. I know you use keybindings more, and that's sort of the vim way. And I like that way too, because I use Emacs in evil mode. But I'd guess that the majority of users are going to favor the code actions. And what @ericdallo has been saying is very true... the code actions guide everyone, but particularly new users, to refactorings that are appropriate for the current cursor position. So, I'd suggest explaining code actions before invoking the "we **strongly** suggest" language for setting up keybindings.

docs/features.md
```diff
@@ -388,38 +374,51 @@ For more information check the [stubs settings section](../settings.md#stub-gene
 
 Commands that client can request. Most code actions use these commands as actions.
 
-#### Refactoring
-
-Commands that change/refactor the code. See [above](#code-actions) for screenshots.
-
-__Note: Most of them are available via code actions and most of the time you want to call the code action and not the command manually__
-
-- Clean namespace *
```
mainej, line 397: Glad to see these two nearly identical lists merged into one!

docs/features.md
snoe: I used the dest-filename because editors can autocomplete it very easily.

docs/features.md
snoe: Here's the thing, code actions don't really need documentation from us in the way that executeCommand does. It falls in the same bucket as go to def and rename; lsp client features.

Fair enough point about the ordering, so I moved the note about code actions above the keybindings paragraph.

lib/src/clojure_lsp/feature/move_form.clj
snoe: I think I had started down the path of only looking for project refs. I see @mainej has added something similar, and this can be updated later if desired. It's generally less important as it's ok imo, for this command to take slightly longer.

snoe:
@ericdallo I believe this is ready.

lib/src/clojure_lsp/parser.clj
snoe: The reason to add it to the test alias is so you can actually log through koacha, not because of this macro.

docs/features.md
ericdallo: I see, even so, as a new clojure-lsp user, I would like to know what features clojure-lsp offer, and mentioning only commands may leave users think this would work only manually.
Well, I'm ok leaving the note as the first thing though

lib/src/clojure_lsp/feature/move_form.clj
ericdallo: @snoe I mean, that find-references function, receives as last arg a `db` right? Why we are passing true instead of a db component?

docs/features.md
ericdallo: I think receiving a `ns` has a better semantic for the command as it doesn't make sense to move to a file outside the project, or not in the source-path?
We do similar on `ns` renames, we ask for the new ns name and not the filename.
I know editors can complete filenames, but as a IDE maybe we should restrict the input to make even clear what we expect to user input

ericdallo:
@snoe , thanks, just a couple of more questions/suggestions, LMK what you think

lib/src/clojure_lsp/feature/move_form.clj
snoe: right, it was a good catch, db is being passed now.

docs/features.md
snoe: I think ns has a ton of ambiguity that files don't.
Macros in cljs often have two files with the same namespace, one .clj file one .cljs file.
If we want to extend this to create a new ns, you'd still have to ask for the path.
Rename can assume the same src folder, whereas this would often be used to move from .clj to a shared .cljc src dir.

If you wanted, you can add a restriction that checks that the filename is within one of the source paths.

ericdallo: I see, makes sense I think

ericdallo:
@snoe I tested and it looks good, some things to discuss for next work on that feature:

- any reason check if zloc is the top-form one? could we just get top-form always and use that?
- I noticed we don't move/copy the requires of the inner usages, something to improve later
- I thought a little about how to make that work with code actions and I didn't find a way to ask for the dest-file via clojure-lsp :/ I know it's possible to make emacs/lsp-mode intercept that, but all clients would need to do that, and some of them probably don't have that flexibility, for some code actions we hard-code a string like extract-function, but this one I can't see it working

ericdallo (approved):

snoe:
> any reason check if zloc is the top-form one? could we just get top-form always and use that?

Yes, probably.

> I noticed we don't move/copy the requires of the inner usages, something to improve later

agree.

> code actions

Yeah, it's more similar to a rename, maybe a code-action specific command like find-suggestions where the server asks client for the file and kicks off the "real" command

ericdallo:
> Yeah, it's more similar to a rename, maybe a code-action specific command like find-suggestions where the server asks client for the file and kicks off the "real" command

Good suggestion, I'll think in something about that later

PR #875, opened 2022-03-26
Scan fewer elements when range formatting

mainej:
This is a small improvement to the work done in #795 to improve range formattting performance. Instead of scanning from the beginning of the file to find the start of the range and again to find the end, we scan to the start, and then from there to the end.


ericdallo (approved):
Looks good!

PR #877, opened 2022-03-27
Namespace suggestion regessions and improvements

snoe:
There are number of decisions that go into ns suggestions and there's a
lot of guessing as to what the user might be intending.

Consider `clojure.data.xml` that the user is trying to require.
Question if what the user types is their intended alias or a search for
a namespace, or something in between.

* `xml` feels like intended alias.
* `d.xml` feels like it could be intended.
* `c.d.x` feels like a search.
* `d.x` feels like a search or do they intend `d.xml`?

We split the given alias and possible ns into segments and follow this
logic:
1. if only a single segment is given, suggest the best segments from the
   ns (first available alias of `xml` and `data.xml`)
2. if the last given segment matches the last ns segment, use given
   alias (`d.xml` => `[clojure.data.xml :as d.xml]`)
3. if the segment counts are the same, suggest best and a plain ns
   require (`[clojure.data.xml :as xml]` and `[clojure.data.xml]`)
4. otherwise, expand the given segment to the last ns segment (`d.x` =>
   `d.xml`)

I believe there is room here for user setting to control some of this.
A setting that follows this logic by default or always pick best, given, or plain-ns could be added in the future.
I believe there's also room for the logic to only accept given prefixes
as well. ex: If your project requires `[models.* :as m.*]` everywhere only take given
if it starts with the known alias prefix `m.` otherwise fallback to best segments.

Since a given alias could match hundreds of namespaces it's important to
be restrictive on suggestions.

- [ ] I created a issue to discuss the problem I am trying to solve or there is already a open issue.

### DIFF
--- a/lib/src/clojure_lsp/feature/code_actions.clj
+++ b/lib/src/clojure_lsp/feature/code_actions.clj
@@ -71,8 +71,12 @@
 
 (defn ^:private require-suggestion-actions
   [uri alias-suggestions]
-  (map (fn [{:keys [ns alias refer position]}]
-         {:title      (format "Add require '[%s %s %s]'" ns (if alias ":as" ":refer") (or alias (str "[" refer "]")))
+  (map (fn [{:keys [ns alias refer position count]}]
+         {:title      (format "Add require '[%s%s%s]'%s"
+                              ns
+                              (if alias (str " :as " alias) "")
+                              (if refer (str " :refer [" refer "]") "")
+                              (if count (str " × " count) ""))
           :kind       :quick-fix
           :preferred? true
           :command    {:title     "Add require suggestion"
--- a/lib/src/clojure_lsp/diff.clj
+++ b/lib/src/clojure_lsp/diff.clj
@@ -1,17 +1,17 @@
 (ns clojure-lsp.diff
   (:require
    [clojure-lsp.shared :as shared]
-   [clojure.string :as str])
+   [clojure.string :as string])
   (:import
    [difflib DiffUtils]))
 
 (set! *warn-on-reflection* true)
 
 (defn- lines [s]
-  (str/split s #"\n"))
+  (string/split s #"\n"))
 
 (defn- unlines [ss]
-  (str/join "\n" ss))
+  (string/join "\n" ss))
 
 (defn unified-diff
   ([old-uri new-uri original revised project-root-uri]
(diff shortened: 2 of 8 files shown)


lib/src/clojure_lsp/feature/code_actions.clj
```diff
@@ -71,8 +71,12 @@
...
 (defn ^:private require-suggestion-actions
   [uri alias-suggestions]
-  (map (fn [{:keys [ns alias refer position]}]
-         {:title      (format "Add require '[%s %s %s]'" ns (if alias ":as" ":refer") (or alias (str "[" refer "]")))
+  (map (fn [{:keys [ns alias refer position count]}]
+         {:title      (format "Add require '[%s%s%s]'%s"
+                              ns
+                              (if alias (str " :as " alias) "")
+                              (if refer (str " :refer [" refer "]") "")
+                              (if count (str " × " count) ""))
```
ericdallo, line 79: Awesome feature! maybe we should just be a little bit more explicit on what ` x 3` means for example, maybe change to `(3 usages)`? WDYT?
```
Add require '[clojure-lsp.feature.clean-ns :as f.clean-ns]' (3 usages)
```

ericdallo:
@snoe looks good, just found a little bug:
For the code `xml/emit`, I see 2 code actions:

```
Add require '[clojure.data.xml :as xml]'
Add require '[clojure.data.xml]'
```

The second one does nothing and it prints on console `:add-require-suggestion made no changes xml/emit`

snoe:
doh, that's not really a regression on this pr but of the add-missing work from before. I thought I had fixed it but I think I ran into it in the move-form branch and fixed there. sorry about this.
https://github.com/clojure-lsp/clojure-lsp/pull/871/files#diff-6b08aa5a645f3e004348468a3d5fb7e92058a76c1bfd14b7d2f4f335463dbbadR194

So either they both go in at once, or we merge to one, or I try cherry picking some of that here.

ericdallo:
np, I'll merge this one and test on the next branch along with the new feature :)

ericdallo:
So I think we are good to merge this one, right?

ericdallo (approved):

snoe:
@ericdallo Yup, good to merge.

lib/src/clojure_lsp/feature/code_actions.clj
snoe: oh sorry, I started with that but it seemed too distracting. I think see how other people feel about it.

PR #880, opened 2022-03-27
Fix expand-let

snoe:
Transform was looking for :list but anything other than :token should be
considered the container.


ericdallo (approved):
Thanks for finding this corner case!

PR #882, opened 2022-03-28
Use JDK source if path provided

dharrigan:
If Java JDK sources are already installed (normally via some package
management), then instead of downloading the sources, allow the user to
point to where the sources can be found on the filesystem.

-=david=-

- [X] I created a issue to discuss the problem I am trying to solve or there is already a open issue.


lib/src/clojure_lsp/feature/java_interop.clj
```diff
@@ -120,16 +120,24 @@
...
+       (spit dest-jdk-result-file "1")))
+    (.getCanonicalPath dest-jdk-file) db))
+
+(defn retrieve-and-analyze! [{:keys [db]}]
+  (let [jdk-source-uri (settings/get db [:java :jdk-source-uri] default-jdk-source-uri)
+        download-jdk-source? (settings/get db [:java :download-jdk-source?] false)]
+    (when-let [jdk-source-uri' (or (and (shared/plain-uri? jdk-source-uri)
+                                        (shared/filename->uri jdk-source-uri db))
+                                   (when download-jdk-source? (download-jdk-source jdk-source-uri db)))]
+      (logger/info (str "[JDK] Analyzing JDK source found at " jdk-source-uri "..."))
```
ericdallo, line 141:
```suggestion
      (logger/info (str "[JDK] Analyzing JDK source found at " jdk-source-uri' "..."))
```

lib/src/clojure_lsp/feature/java_interop.clj
```diff
@@ -120,16 +120,24 @@
...
+      (analyze-jdk-source! jdk-source-uri' db)
+      (logger/info (str "[JDK] Source found at " jdk-source-uri " analyzed successfully.")))))
```
ericdallo, line 143: We already mention `jdk-source-uri` on line 141, I think we don't need to mention again?

dharrigan: np, will remove.

ericdallo (approved):
Thank you!

snoe:
Isn't source.zip placed in JAVA_HOME typically? Seems like that should be looked at first.

ericdallo:
@snoe yep, I'm adding support for finding automatically right now :)

dharrigan:
@snoe bear in mind that some distro's don't set java_home, for example on Arch it's not set.

PR #885, opened 2022-03-30
More quickly identify changed files in notify-references

mainej:
This removes several nested loops from the algorithm that identifies
changed incoming references to or outgoing references from a file. In
clojure2d.color, this reduces the time finding associated files from
over 30s to around 0.2s.

We haven't been logging the main body of work of notify-references, and
it happens async, so it's been out-of-sight and out-of-mind. This also
adds duration logging for notify-references. That should make it easier
to pinpoint as a performance problem later.

There are more optimizations to be done in find-references, but this is
a good start.

- [x] At least partially resolves #884 
- [x] I added a new entry to [CHANGELOG.md](https://github.com/clojure-lsp/clojure-lsp/blob/master/CHANGELOG.md)

PR #887, opened 2022-03-30
Backfill tests for finding reference files to notify

mainej:
As follow up for #885 this backfills tests for finding which other files to notify after a file is updated.

I'm not sure these tests cover everything, and in particular, they don't cover code I don't understand (the re-checking that an internal file is internal) and code I think might be wrong (removing private calls). See TODO comments in code.

### DIFF
--- a/lib/src/clojure_lsp/feature/file_management.clj
+++ b/lib/src/clojure_lsp/feature/file_management.clj
@@ -83,29 +83,33 @@
         usage-signature (juxt :to :name)]
     (find-changed-elems-by usage-signature old-var-usages new-var-usages)))
 
+(defn reference-filenames [filename old-local-analysis new-local-analysis db]
+  (let [project-analysis (into {} (q/filter-project-analysis-xf db) (:analysis @db))
+        source-paths (settings/get db [:source-paths])
+        changed-var-definitions (find-changed-var-definitions old-local-analysis new-local-analysis)
+        references-filenames (->> changed-var-definitions
+                                  ;; TODO: switch from nested-loop join to hash join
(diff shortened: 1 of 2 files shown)


lib/src/clojure_lsp/feature/file_management.clj
```diff
@@ -83,29 +83,33 @@
...
+                                  ;; TODO: switch from nested-loop join to hash join
+                                  (mapcat #(q/find-references project-analysis % false db))
+                                  (map :filename))
+        changed-var-usages (find-changed-var-usages old-local-analysis new-local-analysis)
+        definitions-filenames (->> changed-var-usages
+                                   ;; TODO: remove usages of external namespaces (clojure.core, etc.) here
+                                   ;; TODO: switch from nested-loop join to hash join
+                                   (keep #(q/find-definition project-analysis % db))
+                                   (filter (fn [d]
+                                             ;; TODO: this excludes "private calls", but they are counted in code lens, so maybe shouldn't exclude
```
ericdallo, line 100: Yeah, I thought code elens use to work without this for some reason I can't recall, I think if we try removing it and confirm code lens are updated as well, we can safely remove this private check

mainej: @ericdallo do you mind if I remove the private check as part of #892? If I remove it from this PR it'll cause conflicts with that one.

ericdallo: np

ericdallo (approved):
Awesome, thank you for that!

PR #892, opened 2022-03-31
Improve performance of notify-references

mainej:
This replaces the nested-loop joins in notify-references with hash joins.

This PR doesn't really change the performance when there aren't any files to notify. But when there are, it makes a difference.

To test I inserted an unmatched delimiter into `clojure-lsp.feature.diagnostics`, making the file invalid. This file uses 5 namespaces, and 5 other namespaces use it. They all get notified.

Before this PR, in this scenario notify-references takes about 2 seconds. About half of that is finding the files and the other half is analyzing and linting them. After the patch it takes about 1 second, with about 100ms finding files, 800ms analyzing and 100ms linting. 

For follow-on work, finding files and linting each execute `q/filter-project-analysis`, which takes about 100ms itself, meaning that almost all their time is in that function. So, making `q/filter-project-analysis` faster will improve `notify-references` further.

In addition to removing the nested loops, this PR:

* Renames diagnostic helpers to delineate which calculate custom lint data and which publish that data to clients.
* Lints all reference files in one batch, rather than linting each file individually

- [x] This is follow up work from #885

This PR builds on #887. You can either merge that one first and this second, or do them both by merging this one alone.

### DIFF
--- a/lib/src/clojure_lsp/feature/file_management.clj
+++ b/lib/src/clojure_lsp/feature/file_management.clj
@@ -87,29 +87,44 @@
 (defn reference-filenames [filename old-local-analysis new-local-analysis db]
   (let [changed-var-definitions (find-changed-var-definitions old-local-analysis new-local-analysis)
         changed-var-usages (find-changed-var-usages old-local-analysis new-local-analysis)
+        usage-signature (juxt :to :name)
+        def-signatures (fn [var-def]
+                         (map (fn [var-name]
+                                [(:ns var-def) var-name])
+                              (q/var-definition-names var-def)))
         project-analysis (into {}
                                (q/filter-project-analysis-xf db)
                                (dissoc (:analysis @db) filename)) ;; don't notify self
-        source-paths (settings/get db [:source-paths])
-        references-filenames (into #{}
-                                   (comp
-                                     ;; TODO: switch from nested-loop join to hash join
-                                     (mapcat #(q/find-references project-analysis % false db))
-                                     (map :filename))
-                                   changed-var-definitions)
-        definitions-filenames (into #{}
-                                    (comp
-                                      ;; TODO: remove usages of external namespaces (clojure.core, etc.) here
-                                      ;; TODO: switch from nested-loop join to hash join
-                                      (keep #(q/find-definition project-analysis % db))
-                                      (filter (fn [d]
-                                                ;; TODO: this excludes "private calls", but they are counted in code lens, so maybe shouldn't exclude
-                                                (and (not (:private d))
-                                                     ;; TODO: is this extra work? The definition came from project-analysis, so should be on the source-paths.
-                                                     (some #(string/starts-with? (:filename d) %) source-paths))))
-                                      (map :filename))
-                                    changed-var-usages)]
-    (set/union references-filenames definitions-filenames)))
+        references-filenames (when (seq changed-var-definitions)
+                               (let [def-signs (into #{}
+                                                     (mapcat def-signatures)
+                                                     changed-var-definitions)]
+                                 (into #{}
+                                       (comp
+                                         (mapcat val)
+                                         (filter #(identical? :var-usages (:bucket %)))
+                                         (filter #(contains? def-signs (usage-signature %)))
+                                         (map :filename))
+                                       project-analysis)))
+        definitions-filenames (when (seq changed-var-usages)
+                                (let [usage-signs->langs (->> changed-var-usages
+                                                              ;; TODO: can we easily remove usages of external namespaces (clojure.core, etc.) here?
+                                                              ;; TODO: do we really care if lang doesn't match?
+                                                              (reduce (fn [result usage]
+                                                                        (assoc result (usage-signature usage) (q/elem-langs usage)))
+                                                                      {}))]
+                                  (into #{}
+                                        (comp
+                                          (mapcat val)
+                                          (filter #(identical? :var-definitions (:bucket %)))
+                                          (filter #(when-let [usage-langs (some usage-signs->langs (def-signatures %))]
+                                                     (some usage-langs (q/elem-langs %))))
+                                          ;; TODO: this excludes "private calls", but they are counted in code lens, so maybe shouldn't exclude
+                                          (remove :private)
+                                          (map :filename))
+                                        project-analysis)))]
+    (set/union (or references-filenames #{})
+               (or definitions-filenames #{}))))
 
 (defn ^:private notify-references [filename old-local-analysis new-local-analysis {:keys [db producer]}]
   (async/go
--- a/lib/src/clojure_lsp/feature/file_management.clj
+++ b/lib/src/clojure_lsp/feature/file_management.clj
@@ -86,38 +86,33 @@
 
 (defn reference-filenames [filename old-local-analysis new-local-analysis db]
   (let [changed-var-definitions (find-changed-var-definitions old-local-analysis new-local-analysis)
+        ;; TODO: can we easily remove usages of external namespaces (clojure.core, etc.) here?
         changed-var-usages (find-changed-var-usages old-local-analysis new-local-analysis)
-        usage-signature (juxt :to :name)
-        def-signatures (fn [var-def]
-                         (map (fn [var-name]
-                                [(:ns var-def) var-name])
-                              (q/var-definition-names var-def)))
         project-analysis (into {}
                                (q/filter-project-analysis-xf db)
                                (dissoc (:analysis @db) filename)) ;; don't notify self
         references-filenames (when (seq changed-var-definitions)
-                               (let [def-signs (into #{}
-                                                     (mapcat def-signatures)
-                                                     changed-var-definitions)]
+                               (let [def-signs (->> changed-var-definitions
+                                                    (map q/var-definition-signatures)
+                                                    (apply set/union))]
                                  (into #{}
                                        (comp
                                          (mapcat val)
                                          (filter #(identical? :var-usages (:bucket %)))
-                                         (filter #(contains? def-signs (usage-signature %)))
+                                         (filter #(contains? def-signs (q/var-usage-signature %)))
                                          (map :filename))
                                        project-analysis)))
         definitions-filenames (when (seq changed-var-usages)
                                 (let [usage-signs->langs (->> changed-var-usages
-                                                              ;; TODO: can we easily remove usages of external namespaces (clojure.core, etc.) here?
                                                               ;; TODO: do we really care if lang doesn't match?
                                                               (reduce (fn [result usage]
-                                                                        (assoc result (usage-signature usage) (q/elem-langs usage)))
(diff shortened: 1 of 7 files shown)


lib/src/clojure_lsp/feature/file_management.clj
```diff
@@ -84,37 +85,56 @@
...
+                               (into #{}
+                                     (comp
+                                       (mapcat val)
+                                       (filter #(identical? :var-usages (:bucket %)))
+                                       (filter #(contains? def-signs (q/var-usage-signature %)))
+                                       (map :filename))
+                                     project-analysis)))
+        outgoing-filenames (when (seq changed-var-usages)
+                             (let [usage-signs->langs (->> changed-var-usages
+                                                              ;; TODO: do we really care if lang doesn't match?
```
ericdallo, line 107: For .cljc files, clj-kondo returns 2 analysis for the same element, one for clj and other for cljs, some times one of then have its particularity for the lang. We need to think if that could affect this notify-references feature, I don't see any problems not checking lang here, but maybe I could be missing something.

ericdallo, line 107: For .cljc files, clj-kondo returns 2 analysis for the same element, one for clj and other for cljs, some times one of then have its particularity for the lang. We need to think if that could affect this notify-references feature, I don't see any problems not checking lang here, but maybe I could be missing something.

lib/src/clojure_lsp/feature/file_management.clj
```diff
@@ -84,37 +85,56 @@
...
+                                                           (reduce (fn [result usage]
+                                                                     (assoc result (q/var-usage-signature usage) (q/elem-langs usage)))
+                                                                   {}))]
+                               (into #{}
+                                     (comp
+                                       (mapcat val)
+                                       (filter #(identical? :var-definitions (:bucket %)))
+                                       (filter #(when-let [usage-langs (some usage-signs->langs (q/var-definition-signatures %))]
+                                                  (some usage-langs (q/elem-langs %))))
+                                          ;; TODO: this excludes "private calls", but they are counted in code lens, so maybe shouldn't exclude
```
ericdallo, line 117: I think the purpose of this at the time, was to exclude private calls since they are from the same file and we don't want to lint the same file, only public calls, does that makes sense for you?

ericdallo, line 117: I think the purpose of this at the time, was to exclude private calls since they are from the same file and we don't want to lint the same file, only public calls, does that makes sense for you?

lib/src/clojure_lsp/feature/file_management.clj
```diff
@@ -84,37 +85,56 @@
...
 (defn reference-filenames [filename old-local-analysis new-local-analysis db]
-  (let [project-analysis (into {} (q/filter-project-analysis-xf db) (:analysis @db))
-        source-paths (settings/get db [:source-paths])
-        changed-var-definitions (find-changed-var-definitions old-local-analysis new-local-analysis)
-        references-filenames (->> changed-var-definitions
-                                  ;; TODO: switch from nested-loop join to hash join
-                                  (mapcat #(q/find-references project-analysis % false db))
-                                  (map :filename))
+  (let [changed-var-definitions (find-changed-var-definitions old-local-analysis new-local-analysis)
+        ;; TODO: can we easily remove usages of external namespaces (clojure.core, etc.) here?
```
ericdallo, line 89: yes, these are all db analysis, we should probably q/filter-project-analysis on both for better performance?

lib/src/clojure_lsp/feature/file_management.clj
mainej: Duplicate comment... will respond on the other one.

lib/src/clojure_lsp/feature/file_management.clj
mainej: OK, let's see if I understand... Suppose there's a function defined in two files, defs.clj and defs.cljs. And suppose in my usages.cljc file I add a call to that function. We'd want to notify both defs.clj and defs.cljs. That part seems clear. But suppose I have another file usages.clj. If I add a call in that file, is it important to notify only defs.clj, and not defs.cljs?

lib/src/clojure_lsp/feature/file_management.clj
mainej: It's a little more complex than that.... We were talking about this before in relation to custom linting a single file. It's easy to exclude usages *from* an external ns, but hard to exclude usages *to* an external ns. 

Suppose one of my project files looks like this:

```clojure
(defn foo [a] a)
```
And I change it to this:
```clojure
(defn foo [a] (+ 1 a))
```

I've added a usage of `clojure.core/+`. It doesn't matter whether I filter the analysis to be one file or only project files, that internal file will still have that **usage** of `clojure.core/+`. 

Later, when we're looking through the project analysis (which we've filtered to be only internal files), we discover that `clojure.core/+` isn't **defined** in any internal files, and so we don't need to notify anything.

But the comment is asking whether there's an easy way to exclude that usage early, before we look for any definitions. If we could do that, then we wouldn't need to check whether it's defined internally.

If we could quickly get a set of the internal namespaces, we could exclude external usages like `clojure.core/+` before looking through the project definitions. But, I think to find the internal namespaces you'd have to loop through the project analysis again. I haven't benchmarked it, but I expect that would make the whole process slower. Our set of usages contains 1 element. We'd loop through 140000 elements, finding internal namespaces, just so we could get our set of usages down to 0 elements.

In this particular example, that may save us some time. When we have 0 usages, we don't need to look for any definitions, and so we can skip that step. But if we had say, one usage of an internal var and one of an external var, the difference is whether we loop over 140000 elements once, looking for two definitions, or loop over 140000 elements once looking for internal namespaces, and then loop over 140000 elements again, looking for one definition. The former is probably faster.

Anyway, I think I've convinced myself this would be an unhelpful micro-optimization, so I'll remove the comment.

lib/src/clojure_lsp/feature/file_management.clj
ericdallo: yeah, ideally lint only the .clj one but probably the code readability here is more important than lint both langs or not, so I'm ok with whatever is better to read/maintain

lib/src/clojure_lsp/feature/file_management.clj
ericdallo: you are right, we should lint private calls even if from other files like test ones which is pretty common, we even have on clojure-lsp project some of them

lib/src/clojure_lsp/feature/file_management.clj
mainej: The code's written, so I don't mind leaving it. I'll add a comment saying it could be removed.

lib/src/clojure_lsp/feature/file_management.clj
mainej: That's a good point... even more reason to lint different-file private calls.

lib/src/clojure_lsp/feature/file_management.clj
ericdallo: Yeah, makes sense, I think we are feeling the need for a way to easily(and fast) know all the project or external analysis on multiple places of the project without filtering all the time whole analysis. Maybe we should research a little bit later how to achieve that, maybe have a pos process or something that would classify in the analysis db a `:internal` and `:external`.

mainej: 👍 To elaborate... there are two parts. First, it'd be nice to quickly know which files are `:internal` or `:external`, to speed up `q/filter-project-analysis`. Second, it'd be nice to know which namespaces are `:internal` or `:external`, to speed up filtering of usages. There's overlap between those two things, but they're slightly different.

```clojure
(def db
  {:external-files [{:file/name "..."
                     :file/uri "..."
                     :file/contents-hash "..." ;; to skip analysis of unchanged external files at init
                     :file/namespaces ['clojure.core]}]
   :internal-files [,,,]
   :analysis {"... file name ..." [,,,]}})

(defn filter-project-analysis [db]
  (let [internal-filenames (map :file/name (:internal-files db))]
    (select-keys (:analysis db) internal-filenames)))

(defn internal-usages [db]
  (let [internal-nses (->> (:internal-files db)
                           (mapcat :file/namespaces)
                           (into #{}))]
    (->> (:analysis db)
         (filter #(identical? :var-usages (:bucket %)))
         (filter #(contains? internal-nses (:to %))))))
```

ericdallo: Yes, makes sense, what makes me afraid of creating such data structure is the need to update/invalidate that as analysis change long the time, how we would update? would be fast?

mainej: Yep, I have the same worries, which is part of why I haven't tried it. A good first step would be to add `db/replace-analysis`. That would let us centralize the logic for maintaining `:analysis` and the file metadata caches. If we decided to change the format, say from a vector of internal files to a hashmap keyed by filename, then we'd only have to update a few places, instead of hunting for everywhere we do a `swap!` on `:analysis`. And we'd only need to get the update/invalidate logic correct in one place. In the future, if we decide we want other caches, for example on bucket, we'd have a central place to do that.

My intuition is that a file metadata cache would be fast, even if implemented pretty naively. Except at startup, you usually only get new analysis of one or a small handful of files. So, deciding whether they're internal or external, finding their namespaces, and inserting them into the cache would all be quick, even if for the cache you use an inconvenient data structure like a vector.

Following this idea a bit further, probably the `db` namespace should be responsible for getting analysis out of the db too, whether that's the full analysis, project analysis, single file analysis, or something else. Then the `queries` namespace could operate on vectors of elements, instead of sometimes that and sometimes a hashmap of filenames to elements.

```clojure
(into []
      (comp db/project-analysis-xf
            (q/filter-bucket-xf :var-usages))
      @db)

;; or

(into []
      (comp (db/file-analysis-xf filename)
            (q/filter-bucket-xf :var-usages))
      @db)
```

ericdallo: >  And we'd only need to get the update/invalidate logic correct in one place. In the future, if we decide we want other caches, for example on bucket, we'd have a central place to do that.

I liked it.

> Following this idea a bit further, probably the db namespace should be responsible for getting analysis out of the db too, whether that's the full analysis, project analysis, single file analysis, or something else. Then the queries namespace could operate on vectors of elements, instead of sometimes that and sometimes a hashmap of filenames to elements.

sounds valid too, maybe it could bring other benefits this metadata concept, like avoiding to lint already linted files, that sha idea

ericdallo (approved):
Looks excellent, will manually test it

ericdallo:
All working!
One thing I noticed, it seems kondo doesn't lint vars when used as `(#'my-alias/foo ...)`, not sure if this is on purpose or should be fixed on kondo, but our lint of private calls are useless, right?

Edit:

Not really, just found that if we remove the function, that private call using `#'` gets a `unresolved-var` warning, so it's still valid to lint private calls

PR #893, opened 2022-03-31
move-coll-entry enhancements

### LINKED ISSUE #891: Move-coll-entry should work at top level
So I recognize that the name of the command makes this weird, but semantics aside I believe `move-coll-entry` is a smarter transpose  (`transpose-sexps` in emacs). If the command worked to transpose `(defn...)` forms at the top of level of the namespace, it would be able to completely replace my transpose keybindings. As it is, I need to maintain both. This is useful in clojure when you need to maintain declare order of vars.

```
(def x 0)
|(def y 1)
=> move up
|(def y 1)
(def x 0)
=> move down
(def x 0)
|(def y 1)
```

mainej:
This adds two features to move-coll-entry:

* Top-level forms can be moved. #891 
* Groups inside `clojure.test/are` move intuitively.

- [x] Fixes #891

### DIFF
--- a/lib/src/clojure_lsp/shared.clj
+++ b/lib/src/clojure_lsp/shared.clj
@@ -435,5 +435,8 @@
     {:start {:line (max 0 (dec (or row name-row))) :character (max 0 (dec (or col name-col)))}
      :end {:line (max 0 (dec (or end-row name-end-row))) :character (max 0 (dec (or end-col name-end-col)))}}))
 
+(def full-file-position
+  {:row 1 :col 1 :end-row 1000000 :end-col 1000000})
+
 (defn full-file-range []
-  (->range {:row 1 :col 1 :end-row 1000000 :end-col 1000000}))
+  (->range full-file-position))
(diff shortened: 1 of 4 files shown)


lib/src/clojure_lsp/shared.clj
```diff
@@ -435,5 +435,8 @@
     {:start {:line (max 0 (dec (or row name-row))) :character (max 0 (dec (or col name-col)))}
      :end {:line (max 0 (dec (or end-row name-end-row))) :character (max 0 (dec (or end-col name-end-col)))}}))
 
+(def full-file-position
+  {:row 1 :col 1 :end-row 1000000 :end-col 1000000})
+
 (defn full-file-range []
```
ericdallo, line 441: I'm okay just changing this to a `def`

mainej:
> we should rename at least the code action

Yep, I have a branch ready that changes the terminology to "drag forward/backward". I'll submit a PR after the others are done.

lib/src/clojure_lsp/shared.clj
mainej: Ah, yes, I'll plan on doing that in a future PR.

PR #902, opened 2022-04-02
deref db earlier

mainej:
This is the beginning of a project to move derefs of the `db` as high in the call hierarchy as possible. It's not done yet @ericdallo, but I want to make you aware of it and make a few requests. /cc @snoe @Cyrik 

First, why? I expect these goals to be self-evident, but to state them explicitly:

1. Correctness. This is a multi-threaded app, with several threads updating the db. So, any pair of lines that deref the db are subject to timing bugs. It's safer to deref once earlier than several times later.
2. Debuggability. The code that does need to mutate the db is some of the trickiest to reason about. #796 was a good example. When debugging this code, it's helpful to be able to visually inspect it and know where the db might be updated. It's harder to track down the culprits when the code is sprinkled with lines like `(shared/filename->uri filename db)` which look like they might update the db, but don't.
3. Testability. Functions that take a hashmap are easier to test than ones that take a hashmap wrapped with an atom. The more you take advantage of this, the more you separate the parts of the code that mutate from the parts that don't. This is a virtuous cycle, with easier and faster tests and more isolated mutation.

As I've started this, my main observation is—wow, this is tedious! You start at the bottom of the call hierarchy and manually walk upwards, replacing atoms with hashmaps. There's a lot of iteration, and a lot of room for error. For that reason, I'd like to make a few requests:

1. When this is ready, can it be prioritized for review and merge? Between now and the time it's merged, I'll need to inspect every commit on the main branch for places where an atom is passed when it should be a hashmap and vice-versa. I'd like to minimize additional effort resolving conflicts.
2. If you have large refactorings in progress yourselves, will you let me know? I may postpone this work until the project is more quiescent. I've already run into places where my own pending PRs will conflict.
3. I'd like to rename the atom in the `db` namespace and in the `components` from `db` to `db!`. Any objections? One benefit of the old way was that you could almost always assume that the symbol `db` was an atom. As I've started this refactoring, it's becoming ambiguous whether `db` is an atom or a hashmap. I'd like to restore a visual indicator to distinguish between the two. I like `db!` because as Rich Hickey said in one of his talks "yeah, the exclamation mark means, don't do that!"


lib/src/clojure_lsp/settings.clj
```diff
@@ -76,11 +75,11 @@
 (defn all
   "Get memoized settings from db.
   Refreshes settings if memoize threshold met."
-  [db]
```
mainej, line 79: Passing the derefed db to `settings/all` and `setting/get` reveals many other places that don't need the atom. I was careful to test that the memoization works the same... The settings are cached for 1 second, after which the settings file is re-read, to check for changes. This can still lead to timing bugs, but only if the user is updating their settings file. Still better would be if the settings were calculated less frequently, much higher in the call hierarchy, so that all main entrypoints could rely on having a consistent view of the settings. But that's a separate project.

lib/src/clojure_lsp/queries.clj
```diff
@@ -334,13 +334,13 @@
           (get analysis filename))))
 
 (defmethod find-references :var-usages
```
mainej, line 336: It's not obvious, but the queries namespace went through several changes to get to this point. Now it only deals with the derefed db, never the atom. This is as it should be—a namespace called "queries" should not mutate anything.

lib/src/clojure_lsp/shared.clj
```diff
@@ -171,7 +171,7 @@
         (uri-obj->filepath uri-obj)))))
 
 (defn ensure-jarfile [uri db]
-  (let [jar-scheme? (= "jar" (get @db [:settings :dependency-scheme]))]
```
mainej, line 174: As with the `queries` namespace, the `shared` namespace now deals exclusively with a derefed db, not an atom. Again, this reveals many places that look like they might be updating the db but aren't.

ericdallo:
Thanks for this @mainej ! I agree with all those 3 points of the "why?", this is something I always tried to make few improvements like inverse of control and things like that, but the deref is hard since there are places a swap needs to happen.

> 1. When this is ready, can it be prioritized for review and merge? 

Sure, that makes total sense.

> 2. If you have large refactorings in progress yourselves, will you let me know? I may postpone this work until the project is more quiescent. I've already run into places where my own pending PRs will conflict.

My idea is to work a little more on java analysis, but nothing related to db usage I think, so this probably wouldn't affect you or I'll take care to check if that would cause issues, but makes sense, I intend to review #893 soon, so you are free to work on this.

> 3. I'd like to rename the atom in the db namespace and in the components from db to db!. Any objections?

I'm ok, but I'll leave my suggestion of `db*`, I like that `*` which means it's a ref, a future, a delay or something you could deref, but since it's a atom that could be `swap!`ed too, `db!` would work too, LMK what you think.

lib/src/clojure_lsp/settings.clj
ericdallo: yes, sounds good enough to me this way for now, nice!

lib/src/clojure_lsp/queries.clj
ericdallo: sounds perfect

lib/src/clojure_lsp/shared.clj
ericdallo: sounds perfect too!

ericdallo:
I just would like to add that I'm total in favor of pure functions that don't receive atoms but maps, things that would make all those 3 whys true, `shared`, `queries` are hard logic namespaces, they are priority to be pure indeed, so that's is a really good start!

mainej:
Whew... this is ready. I've tried to name things consistently—`db*` means you have an atom, and `db` means you have a hashmap. (With one exception that I know of, in lsp4clj.core/LSPProducer.) 

I also moved derefs earlier in the call sequence. The biggest wins are `features.code-actions/all`, and `features.refactor/call-refactor`, which both have a derefed db passed into them. That's not to say these functions are pure, but at least we know they don't modify the db. (Technically, a few of the refactorings receive `db*` via the components, but they don't use it.)

@ericdallo thanks for keeping activity in the repo light. I don't know if it was intentional, but it was certainly helpful. I'll deal with conflicts in #893 after this is merged.

Cyrik:
Big fan of this, thank you @mainej! I wanted to start doing the same thing on the shared namespace, but I'm super happy with the way you did it on the whole codebase.
As a follow-up, we might start thinking about passing just parts of the db map into some of those functions, so it's easier to see what they really need.

mainej:
@ericdallo in several places I replaced `components` with `db*` or `db`. I was thinking this morning... you probably made those locations take `components` for a reason. I think you had plans to pass through `@settings*` someday. So... I undid some of your work. Sorry about that.

When the time comes to pass both settings and the db, we might want to follow the new pattern in `call-refactor`. There we pass both `db` and `components`. That way we have a consistent view of the db, plus the full `components` map for the few refactorings that need to publish a message to the client. If this becomes a common pattern we might want something like the following, which is a declarative way of saying "nothing beyond here modifies the db or settings".

```clojure
(ns lsp4clj.components)

(defn snapshot [{:keys [db* settings*] :as components}]
  (-> components
      (dissoc :db*, :settings*)
      (assoc :db @db*, :settings @settings*)))

;; elsewhere

(let [components (components/snapshot components)]
  ,,,)
```

> passing just parts of the db map

@Cyrik good point... there's more that can be done to keep functions really focused. I tried that in a few places, but mostly decided to postpone it—this is already a giant PR.

ericdallo:
> @ericdallo in several places I replaced components with db* or db. I was thinking this morning... you probably made those locations take components for a reason. I think you had plans to pass through @settings* someday. So... I undid some of your work. Sorry about that.

Yeah 😅 , np, but it was intented, I think we should have more components like settings, that would make it easier to understand and maintain.
What we do at Nubank is: when we receive a http request we have a interceptor that assoc to the components the actual db state snapshot, I was thinking to follow with something similar here, have the db snapshot during the request input or something.

mainej:
> What we do at Nubank is: when we receive a http request we have a interceptor that assoc to the components the actual db state snapshot

Yeah, that's a common pattern in Datomic projects, and a really nice one IMO. From that perspective, `db*` is like `conn`... something you can `transact` into if you really need to, but which mostly shouldn't be touched.

ericdallo:
> Yeah, that's a common pattern in Datomic projects, and a really nice one IMO. From that perspective, db* is like conn... something you can transact into if you really need to, but which mostly shouldn't be touched.

Yes, exactly

mainej:
I'm working on resolving the conflicts in handlers.

ericdallo:
@mainej I'll try to review carefully as not only the PR is big but could be some missing `@db`-> `db` somewhere without tests

lib/src/clojure_lsp/crawler.clj
```diff
@@ -145,7 +148,7 @@
    client-settings
```
ericdallo, line 148: probably this is the most critical function to change as it needs to update and get updated values all the time

mainej: Yeah, elsewhere I tried to do `(let [db @db*] ,,,)` but not for this function. Clearly, it needs to do several sequential swaps, and so I didn't want to end up with a stale db anywhere.

mainej:
> could be some missing @db-> db somewhere without tests

I know, I'm a little worried about that too. I've been using it locally and haven't noticed problems, but there are some features I don't use very often. There are so many changes, I'm not sure you'll be able to reason about them all... It'll be easier to test and watch for exceptions.

To help avoid errors in the refactoring, I tried to do it as mechanically as possible... Rather than reason about when `db*` could be replaced with `db`, I let `clojure-lsp` guide me. I found a list of functions that used `@db*`, but otherwise didn't need `db*` and added them to my queue of things to change. For each function in the queue, I changed it to accept `db` and asked `clojure-lsp` to find all of its callers, which I changed to pass `@db*`. Often that made the caller not need `db*` either. In those cases I added the caller to my queue, and just kept working until the queue was empty. If you look back through the later commits you can see `TODO: deref` which was a note to myself about which functions were still in the queue.

In hundreds of cases, functions that originally took `db` (the atom, before the rename to `db*`) also eventually took `db` (the hashmap, after the rename). So, they don't appear in the diff. But for each one of them, they went through a stage where they were `db*`, then `@db*`, then back to `db`. So, though it's hard to believe, there are many more changed lines of code than you see in the final diff.

Since this is a classic breadth-first algorithm, it probably could have been automated, and would have been safer that way. Unfortunately, I didn't have that foresight, or ambition, so you'll have to trust I did it correctly. 😅

ericdallo:
> To help avoid errors in the refactoring, I tried to do it as mechanically as possible...

That's good.

> So, though it's hard to believe, there are many more changed lines of code than you see in the final diff.

Yeah, I realized that while reviewing, the mechanical step adds this intermediate step, but the diff is smaller which is good.

> Unfortunately, I didn't have that foresight, or ambition, so you'll have to trust I did it correctly

Np, I already did things like that on clojure-lsp and it turns out these are good refactors that helps find what is not being tested :)

I'm reviewing it but it's hard to keep track of everything Indeed and I plan to make more clojure-lsp changes, so I'm considering finishing the review and leave to keep testing it on merged master until next release. I don't want conflicts on this branch as a wrong rebase can easily introduce a regression.

ericdallo (approved):
We found and fixed issues on prepareRename which is now fixed, we intend to add integration tests for that soon.

Manually tested and it looks good, thanks for this!

PR #910, opened 2022-04-05
Rename move-coll-entry to drag

mainej:
`move-coll-entry-up` and `move-coll-entry-down` now do more than move collection entries. So, they have been renamed to `drag-backward` and `drag-forward`. The old commands still exist, but are deprecated. The action menu and docs use the new terminology.

I believe the "drag" terminology originated in Calva, though it may have been inspired by Paredit and Cursive.

This PR also takes the opportunity to clarify some of the other terminology associated with this feature.

And, when the screencasts were moved into the table recently, they got much smaller. This adds links to the gifs which I hope will link to the images in their full size. I haven't tested that. Is there a way to test the docs? If not, and if it breaks the docs site, I'll revert that part in a follow-up PR.

- [x] See discussion in #891 


ericdallo:
> Is there a way to test the docs? 

yes, `make local-webpage`

mainej:
> yes, `make local-webpage`

Thanks—I confirmed that the table links to the full size gifs.

ericdallo (approved):
Great!

ericdallo:
Sorry for the changelog conflict @mainej, also lint seems to be failing

PR #912, opened 2022-04-05
Add integration test for prepareRename

mainej:
This adds an integration test for prepareRename. I confirmed that it would have caught the [missing deref](https://github.com/clojure-lsp/clojure-lsp/pull/902/commits/fb5bef502baf8974b42dc6483c3bf392bcebb2be) @ericdallo noticed during review of #902. There are a few other integration tests that call rename... let me know if you want me to add prepareRename tests for them too, or if you want a test that demonstrates how prepareRename behaves when the cursor is on a location that can't be renamed.

### DIFF
--- a/cli/integration-test/integration/fixture.clj
+++ b/cli/integration-test/integration/fixture.clj
@@ -49,6 +49,11 @@
                  :options {:tabSize 2
                            :insertSpaces true}}))
 
+(defn prepare-rename-request [path row col]
+  (lsp-json-rpc :textDocument/prepareRename
+                {:textDocument {:uri (h/source-path->uri path)}
+                 :position {:line row :character col}}))
+
 (defn rename-request [path new-name row col]
   (lsp-json-rpc :textDocument/rename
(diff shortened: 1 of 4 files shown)


ericdallo:
I'm wokring on the ingration-tests fix, don't worry

ericdallo:
@mainej one of the main reasons we have prepareRename is for those cases where the cursor is in a loc that couldn't not be renamed, it'd be great to have a test for that, WDYT?

mainej:
@ericdallo I added an integration test for cases where the cursor can't be renamed. 

In the process I realized that we didn't short-circuit when there was no element under the cursor, e.g. `|[]`. In those cases we were still running all the queries for references, but returning `nil` in the end. lsp-mode displayed that as something like `rename not valid here`. Now we short-circuit and return a proper error message for the client to display `Can't rename, no element found.`

I also realized there weren't any unit tests for prepare-rename, so I added some. So, tests expanded and bug fixed!

ericdallo (approved):
Looks fantastic! thanks

PR #914, opened 2022-04-06
Update clojure.core completion items for 1.10 and 1.11

mainej:
This adds new definitions added in Clojure 1.10 and 1.11, or in a few cases, moves them from ClojureScript.

* I moved `ArrayChunk` from cljs to clj. clj already had `->ArrayChunk`, so now it has both. Is this correct? It's not obvious because the `->*` versions must have all been added manually to `clojure-lsp`, since they're only implied to exist.
* Same question for `Eduction`.
* There's a new symbol `ArrayManager` (of kind `:variable`). Should I add `->ArrayManager`, of kind `:reference` too? Again, it's not clear because the `->*` versions must have been added manually.
* Same question for `IVecImpl`.
* The clj syms were all calculated from the analysis of `clojure-lsp`, which uses Clojure 1.11. I didn't re-calculate the syms for ClojureScript. I could do that, but the ClojureScript [changelog](https://github.com/clojure/clojurescript/blob/master/changes.md) doesn't suggest anything has changed recently. Up to you.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
   - drag: clauses move intuitively in `clojure.test/are`
   - drag: top-level forms can be dragged #891
   - Improve completion performance for most cases, reducing time to compute clj/cljs core symbols.
+  - completion: suggest functions defined in Clojure 1.10 and 1.11
 
 ## 2022.03.31-20.00.20
 
--- a/lib/src/clojure_lsp/common_symbols.clj
+++ b/lib/src/clojure_lsp/common_symbols.clj
@@ -284,11 +284,18 @@
      {:name ->> :kind :function}
      {:name ->ArrayChunk :kind :reference}
      {:name ArrayChunk :kind :variable}
+     {:name ArrayManager, :kind :variable}
      {:name ->Eduction :kind :reference}
      {:name Eduction :kind :variable}
+     {:name IVecImpl, :kind :variable}
+     {:name NaN?, :kind :function}
+     {:name PrintWriter-on, :kind :function}
      {:name ->Vec :kind :reference}
+     {:name Vec, :kind :variable}
      {:name ->VecNode :kind :reference}
+     {:name VecNode, :kind :variable}
      {:name ->VecSeq :kind :reference}
+     {:name VecSeq, :kind :variable}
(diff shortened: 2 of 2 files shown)


ericdallo:
When clj-kondo analyzes a defrecord, it return the 3 analysis for its var-usage, a `MyRecord`, `->MyRecord` and a `map->MyRecord. Since we are adding the clj/cljs syms manually, we need to add all 3 variations

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
   - drag: clauses move intuitively in `clojure.test/are`
   - drag: top-level forms can be dragged #891
   - Improve completion performance for most cases, reducing time to compute clj/cljs core symbols.
+  - completion: suggest functions defined in Clojure 1.10 and 1.11
```
ericdallo, line 17: We already suggest those functions if user using clojure 1.10/1.11, it's just that this PR makes them available in its cache, so maybe rename this to `completion: Make Clojure 1.10 and 1.11 new functions available on completion cache`

mainej:
> we need to add all 3 variations

OK, so I think that means it's correct to have both `->Eduction` and `Eduction`. But the hard-coded syms have never included `map->*`. Do the `map->*` versions need to be added?

CHANGELOG.md
mainej: Will you double check that's true? For example, I don't get `update-vals` as a completion option until this patch is applied.

ericdallo:
> Do the map->* versions need to be added?

Yeah, I guess, but not sure how common is that, so I'm ok not adding too

mainej:
Figured it out... tldr; the patch is fine as is.

* `defrecord` defines both `->` and `map->`. 
* `deftype` defines `->` only. 
* Neither `defprotocol` nor `definterface` define either. 

Both `ArrayManager` and `IVecImpl` are `definterface`, so they don't need any other version.

Both `ArrayChunk` and `Eduction` are `deftype`, so they get a `->` version, but not a `map->` version.

ericdallo (approved):
Great!

CHANGELOG.md
ericdallo: Yeah, you are right, there is a chance we have a bug here or on kondo or some cache

PR #916, opened 2022-04-06
Add `external?` to analysis for better performance all over the project

ericdallo:
Talking with @mainej we noticed we could improve `q/filter-project-analysis-xf` and `q/filter-external-analysis-xt` if we could easily/faster know an element is external or not instead of checking filename, this PR makes that possible as a first step, adding the `:external?` keyword to all possible elements during the kondo analysis normalization.

This basically solves all performance issues with those 2 functions that are used on most queries and all over clojure-lsp, completion for example improved performance by **50%** on most cases!

I didn't test other features yet, and intend to do that tomorrow bringing more data about performance improvements

### DIFF
--- a/lib/src/clojure_lsp/db.clj
+++ b/lib/src/clojure_lsp/db.clj
@@ -19,7 +19,7 @@
 (defonce created-watched-files-chan (async/chan 1))
 (defonce edits-chan (async/chan 1))
 
-(def version 1)
+(def version 2)
 
 (defn ^:private sqlite-db-file [project-root]
   (io/file (str project-root) ".lsp" ".cache" "sqlite.db"))
--- a/lib/src/clojure_lsp/feature/completion.clj
+++ b/lib/src/clojure_lsp/feature/completion.clj
@@ -206,7 +206,7 @@
 (defn ^:private with-elements-from-alias [cursor-loc cursor-alias cursor-value matches-fn db]
   (when-let [aliases (seq (into []
                                 (comp
-                                  (q/filter-project-analysis-xf)
+                                  q/filter-project-analysis-xf
                                   (mapcat val)
                                   (filter #(identical? :namespace-alias (:bucket %))))
                                 (:analysis db)))]
--- a/lib/src/clojure_lsp/feature/completion.clj
+++ b/lib/src/clojure_lsp/feature/completion.clj
@@ -241,7 +241,7 @@
                               (contains? alias-namespaces (:ns %))
                               (or (simple-ident? cursor-value) (matches-fn (:name %))))
                      [(:ns %) (element->completion-item % cursor-alias :unrequired-alias)]))
-                distinct
+                (distinct)
                 (map
                   (fn [[element-ns completion-item]]
                     (let [require-edit (some-> cursor-loc
--- a/lib/src/clojure_lsp/feature/completion.clj
+++ b/lib/src/clojure_lsp/feature/completion.clj
@@ -206,7 +206,7 @@
 (defn ^:private with-elements-from-alias [cursor-loc cursor-alias cursor-value matches-fn db]
   (when-let [aliases (seq (into []
                                 (comp
-                                  (q/filter-project-analysis-xf db)
+                                  (q/filter-project-analysis-xf)
                                   (mapcat val)
                                   (filter #(identical? :namespace-alias (:bucket %))))
                                 (:analysis db)))]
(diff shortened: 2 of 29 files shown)


lib/test/clojure_lsp/feature/move_form_test.clj
```diff
@@ -73,22 +73,22 @@
                      "(def bar foo)")
```
ericdallo, line 73: @mainej this was the only failing test after the change, I took a quick look and it seems like a sorting issue after the change but not critical I guess, could you help confirm it please?

lib/src/clojure_lsp/db.clj
```diff
@@ -19,7 +19,7 @@
 (defonce created-watched-files-chan (async/chan 1))
 (defonce edits-chan (async/chan 1))
 
-(def version 1)
+(def version 2)
```
ericdallo, line 22: Very important change here. Since there are already projects with external analysis cache on filesystem, those analysis doesn't have the new :external, so we invalidate that cache here, we could add a `(or (:external? %) true)` all over the project, but it seems ugly, this change should not be a problem.

lib/src/clojure_lsp/feature/completion.clj
```diff
@@ -217,36 +217,38 @@
...
-                  [(:ns %) (element->completion-item % cursor-alias :unrequired-alias)]))
-             set
-             (map
-               (fn [[element-ns completion-item]]
-                 (let [require-edit (some-> cursor-loc
-                                            (f.add-missing-libspec/add-known-alias (symbol cursor-alias) element-ns db)
-                                            r.transform/result)]
-                   (cond-> completion-item
-                     (seq require-edit) (assoc :additional-text-edits (mapv #(update % :range shared/->range) require-edit)))))))))))
+          (into []
```
ericdallo, line 220: Took the oportunity to transform this code to a transducer, it makes easier to review this piece with Github "hide whitespace" option enabled:
image

mainej:
Amazing! Great work. It's a nice bonus that `db` isn't needed in so many places anymore.

lib/src/clojure_lsp/kondo.clj
```diff
@@ -114,10 +114,10 @@
        name-end-row
        name-end-col))
 
-(defn normalize-analysis [analysis]
+(defn normalize-analysis [external? analysis]
```
mainej, line 117: Could accept a hashmap `{:keys [external?]}`, if you think we'll add more context over time. Might not be necessary though.

lib/src/clojure_lsp/kondo.clj
```diff
@@ -131,27 +131,28 @@
       (assoc-in [:config :linters :unresolved-var :report-duplicates] true))))
 
 (defn ^:private custom-lint-project!
-  [db* {:keys [analysis config] :as kondo-ctx}]
```
mainej, line 134: Looking through clojure-lsp.kondo, there are more places that don't really need `db*`, and would be happy with `db`. It's tricky because the custom-lint calls are all callbacks that should deref the db when they're called, not when they callback is constructed. Anyway, I'll plan to take another look at this file some time.

lib/src/clojure_lsp/queries.clj
```diff
@@ -24,24 +23,22 @@
     nil
     coll))
 
-(defn filter-project-analysis-xf [db]
```
mainej, line 27: Could be a `def`. Same for `filter-external-analysis-xf`.

lib/src/clojure_lsp/queries.clj
```diff
@@ -24,24 +23,22 @@
...
-  (let [source-paths (settings/get db [:source-paths])]
-    (remove #(shared/external-filename? (first %) source-paths))))
+(defn filter-project-analysis-xf []
+  (remove (comp :external? first second)))
```
mainej, line 27: This took me a few minutes to figure out. Could probably use a comment explaining that it's using the first element as a proxy for whether the file is external.

lib/test/clojure_lsp/feature/move_form_test.clj
mainej: I've never looked at the move-form code, but I'll do what I can. Will report back.

mainej: From what I can see, there's code to pick from the known aliases, preferring aliases that are used the most. But, in this test, `bread` is aliased as `b` once and as `br` once. Since there's no tie breaker, they're equally valid choices. I'd suggest breaking the tie to make the code deterministic...

```diff
diff --git a/lib/src/clojure_lsp/feature/add_missing_libspec.clj b/lib/src/clojure_lsp/feature/add_missing_libspec.clj
index 81cc825b..e5a90ac5 100644
--- a/lib/src/clojure_lsp/feature/add_missing_libspec.clj
+++ b/lib/src/clojure_lsp/feature/add_missing_libspec.clj
@@ -312,8 +312,8 @@
                                                     (->> vs
                                                          (map first)
                                                          frequencies
-                                                         (sort-by val)
-                                                         reverse
+                                                         (sort-by (fn [[ns freq]]
+                                                                    [(- freq) ns]))
                                                          (into (array-map))))))
         namespaces->aliases (->> ns-alias-pairs
                                  (group-by first)
@@ -322,8 +322,8 @@
                                                          (map second)
                                                          (remove nil?)
                                                          frequencies
-                                                         (sort-by val)
-                                                         reverse
+                                                         (sort-by (fn [[alias freq]]
+                                                                    [(- freq) alias]))
                                                          (into (array-map))))))
         alias-namespaces (get aliases->namespaces cursor-namespace-str)
         namespace-aliases (get namespaces->aliases cursor-namespace-str)
```

lib/src/clojure_lsp/queries.clj
mainej: Also, if it works, `(comp :external? first val)` might be slightly clearer.

lib/src/clojure_lsp/kondo.clj
ericdallo: Yeah, maybe we can follow 1..2..n rule here, and change it next time we need a third one

lib/src/clojure_lsp/queries.clj
ericdallo: Makes sense!

ericdallo:
Done @mainej I'm doing some tests, and the improvement was wider than I thought, all find-definitions and find-references were positively affected with those changes what's is pretty good :) Almost all features that relies on analysis were improved!

ericdallo:
From my tests, `q/find-last-order-by-project-analysis` used by almost all queries, is 50% faster too, which is a lot

mainej:
> `q/find-last-order-by-project-analysis` is 50% faster

Yeah, that one makes a big difference for all the smaller find-one queries. They didn't take much time originally, but now they feel instant.

> all find-definitions and find-references were positively affected

Everywhere you removed `db` is faster. It's great.

lib/src/clojure_lsp/feature/completion.clj
```diff
@@ -217,36 +217,38 @@
...
-                           (or (simple-ident? cursor-value) (matches-fn (:name %))))
-                  [(:ns %) (element->completion-item % cursor-alias :unrequired-alias)]))
-             set
-             (map
-               (fn [[element-ns completion-item]]
-                 (let [require-edit (some-> cursor-loc
-                                            (f.add-missing-libspec/add-known-alias (symbol cursor-alias) element-ns db)
-                                            r.transform/result)]
-                   (cond-> completion-item
-                     (seq require-edit) (assoc :additional-text-edits (mapv #(update % :range shared/->range) require-edit)))))))))))
```
mainej, line 249: The logic has changed here. It used to be
```clojure
(->> elements
     (filter var-def?)
     (map (juxt ns completion-item))
     distinct
     (map with-require-edits))
```

Now it's
```clojure
(->> elements
     (filter var-def?)
     (map (juxt ns completion-item))
     (map with-require-edits)
     distinct)
```
So, two questions. 1. is it still correct? And 2. is it actually faster? Usually, moving a `distinct` later hurts performance, because you have to do extra work that will be thrown away.

mainej:
One last question (above). Otherwise looks ready!

lib/src/clojure_lsp/feature/completion.clj
ericdallo: Not I got the code snippet, but I only removed the `set` since we are `into #{}` now

mainej: Yeah, looks right 👍

PR #917, opened 2022-04-06
Add tools for profiling clojure-lsp

mainej:
This moves the criterium and clj-async-profiler deps into the main `debug` build for clojure-lsp development. It also adds a file with some profiling utilities I wish were built into those tools: `cli/dev/clojure_lsp/debug.clj`. See, in particular `auto-profile`.

I went with the generic name `clojure-lsp.debug` as @snoe suggested because I figure it'll be a good scratchpad for other kinds of debugging.

One thing I wasn't able to figure out is how to tell clojure-lsp to analyze that file. It feels strange to work on clojure-lsp without being able to use clojure-lsp. Any suggestions?

Context from Slack:

> [snoe](https://app.slack.com/team/U0BUV7XSA)  [1 day ago](https://clojurians.slack.com/archives/CPABC1H61/p1649184644366569?thread_ts=1649055048.585809&cid=CPABC1H61)
[@jacob.maine](https://clojurians.slack.com/team/U07M2C8TT) thanks for the debug-perf-cli build, I wonder if there's a reason to keep it separate from debug. I also think it might be nice to have a ./dev/clojure-lsp/debug.clj file with snippets and instructions to run a profile or a criterium.

> [ericdallo](https://app.slack.com/team/UKFSJSM38):clojure-lsp:  [1 day ago](https://clojurians.slack.com/archives/CPABC1H61/p1649184717049139?thread_ts=1649055048.585809&cid=CPABC1H61)
Yeah, both ideas sounds good to me, usually there are user.clj namespaces on lein projects, maybe we could have a performance one or something

> [jacob.maine](https://app.slack.com/team/U07M2C8TT)  [1 day ago](https://clojurians.slack.com/archives/CPABC1H61/p1649184785970549?thread_ts=1649055048.585809&cid=CPABC1H61)
Yeah, agreed. [@ericdallo](https://clojurians.slack.com/team/UKFSJSM38) I think you suggested merging the perf deps into the main debug build when I first proposed it. I can’t remember why I was hesitant then. In retrospect, it doesn’t seem like it would hurt

### DIFF
--- a/cli/dev/clojure_lsp/debug.clj
+++ b/cli/dev/clojure_lsp/debug.clj
@@ -10,14 +10,19 @@
    [clojure-lsp.db :as db]
    [clojure-lsp.feature.completion :as f.completion]
    [clojure.java.shell :as sh]
-   [criterium.core :as bench]))
+   [criterium.core :as bench]
+   [clojure-lsp.shared :as shared]))
 
 (defn uri-in-project [filepath]
-  (str (:project-root-uri @db/db*) "/" filepath))
+  (let [db @db/db*]
+    (-> filepath
+        (shared/absolute-path db)
+        (shared/filename->uri db))))
 
 (defn open-file
   "Opens a `file` by shelling out to `open`."
   [file]
+  ;; TODO: xdg-open for Linux; start for Windows
   (sh/sh "open" (str file)))
 
 (defmacro profile-times
--- a/cli/dev/clojure_lsp/debug.clj
+++ b/cli/dev/clojure_lsp/debug.clj
@@ -1,0 +1,107 @@
+(ns clojure-lsp.debug
+  "This namespace is for debugging clojure-lsp during development.
+
+  It provides a few examples of how to conduct performance analysis.
+
+  It's on the classpath when building with `make` (or `make debug-cli`) so don't
+  leave any clutter that would break the build."
+  (:require [clojure-lsp.feature.completion :as f.completion]
+            [clojure-lsp.db :as db]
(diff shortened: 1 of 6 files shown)


ericdallo:
@mainej we just need to tell clojure-lsp to consider that source-path, there are multiple ways, but the most recommended for that kind of source-paths is via the `:source-aliases`, I think we can easily make that happen with `:source-aliases #{:dev :test :debug}` on .lsp/config.edn

cli/dev/clojure_lsp/debug.clj
```diff
@@ -0,0 +1,107 @@
...
+     (println "profiling" n# "iterations")
+     (let [time-start# (System/nanoTime)
+           result# (profiler/profile {:min-width 5
+                                      :return-file true}
+                                     (dotimes [_n# n#] ~body))
+           elapsed-s# (/ (- (System/nanoTime) time-start#) 1e9)]
+       (println "Profiled for" elapsed-s# "seconds. (Should be 5-10 seconds to have enough samples.)")
+       result#)))
+
+(defmacro profile-by-runtime
```
ericdallo, line 36: Nice

ericdallo (approved):
Looks really good!

mainej:
@ericdallo `:source-aliases` worked. Thanks!

mainej:
Or rather... `:source-aliases` mostly worked. Now it highlights and refactors the file, but it doesn't lint it. Does kondo need similar config?

ericdallo:
The PR is correct anyway, thank you!

PR #922, opened 2022-04-09
promote-fn

### LINKED ISSUE #783: Promote anonymous function
I often want to promote anonymous functions to be named private functions.

When I'm on code like this:
```clojure
(map |#(* % 2) coll)
```
I'd like `extract-function` to create something like this:
```clojure
(defn ^:private new-function [element1]
  (* element1 2))

(map new-function coll)
``` 
Similarly:
```clojure
(map |(fn [x] (* x 2)) coll)
```
Should become:
```clojure
(defn ^:private new-function [x]
  (* x 2))

(map new-function coll)
```

Ideally, the parameter list of the new function should include any locals used within the function that were defined in the original enclosing scope. (Though this sounds a lot harder!)

```clojure
(let [m 2]
  (map |(fn [x] (* x m)) coll))

;; becomes

(defn ^:private new-function [x m]
  (* x m))

(let [m 2]
  (map #(new-function %1 m) coll))
```

mainej:
This implements promote-fn, as discussed in #783. It converts a literal `#()` to a `fn`, or a `fn` to a `defn`. It's similar to https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-promote-function

* Retains a fn's name, if it had one.
* If the fn used any locals from the surrounding scope, passes those to the new defn.
* Replaces the original form with the smallest safe replacement, depending on context
  * `new-function`
  * `#(new-function a b %1 %2)`
  * `(partial new-function a b)`

- [x] Resolves #783 


lib/src/clojure_lsp/queries.clj
```diff
@@ -79,20 +79,26 @@
        (= (:from usage) (:to usage))))
 
 (defn find-local-usages-under-form
-  [analysis filename line column end-line end-column]
-  (let [local-analysis (get analysis filename)]
+  [analysis filename row col end-row end-col]
```
mainej, line 82: These changes aren't crucial to the feature... I was just trying to understand how this code works and decided to clarify it a little.

lib/src/clojure_lsp/refactor/transform.clj
```diff
@@ -397,6 +397,46 @@
           (recur (first (expand-let loc false uri db)) current-edit)
           (some-> previous-edit vector))))))
 
+(defn new-defn-zloc [fn-name private? params body db]
+  (let [root (z/of-string
```
mainej, line 401: `create-function`, `extract-function` and `convert-fn-to-defn` (`promote-fn`) now all follow the same logic for creating a `defn` node.

lib/src/clojure_lsp/refactor/transform.clj
```diff
@@ -397,6 +397,46 @@
...
+
+                 :else
+                 (format "(defn- %s)" fn-name)))]
+    (reduce z/append-child*
+            (-> root
+                (z/append-child* (n/spaces 1))
+                (z/append-child* params))
+            body)))
+
+(defn prepend-preserving-comment
```
mainej, line 417: Similarly, all three functions share logic for prepending the `defn` before the current form and before any comments on that form.

mainej: This was done in two different ways before, only one of which preserved comment affiliation.

lib/src/clojure_lsp/refactor/transform.clj
```diff
@@ -433,15 +463,13 @@
...
-              defn-edit            (-> (z/of-string "\n(defn)\n")
-                                       (z/append-child fn-sym)
-                                       (z/append-child used-syms)
-                                       (z/append-child* (n/newlines 1))
-                                       (z/append-child* (n/spaces (+ defn-col 1)))
-                                       (z/append-child expr-node)
-                                       z/up)]
-          [{:loc   defn-edit
-            :range defn-range}
+              private? false ;; TODO: shouldn't this be true?
```
mainej, line 466: I wish that functions extracted with `extract-function` were private, since what I use it for most often is to create private helper functions. If that'd be an acceptable behavior change, I can submit a follow-up PR. I don't want to change it as part of this PR, because it's separate from `promote-fn`.

ericdallo: Sounds like a good idea to me

lib/src/clojure_lsp/feature/code_actions.clj
```diff
@@ -300,6 +308,9 @@
       @can-cycle-fn-literal?*
       (conj (cycle-fn-literal-action uri line character))
 
+      @can-promote-fn?*
+      (conj (promote-fn-action uri line character))
+
```
ericdallo, line 313: I wonder if those 2 could cause questions what's the difference between then when on a `#()`, since both do the same, right?
Why `promote function` couldn't move `#()` -> `defn` directly? and for the other case we could use cycle fn literal, WDYT?

mainej: Yeah, that's true, they both do the same thing. I suppose that is confusing. I think you had proposed `anything` -> `defn` earlier. OK, I'll experiment with that. It's a little harder but I'll see what I can figure out.

mainej:
This should probably handle multi-arity `fn`s too. OK... more to do still. I'll let you know when I think it's ready.

lib/src/clojure_lsp/feature/code_actions.clj
mainej: Another way to resolve the confusion would be to leave `promote-fn` as is, and replace `cycle-fn-literal` with `demote-fn`. So:

* `promote-fn`: `#()` -> `fn` -> `defn`
* `demote-fn`: `fn` -> `#()`

I think either is possible, so it's just what you think is clearest.

ericdallo: Yeah, your suggestion sounds good enough, as we would have clearer code action names for each operation.
Since cycle-fn-literal is quite new, I'm ok of just renaming it too, as a breaking change

snoe: +1 . At some future point it might be nice for `demote-fn` to inline defn too: `defn` -> `fn` -> `#()`, then `promote-fn` would look for all matching `fn` in the project and replace with call to def'd var.

```clojure
[(map |#(inc %) xs) (map #(inc %) ys)]
;; promote =>
(defn myinc [x] (inc x))
[(map |myinc xs) (map myinc ys)]
;; demote =>
[(map |#(inc %) xs) (map #(inc %) ys)]
```

lib/src/clojure_lsp/feature/refactor.clj
```diff
@@ -43,6 +43,9 @@
 (defmethod refactor :cycle-fn-literal [{:keys [loc]}]
   (r.transform/cycle-fn-literal loc))
 
+(defmethod refactor :promote-fn [{:keys [loc uri db]}]
```
snoe, line 46: I feel this should allow the function name to be passed in.

lib/src/clojure_lsp/feature/code_actions.clj
mainej: > Since cycle-fn-literal is quite new, I'm ok of just renaming it too, as a breaking change

👍

mainej: > At some future point it might be nice for demote-fn to inline defn too: defn -> fn

I was thinking that'd be a nice enhancement too. Someday.

mainej: @snoe your examples are a little confusing, and I want to make sure I'm not missing anything. The `demote` part makes sense. But the `promote` examples aren't really in line with how `promote-fn` is working in this PR. I'm not sure if that's because you're showing pseudocode or because you expect `promote-fn` to behave differently.

The first difference is that I don't plan to make `promote-fn` convert `#()` directly to `defn`, though we talked about that at one point. Instead, it'll work more like how clj-refactor worked, `#() -> fn -> defn`. 

Second, it won't replace every identical occurrence, which is what your example looks like it's suggesting. Perhaps it could, but IIRC, that's not how clj-refactor worked. Also I think it'd be very surprising, since it could replace occurrences several screens away.

Are we on the same page?

lib/src/clojure_lsp/feature/refactor.clj
mainej: Easy enough. Good suggestion.

lib/src/clojure_lsp/feature/code_actions.clj
snoe: yeah i skipped the fn step for brevity.

This is all in the someday enhancement. if demote inlines defn, promote should do the inverse, which I believe would have the effect you describe. Maybe both only operate on private defn to minimize the surface area. Maybe it would be too surprising and be a bad idea, but in concept i think i like symmetry.

mainej: I see. Thanks for the clarification. The symmetry is appealing. And I loved clj-refactor's `expand-let` feature where it [replaced identical entries](https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-expand-let). (Wouldn't it be nice if clojure-lsp did that?) But `let`s are much more local than `defn`s. Anyway, something to ponder in the future.

ericdallo (approved):
Looks great, I loved the text of code actions pretty clear :)
Will manually test later today

mainej:
Thanks @ericdallo I have a few tiny cleanups to do still. I'll leave a not here when I'm done.

mainej:
OK, this is ready @ericdallo. Thanks for reviewing!

ericdallo (approved):

ericdallo:
Thank you!

PR #933, opened 2022-04-14
Optimize finding ns def by filename

mainej:
Reduces CPU usage of refresh-test-tree and probably code-lens by querying for a file in the analysis before searching for its ns definitions, rather than finding all ns definitions and then filtering them by filename. More noticeable on startup, where we refresh the test tree of each project file individually.

refresh-test-tree can still take several seconds at startup, but now the majority of that time appears to be in `(.publishTestTree client tree)`, which is something to research later (it doesn't seem like it should be slow).

In clj-kondo this patch cuts the initial refresh-test-tree time down from about 4-5 seconds to 2 or less.

Also, fix exception when testing label isn't a string.

The first argument to clojure.test/testing is a string, but it can be constructed via `(str "test thing " thing)`, or similar. The test-tree code parses these nodes as sexps, but assumes that they're strings, causing an exception. This fixes the exception.

### DIFF
--- a/lib/src/clojure_lsp/feature/test_tree.clj
+++ b/lib/src/clojure_lsp/feature/test_tree.clj
@@ -16,8 +16,9 @@
                                                  (not= testing root-testing)))
                                           testings)]
               (shared/assoc-some
-                {:name (or (-> root-testing :context :clojure.test :testing-str)
-                           "")
+                {:name (str
+                         (or (-> root-testing :context :clojure.test :testing-str) ;; not always a string
+                             ""))
                  :range (shared/->scope-range root-testing)
                  :name-range (shared/->range root-testing)
                  :kind :testing}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
   - demote-fn: Demote a fn to a literal #()
   - *breaking* remove cycle-fn-literal, since the same refactorings can be performed with the more clearly named promote-fn and demote-fn
   - drag: Fix to drag element-wise in destructured keys, not pair-wise. #927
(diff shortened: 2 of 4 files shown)


lib/src/clojure_lsp/feature/test_tree.clj
```diff
@@ -16,8 +16,9 @@
                                                  (not= testing root-testing)))
                                           testings)]
               (shared/assoc-some
-                {:name (or (-> root-testing :context :clojure.test :testing-str)
-                           "")
+                {:name (str
+                         (or (-> root-testing :context :clojure.test :testing-str) ;; not always a string
```
ericdallo, line 20: good catch, probably clj-kondo should not name it as `-str` 😛

ericdallo (approved):
Looks great! Do you have how much this improved in performance roughly?

mainej:
I tested it on clj-kondo, clojure-lsp and rewrite-clj. The time for all of them was cut in half, approximately. But I don't think that's a great number to advertise. The reason is two-fold. First, none of them took more than about 5 seconds, before the patch, which is much shorter than I remember. Perhaps one of the other performance fixes had already sped up those queries. And second, I didn't have a full benchmark set up, so those numbers aren't very rigorous. When I `time`d the refresh-test-tree code in isolation in clj-kondo, it went from 1700ms to 30ms. So, I expected a big drop like that. But when I look at the logs, there's still several seconds of time running the initial refresh-test-tree. So, I don't understand what's happening there, and it's possible that there's still some CPU hog in some situations.

ericdallo:
Yeah, for now looks a great improvement, let's keep an eye if we find any bug outlier in some project or report, but I'm happy with the improvements

PR #935, opened 2022-04-23
Integration test cleanup

mainej:
I've been working on understanding the flaky tests #704. I still don't have any real progress to report there, but I figured I'd submit the cleanup I've done so far.

See the commits for a more granular description of what each change does.

### DIFF
--- a/cli/integration-test/integration/lsp.clj
+++ b/cli/integration-test/integration/lsp.clj
@@ -107,14 +107,14 @@
   (reset! server-responses {})
   (reset! server-notifications [])
   (reset! client-request-id 0)
-  (when *clojure-lsp-listener*
-    (client-log :red "closing:" "test cleanup")
-    (flush)
-    (future-cancel *clojure-lsp-listener*)
-    (alter-var-root #'*clojure-lsp-listener* (constantly nil)))
+  (flush)
   (when *clojure-lsp-process*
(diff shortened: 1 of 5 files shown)


cli/integration-test/integration/lsp.clj
```diff
@@ -128,65 +135,44 @@
...
+        (swap! server-responses dissoc (:id params))
         (if (:error response)
           response
           (:result response)))
       (do
         (Thread/sleep 500)
-        (recur (get @server-responses @client-request-id))))))
+        (recur)))))
+
+(defn await-first-and-remove! [pred coll*]
```
ericdallo, line 149: nice

cli/src/clojure_lsp/server.clj
```diff
@@ -205,8 +213,6 @@
         timbre-logger (doto (->TimbreLogger db*)
                         (logger/setup))
         _ (logger/info "[SERVER]" "Starting server...")
-        is (or System/in (lsp/tee-system-in System/in))
-        os (or System/out (lsp/tee-system-out System/out))
```
ericdallo, line 209: I can't recall why we have tee-system-in, if that is not needed indeed, we could remove from lsp4clj in the future

ericdallo (approved):
Looks good!

PR #938, opened 2022-04-25
Centralize kondo merge

mainej:
After invoking kondo, several codepaths took responsibility for merging the analysis results into the db. They each did this in similar but slightly different ways. This patch moves all that logic into the db namespace, standardizing how kondo results are saved.

Aside from deleting a lot of duplicate code, this also helps prepare for caching internal project analysis by reducing the number of locations that will have to be updated to make that change.


lib/src/clojure_lsp/crawler.clj
```diff
@@ -46,34 +38,22 @@
       "Analyzing external classpath"
       progress-token)))
 
-(defn ^:private analyze-external-classpath!
+(defn ^:private classpath-analysis
```
mainej, line 41: Can I get a careful review of this, to make sure this refactoring is correct?

Originally both `analyze-classpath!` and `analyze-external-classpath!` swapped the kondo results into the db, which (I think) was redundant, since one was called by the other. I made it into two functions, one to calculate the kondo results (`classpath-analysis`) and another to swap into the db (`analyze-classpath!`). Does that seem correct?

lib/src/clojure_lsp/feature/java_interop.clj
```diff
@@ -132,33 +132,22 @@
...
-        (recur @db*)))
-    analysis))
-
-(defn ^:private cache-jdk-source-analysis! [analysis new-checksums db*]
-  (loop [state-db @db*]
-    (when-not (compare-and-set! db* state-db (-> state-db
-                                                 (update :analysis merge analysis)
-                                                 (update :analysis-checksums merge new-checksums)))
-      (logger/warn java-logger-tag "Analyzis outdated from java analysis, trying again...")
-      (recur @db*)))
```
mainej, line 155: This also needs a careful review. Originally both `analyze-jdk-source!` and `cache-jdk-source-analysis!` merged the analysis into the db, even though they were called in succession. I made it into one function that analyzes (and saves once) and caches.

Also, I replaced a `loop + compare-and-set!` with `swap!`. My reasoning is that `swap!` does a `loop + compare-and-set!` internally anyway. This loses an error log, but I doubt we would have ever seen that error because the time between the `deref` and the `compare-and-set!` is tiny, and so the `compare-and-set!` will practically always succeed.

lib/src/clojure_lsp/feature/stubs.clj
```diff
@@ -48,19 +48,14 @@
...
-                      (lsp.kondo/normalize-analysis true)
-                      (group-by :filename))]
-    (loop [state-db @db*]
-      (when-not (compare-and-set! db* state-db (update state-db :analysis merge analysis))
-        (logger/warn "Analyzis divergent from stub analysis, trying again...")
-        (recur @db*)))
+        result (-> result
+                   (update :analysis dissoc :namespace-usages :var-usages)
+                   (db/normalize-kondo {:external? true}))]
+    (swap! db* db/merge-kondo-results result)
```
mainej, line 54: This is another place where I replaced a `loop + compare-and-set!` with `swap!`. See above for why I think this is safe.

lib/src/clojure_lsp/shared.clj
```diff
@@ -421,8 +421,7 @@
        result#)))
 
 (defmacro logging-task [task-id & body]
-  (with-meta `(let [msg# (str ~task-id " %s")]
-                (logging-time msg# ~@body))
+  (with-meta `(logging-time (str ~task-id " %s") ~@body)
```
mainej, line 424: This restores file line numbers to the logs, something I accidentally broke recently.

lib/src/clojure_lsp/crawler.clj
ericdallo: I think they were doing different things originally, `analyze-classpath!`  was saving into db `:analysis` and other related stuff, while `analyze-external-classpath!` was upserting only `:kondo-config`, right?

These 2 functions splitted were a result of previous refactorings, where we had analyze-source-paths calling `analyze-classpath`, I think we could merge those into 1 since it' s only being called to analyze external ones?

lib/src/clojure_lsp/db.clj
```diff
@@ -104,3 +106,98 @@
...
+       name-end-col))
+
+(defn ^:private normalize-analysis [external? analysis]
+  (for [[bucket vs] analysis
+        v vs
+        element (entry->normalized-entries (assoc v :bucket bucket :external? external?))
+        :when (valid-element? element)]
+    element))
+
+(defn normalize-kondo
```
ericdallo, line 162: I was trying to avoid couple the db namespace with kondo, maybe we can move those specific kondo functions to `kondo.clj`?

lib/src/clojure_lsp/crawler.clj
ericdallo: or if you prefer, have a function to side effect swap and another to return the built map (I think that's what you did right?)

mainej: It ended up all in one function. It seems easy enough to read, to me, but I'd be happy to separate it if you prefer.

lib/src/clojure_lsp/db.clj
mainej: @ericdallo, now all the `clojure-lsp.kondo/run-*!` functions return normalized data.

Generally, I think this is cleaner, but has one small drawback in terms of the plans to cache project analysis. In the first patch, the benefit of coupling the db ns to kondo was that `:external?` was being passed into `db/with-kondo-results`. The db ns could have used that key to decide whether it should cache the new analysis. Now it'll be a bit more work to make that decision. It's an easy problem to fix, so I'm not worried about it, and generally I'm happier with how responsibilities are divided up at the moment.

One final question—the db namespace still has a function called `merge-kondo-results`. Would you prefer if it doesn't use the word "kondo" at all?

lib/src/clojure_lsp/kondo.clj
```diff
@@ -130,29 +138,29 @@
...
 (defn ^:private custom-lint-files!
-  [files db* {:keys [analysis] :as kondo-ctx}]
-  (shared/logging-task
-    :reference-files/lint
-    (let [db @db*
-          new-analysis (group-by :filename (normalize-analysis false analysis))
-          updated-analysis (merge (:analysis db) new-analysis)]
-      (f.diagnostic/custom-lint-files! files updated-analysis kondo-ctx))))
+  [files db {:keys [config analysis] :as kondo-ctx}]
+  (when (run-custom-lint? config)
```
mainej, line 153: This didn't check whether to run custom lint before. I think this is a bug fix?

lib/src/clojure_lsp/kondo.clj
```diff
@@ -235,39 +237,82 @@
...
+                  (update-callback index batch-count)
+                  (run-kondo-on-paths! paths db* normalization-config)))
            (reduce shared/deep-merge)))))
 
 (defn run-kondo-on-reference-filenames! [filenames db*]
-  (catch-kondo-errors (str "files " (string/join ", " filenames))
-    (kondo/run! (kondo-for-reference-filenames filenames db*))))
+  (let [db @db*
+        normalization-config {:external? false
+                              :ensure-filenames filenames}
```
mainej, line 295: There's a chance this fixes a bug, though I didn't confirm whether it was really a bug in the first place.

My hypothesis is that if before the initial file was changed a reference file had some findings, and that after the changes it had none, we were leaving the original findings in the db. With this `:ensure-filenames`, we should be clearing those findings.

lib/src/clojure_lsp/db.clj
ericdallo: not sure, there is a :kondo-config key [here](https://github.com/clojure-lsp/clojure-lsp/pull/938) as well, so not sure, still reviewing the code, but I'd like too keep all/most of kondo specific things on kondo.clj, although we are quite coupled to kondo analysis structure on multiple places already 🤷🏻‍♂️

lib/src/clojure_lsp/kondo.clj
ericdallo: yeah, almost sure this fixes that behavior, I was about to debug why that happens, quite annoying, will check later if that fixed it indeed

lib/src/clojure_lsp/crawler.clj
```diff
@@ -26,56 +26,16 @@
...
-        empty-findings (reduce (fn [map filename] (assoc map filename [])) {} filenames)
-        new-findings (merge empty-findings (group-by :filename (:findings result)))]
-    (swap! db* (fn [db]
-                 (-> db
-                     (update :analysis merge analysis)
-                     (assoc :kondo-config (:config result))
-                     (update :findings merge new-findings))))
-    analysis))
+  (let [result (lsp.kondo/run-kondo-on-reference-filenames! filenames db*)]
+    (swap! db* db/merge-kondo-results result)))
```
ericdallo, line 34: I still have the feeling `merge-kondo-results` is something that could be on kondo ns, not that related with db but the result map we would create from kondo analysis?

lib/src/clojure_lsp/kondo.clj
ericdallo: Yes, I probably missed that when refactoring references lint feature 😅

lib/src/clojure_lsp/kondo.clj
```diff
@@ -114,13 +95,40 @@
...
 
-(defn normalize-analysis [external? analysis]
+(defn ^:private normalize-analysis [external? analysis]
   (for [[bucket vs] analysis
         v vs
         element (entry->normalized-entries (assoc v :bucket bucket :external? external?))
         :when (valid-element? element)]
     element))
 
+(defn ^:private normalize
```
ericdallo, line 105: not related with your refactor, but this PR made me realize how much we iterate and "normalize" analysis all the time, I wonder how much time we lose here, maybe there are quick wins

ericdallo (commented):
I liked the changes overall, glad we moved the normalize stuff to a single place, that was a missing tech debt indeed, also the code looks easier to evolve which is good, the way we ensure filenames will have updated analysis/filenames looks great, LMK if you make any more changes

lib/src/clojure_lsp/crawler.clj
mainej: I'll work on moving `merge-kondo-results` over to the kondo ns. Will let you know when that's done.

lib/src/clojure_lsp/kondo.clj
mainej: Perhaps... something to research sometime.

lib/src/clojure_lsp/db.clj
mainej: Moved `db/merge-kondo-results` to `kondo/db-with-results`.

mainej:
@ericdallo I moved `db/merge-kondo-results` to `kondo/db-with-results`. Let me know if you want any other changes.

ericdallo (approved):
looks great, I didn't test that much but intend to keep testing on master during this week, although the changes are pretty specific and easy to track I think

PR #943, opened 2022-04-27
Avoid processing duplicate alias/ns completions

snoe:
In metabase/metabase, I was getting very slow completion times.
When profiling, I had a mean time of around 2500ms when completing `meta`

With this change I'm getting a mean time around 350ms. The biggest
culprit was doing `add-known-aliases` multiple times for the same
`alias/to` tuples.

This also revealed a bug where the completion-item was returning labels
that didn't match the input, this caused coc to throw them away.

So I kept the information I needed and distinct the alias/to tuple. Then
depending on which matched, show a different label/detail than
completion-item (which was pretty aenemic for these cases anyways). This
did throw away the `:data` portion but the source namespace for an alias
is actually unwanted here as it causes duplication in completions.

before:
image

after:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - General
   - Bump clj-kondo to `2022.04.25`.
+  - Improve speed of alias/ns completions.
+  - Change alias/ns completions to return a label that matches the input.
 
 - Editor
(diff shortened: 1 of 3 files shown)


ericdallo (commented):
Looks good @snoe could you fix the tests?

snoe:
@ericdallo thanks, there you go.

ericdallo (approved):
Thank you!

PR #950, opened 2022-04-29
Skip var bodies when analyzing deps

mainej:
This decreases un-cached startup time by 60-70%. See https://github.com/clj-kondo/clj-kondo/pull/1674.

This also triples the batch size of the dep analysis, since clj-kondo seems to analyze about 3 times as many files in the same amount of time now. I don't have numbers on how much memory was being saved by batching analysis, so that needs to be double-checked. @ericdallo can you do that or give some guidance on what to look for?

This also skips all linting, and analysis of var-usages for deps since we don't use those. The performance gain is almost exclusively from skipping var bodies, however, so that's the interesting part.

- [x] I created an issue to discuss the problem I am trying to solve or an open issue already exists. See https://github.com/clj-kondo/clj-kondo/pull/1674

### DIFF
--- a/lib/src/clojure_lsp/kondo.clj
+++ b/lib/src/clojure_lsp/kondo.clj
@@ -189,6 +189,16 @@
                   (f.diagnostic/sync-publish-diagnostics! uri @db*))))))
         (f.diagnostic/custom-lint-file! filename updated-analysis kondo-ctx)))))
 
+(def ^:private config-for-internal-analysis
+  {:arglists true
+   :locals true
+   :keywords true
+   :protocol-impls true
+   :java-class-definitions true
+   :java-class-usages true
+   :context [:clojure.test
+             :re-frame.core]})
+
 (defn ^:private config-for-paths [paths db]
   (-> {:cache true
        :parallel true
@@ -200,13 +210,7 @@
 (defn ^:private config-for-internal-paths [paths db custom-lint-fn]
   (-> (config-for-paths paths db)
       (assoc :custom-lint-fn custom-lint-fn)
-      (assoc-in [:config :analysis]
-                {:arglists true
-                 :locals false
-                 :keywords true
-                 :protocol-impls true
-                 :java-class-usages true
-                 :java-class-definitions true})))
+      (assoc-in [:config :analysis] config-for-internal-analysis)))
 
 (defn ^:private config-for-external-paths [paths db]
   (-> (config-for-paths paths db)
@@ -244,14 +248,7 @@
          :config-dir (project-config-dir (:project-root-uri db))
          :custom-lint-fn custom-lint-fn
          :config {:output {:canonical-paths true}
-                  :analysis {:arglists true
-                             :locals true
-                             :keywords true
-                             :protocol-impls true
-                             :java-class-definitions true
-                             :java-class-usages true
-                             :context [:clojure.test
-                                       :re-frame.core]}}}
+                  :analysis config-for-internal-analysis}}
         (with-additional-config (settings/all db)))))
 
 (defn ^:private run-kondo! [config err-hint]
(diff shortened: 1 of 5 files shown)


lib/src/clojure_lsp/kondo.clj
```diff
@@ -194,14 +194,31 @@
...
+(defn ^:private config-for-internal-paths [paths db custom-lint-fn]
+  (-> (config-for-paths paths db)
+      (assoc :custom-lint-fn custom-lint-fn)
+      (assoc-in [:config :analysis]
+                {:arglists true
+                 :locals false
+                 :keywords true
+                 :protocol-impls true
+                 :java-class-usages true
+                 :java-class-definitions true})))
```
mainej, line 209: The config for a single file includes `:context [:clojure.test :re-frame.core]`. Should that be the same here?

ericdallo:
> can you do that or give some guidance on what to look for?

Usually this is something to be tested with graal to make sure it's ok, I can help with that

lib/src/clojure_lsp/kondo.clj
ericdallo: these configs include the following features from clj-kondo analysis:
- `:clojure.test`: return more analysis related to test, crucial for test-tree feature, IMO completely not needed for external analysis :)
- `:re-frame.core`: More analysis related to reframe subscriptions/events, not needed for external as well, usually this is something needed for projects that use re-frame for cljs projects

mainej: Sorry... I should have been clearer. I'm not suggesting we add those to external analysis. What I mean is, when we do full-project analysis and referenced-file analysis (internal, but with many paths) should we be including clojure.test and reframe analysis?

ericdallo: Yes, that's weird that didn't cause a issue until now 🤔

ericdallo (approved):
Codewise looks good, will keep testing it manually

lib/src/clojure_lsp/kondo.clj
```diff
@@ -16,7 +16,7 @@
 (defn clj-kondo-version []
   (string/trim (slurp (io/resource "CLJ_KONDO_VERSION"))))
 
-(def clj-kondo-analysis-batch-size 50)
+(def clj-kondo-analysis-batch-size 150)
```
ericdallo, line 19: Something I noticed with this change, is that we have a little bit less feedback on the progress bar of editor and CLI when analyzing external deps since we have fewer batches, this is ok since performance is way more important than feedback.
I wonder if we could use some kondo hook to improve that, for example: we know there are 80 paths on classpath passed to kondo, every path linted we update the progress from a kondo hook or something, sounds like something nice to explore later

lib/src/clojure_lsp/kondo.clj
mainej: Weird... well, I added it to the full-project and referenced-file analysis. 🤷

lib/src/clojure_lsp/kondo.clj
ericdallo: and we could use this for the 95% (Project code analysis), making it more real with more precise updates

mainej: That's true... I like the idea of a progress hook. clj-kondo already has a mechanism for printing dots "." when it's in its `:text` output format, so at least a little of the groundwork is there.

ericdallo: started a discussion [here](https://clojurians.slack.com/archives/CHY97NXE2/p1651283825804139)

ericdallo:
I didn't notice any issues with that batch size on native-image, it actually seems to have improved memory usage as kondo is not holding that much data anymore which is great, I'll work on https://github.com/clj-kondo/clj-kondo/issues/1675 so we can "fix" the feedback, besides that we can merge this.
Thank you for the awesome improvement!

PR #955, opened 2022-05-01
Fix drag for calva

mainej:
Two improvements to drag, noticed while working on exposing drag to Calva.

1. Wait for editor to apply edits before requesting cursor re-positioning. Fixes out-of-order handling of these two requests.
2. Request edit only of changed clauses, not entire parent, reducing flicker.

### DIFF
--- a/cli/integration-test/integration/client.clj
+++ b/cli/integration-test/integration/client.clj
@@ -49,13 +49,16 @@
             (colored color (str "Client " client-id " " msg))
             (colored :yellow params))))
 
+(def ^:private wire-lock (Object.))
+
 (defn wire-send [server-in params]
   (let [content (json/generate-string params)]
     (binding [*out* server-in]
-      (println (str "Content-Length: " (content-length content)))
-      (println "")
-      (println content)
-      (flush))))
+      (locking wire-lock
+        (println (str "Content-Length: " (content-length content)))
+        (println "")
+        (println content)
+        (flush)))))
 
 (defn ^:private lsp-rpc [{:keys [request-id]} method params]
   {:jsonrpc "2.0"
--- a/cli/integration-test/integration/client.clj
+++ b/cli/integration-test/integration/client.clj
@@ -99,12 +99,14 @@
   (exit [_this] ;; wait for shutdown of server to propagate to listener
     @@listener)
   (send-request [this method body]
-    (let [req (lsp-rpc this method body)]
+    (let [req (lsp-rpc this method body)
+          p (promise)]
       (log this :cyan "sending request:" req)
+      ;; Important: record request before sending it, so it is sure to be
+      ;; available during receive-response.
+      (swap! sent-requests assoc (:id req) p)
       (wire-send server-in req)
-      (let [p (promise)]
-        (swap! sent-requests assoc (:id req) p)
-        p)))
+      p))
   (send-notification [this method body]
     (let [notif (lsp-rpc this method body)]
       (log this :blue "sending notification:" notif)
--- a/cli/integration-test/integration/client.clj
+++ b/cli/integration-test/integration/client.clj
@@ -1,0 +1,192 @@
+(ns integration.client
+  (:require
+   [cheshire.core :as json]
+   [integration.helper :as h])
+  (:import
+   [java.time LocalDateTime]
+   [java.time.format DateTimeFormatter]))
+
+(def ^:private ESC \u001b)
+
+(def ^:private colors
+  {:black     (str ESC "[30m")
+   :red-bg    (str ESC "[41m")
+   :red       (str ESC "[31m")
+   :green     (str ESC "[32m")
+   :yellow    (str ESC "[33m")
+   :blue      (str ESC "[34m")
+   :magenta   (str ESC "[35m")
+   :cyan      (str ESC "[36m")
+   :white     (str ESC "[37m")
+   :underline (str ESC "[4m")
+   :reset     (str ESC "[m")})
+
+(defn ^:private colored [color string]
+  (str (get colors color) string (:reset colors)))
+
+(defn ^:private content-length [json]
+  (+ 1 (.length json)))
+
+(def ^:private ld-formatter DateTimeFormatter/ISO_LOCAL_DATE_TIME)
+(defn ^:private local-datetime-str [] (.format ld-formatter (LocalDateTime/now)))
+
+(defprotocol IClient
+  (start [this])
+  (shutdown [this])
+  (exit [this])
+  (send-request [this method body])
+  (send-notification [this method body])
+  (receive-response [this resp])
+  (receive-request [this req])
+  (receive-notification [this notif]))
+
+(defprotocol IMockClient
+  (mock-response [this method body]))
+
+(defn ^:private log
+  ([{:keys [client-id]} color msg params]
+   (println (local-datetime-str)
+            (colored color (str "Client " client-id " " msg))
+            (colored :yellow params))))
+
+(defn wire-send [server-in params]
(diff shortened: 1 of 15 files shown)


cli/integration-test/integration/client.clj
```diff
@@ -0,0 +1,192 @@
...
+        (if-let [elem (first (filter pred @coll*))]
+          (do
+            (swap! coll* #(->> % (remove #{elem}) vec))
+            elem)
+          (do
+            (Thread/sleep 500)
+            (recur (inc tries))))
+        (do
+          (log client :red "timeout waiting:" coll-type)
+          (throw (ex-info "timeout waiting for client to receive req/notif" {:coll-type coll-type})))))))
```
mainej, line 168: This is new code. It's unrelated to the rest of this PR... I'm tired of waiting the full 10 minutes for the flaky tests to time out. Now, if the server request doesn't arrive within 10 seconds, the tests fail.

cli/integration-test/integration/code_action_test.clj
```diff
@@ -48,6 +48,9 @@
         :kind "source.organizeImports"}]
       (lsp/request! (fixture/code-action-request sample-file-name 5 4)))
 
+    (lsp/mock-response :workspace/applyEdit {:applied true})
+    (lsp/mock-response :window/showDocument {:success true})
```
mainej, line 52: These are required ... without them, the server waits forever for the client to respond to applyEdit, and never sends showDocument.

mainej:
I've never been so excited to see a completely red test suite!

First, the unit test failures are all in `clojure-lsp.feature.clean-ns-test` and seem to be related to a recent commit to master that changed the formatting of `:import` statements. @ericdallo will you take a look at those?

Second, the integration tests all fail in the same way. There are two things that are really interesting about this, pointing to a possible cause of the test flakiness.

First, they all failed in the same place, in `integration.code-action-test` (this test is frequently, although not always, the location of the flakiness). They don't fail on my machine at that location. So, if something is going wrong only on CI, it's likely to be around the code exercised in those tests. It will only help target the research a little, but every little bit helps.

Second, thanks to some new logging in the `integration.client` namespace, a few lines before the failure we now see "Client 5 received response for unmatched request". I thought that line would be impossible to reach... but clearly it isn't. In some prior research on the flaky tests I had come to the conclusion that the flakiness could come from a failure in the penultimate client request. The server responds, and then dies, and so the final client request is never answered. The fact that this error line shows up just before the tests fail makes me very suspicious. Maybe this is a real lead 🤞

ericdallo:
> @ericdallo will you take a look at those?

Oh, I was sure everything was passing, my bad, will fix it asap

> The fact that this error line shows up just before the tests fail makes me very suspicious. Maybe this is a real lead crossed_fingers

Makes sense!

cli/integration-test/integration/client.clj
ericdallo: awesome!
The only concern I have is when tests really take some time to finish, I mean, non graalvm tests usually takes way more time since we start the server multiple times, do you see any chance to affect that?

mainej:
Hmm... the fix for the integration tests was essentially just to record the request before sending it. That wouldn't have been a problem with earlier implementations of the integration test client, so perhaps it won't help flaky tests. On the other hand, since this PR changes the code near the common flaky tests—by waiting for the client to process edits before requesting cursor positioning—there's a chance. If it fixes the flakiness, it won't be a very satisfying fix, because I don't understand how it would have helped. But I'd take it anyway!

ericdallo:
@mainej tests fixed, feel free to rebase with master

ericdallo:
> But I'd take it anyway!

Me too :)

lib/src/clojure_lsp/handlers.clj
```diff
@@ -360,7 +360,9 @@
       :execute-command
       ;; TODO move components upper to a common place
       (when-let [{:keys [edit show-document-after-edit]} (refactor command arguments components)]
-        (producer/publish-workspace-edit producer edit)
+        (some-> (producer/publish-workspace-edit producer edit)
+                ;; wait for client to apply edit before showing doc/moving cursor
+                deref)
```
ericdallo, line 365: Shouldn't we deref [here](https://github.com/clojure-lsp/clojure-lsp/pull/955/files#diff-1793b003b94c68bc4b15e7fa7c0fd503012b02fc891b2045ddd82240051d1a99L367) as well?

ericdallo:
Looks good, especially the more robust client, nice!

cli/integration-test/integration/client.clj
mainej: I don't think it'll be a problem... this isn't a timeout for the whole test, or even for the entire lifetime of a server, just for a moment in its lifetime. More specifically we use this helper when we've done something (A) that'll cause the server to send a message to the client asynchronously (B). We do A and then start waiting, so, the timeout applies to the time delta between A and B.

In the sample project, no commands, or analysis, or anything like that should take more than 10 seconds. The one time I can think of where this delta might get anywhere near 10 seconds is startup, which is tested in `integration.initialize-test`. But even there it looks like pretty safe. There the client sends the server a request to initialize. The server creates a bunch of $/progress notifications asynchronously, then responds with its capabilities. The test waits for this response as its first assertion. Its second assertion uses the helper in question to assert that $/progress was sent. Since it should have already been sent, there shouldn't be any wait. Finally, the test sends a notification from the client to the server that the client has been initialized. Then it uses the helper to start waiting for the server to ask the client to register file watchers. I'm not sure whether that file-watcher request happens async during server init, or if it's triggered async when the client says it's finished init. n either case, since we've already waited for the slow part—server init—it shouldn't take 10 more seconds for the file-watcher request to come through.

Anyway, that's a very long description. tldr; I think we're safe.

ericdallo: yes, makes sense, thanks for the explanation.
BTW, officially, any request can only be sent after server responded with initialized

lib/src/clojure_lsp/handlers.clj
mainej: I suppose so. We wouldn't want to tell the editor to navigate to a file before it had been created. Will change and check that the integration tests still work.

mainej:
See https://github.com/clojure-lsp/clojure-lsp/pull/955/commits/313f1d1a020c5ae9cc4787e967304603056e5289 for another potential fix for test flakiness.

mainej:
Same failure, in same location, so I don't think the flake is fixed. https://github.com/clojure-lsp/clojure-lsp/runs/6248867759?check_suite_focus=true

It seems like it fails slightly earlier now (IIRC before it failed after finishing the executeCommand, and now it fails during it) so maybe something's different. Will have to think about it more.

In the meantime, I'll re-run the test suite to double-check this failure is just flake.

mainej:
The tests passed on the second run, so this is ready for a final review.

The flakiness is still around. I'll keep thinking about that. The reason that the logs have changed slightly is that we used to send both applyEdit and showDocument asynchronously, and return from executeCommand immediately. Now we wait for the client to respond to applyEdit before finishing executeCommand. But, since the client never sees the applyEdit request, it never responds, and so we never finish executeCommand. That's why we used to see :execute-command in the server log and now we don't.

My conclusion from that, which I think I knew before, is that either something happens before .applyEdit is called, which makes it impossible for it to write a message to the client (e.g., a blocked queue), or .applyEdit itself breaks and fails to write the message, or the client listener fails to receive the message. If IIRC, the lsp4j trace logs show that it's sending the message, pointing to the third possibility, though not completely excluding the possibility that lsp4j traces the message before actually sending. Or in other words... no progress on understanding this.

PR #971, opened 2022-05-06
Optimize semantic tokens

### LINKED ISSUE #970: Semantic tokens should be refreshed quickly, even in large files
**Describe the bug**

After dragging a clause in a large file, it takes an excessively long time to fix semantic tokens. At the moment for me in clojure.core it's taking 25s between the drag and the time that the tokens are fixed. The time seems to be getting longer the more I play with it. 

**To Reproduce**
1. Go to the definition of `defmacro`, in clojure.core.
2. Drag clauses within the let form
3. Observe that it can take seconds or tens of seconds for the semantic token coloring to correct itself.

What's particularly strange is that the logs show that clojure-lsp is actually returning the semantic token data pretty quickly. After the edit, there's a 300ms debounce time, followed by 1500-2000ms of clj-kondo analysis time, and then 30ms later the tokens are returned. Let's say 2 seconds on average. 

However, watching the CPU tells a different story. clojure-lsp uses an entire core for tens of seconds after the edit. The editor, even though it's been given the fixed semantic token data, doesn't make any changes until clojure-lsp is done doing whatever it's doing. Unfortunately, clojure-lsp doesn't emit any logs at the end of its CPU hogging, so I don't know what it's been doing.

This was noticed in Calva, but, the experience in Emacs is similar. Long pauses, getting longer the more I play with it.

**Expected behavior**
The semantic tokens are fixed soon after the file is analyzed.

mainej:
This replaces a quadratic algorithm with a linear algorithm when calculating semantic tokens.

To find the prior token (while converting absolute tokens to relative), the system was using `nth` on a seq, which is a linear time operation (as opposed to on a vector, where it would be a constant time operation). Since it was doing this in a loop, semantic token calculation was a quadratic time operation overall.

This switches from `nth` to `(partition 2 1)`, creating tuples of the previous and subsequent token. This brings semantic token calculation back to a linear time operation.

In `clojure.core` this reduces the time to calculate the full semantic tokens from 7s to 0.3s, a 95% reduction in time. There's a corresponding reduction in CPU usage. There isn't a noticeable change in range semantic tokens, because they're a much smaller list.

This also ensures that the logs report the full time to calculate semantic tokens. Before they were logging the time it took to prepare a lazy sequence that would eventually be realized into the full list of semantic tokens.

Fixes #970.

- [x] I created an issue to discuss the problem I am trying to solve or an open issue already exists. #970

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Editor
   - Fix drag in `are` when `clojure.test` is aliased. #967
+  - Reduce time to calculate semantic tokens, reducing CPU usage in large files. #970
 
 ## 2022.05.03-12.35.40
 
--- a/lib/src/clojure_lsp/feature/semantic_tokens.clj
+++ b/lib/src/clojure_lsp/feature/semantic_tokens.clj
@@ -160,7 +160,7 @@
   [elements]
   (->> elements
        (sort-by (juxt :name-row :name-col))
-       (map
+       (keep
          (fn [{:keys [bucket] :as element}]
            (cond
              (contains? #{:java-class-definitions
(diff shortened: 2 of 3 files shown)


ericdallo (approved):
Amazing!
I was not aware we had performance issues on clojure-lsp, the logs were always fast enough to me, glad we fixed with the doall.

Should we try to `doall` in a higher level to avoid this kind of issue in other features?

Also, if we really want to make semantic tokens fast, we should consider implementing support for delta semanticTokens, this is on LSP 3.16 spec but I found too difficult to implement on server-side and never really tried, feel free to give it a try if you think it'd help, I know emacs and vscode have already support for that.

mainej:
Moving `doall` higher would help expose similar deceptive logs. Generically, the responses are all hashmaps, which may contain lazy seqs as values. The only way I know of to recursively realize lazy seqs is with `(clojure.walk/postwalk identity result)`, which works because `clojure.walk/postwalk` recursively calls `clojure.walk/walk`, which consumes seqs with `doall`. I always worry that walking responses will be slow (e.g., you'd call `identity` on every long in the semantic tokens list, before calling `doall` on the list as a whole.) But that concern may not be justified—it's possible it'd only add microseconds to the time. Anyway, it's something to look into sometime.

Thanks for the pointer to the delta semanticTokens spec. Those deltas are a clever representation—I see how they'd make the responses tiny. But I'm with you—I can't picture an algorithm that could calculate those deltas efficiently, even if you cached the token lists between edits. I assume Microsoft has some clever way of doing it. It sounds like a good research project—got any summer interns?

PR #976, opened 2022-05-08
Fix powershell.exe command execution's error

cdzwm:
When open a project root folder(https://github.com/thheller/shadow-cljs) using VS Code in Windows 7, I got the following error message from Calva:

Classpath lookup failed when running powershell.exe -NoProfile npx shadow-cljs classpath. Some features may not work properly. Error: The argument 'npx' is not recognized as the name of a script file. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.

I think the right command should be powershell.exe -NoProfile -Command npx shadow-cljs classpath

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
   - Allow `:linters` to be configured by LSP clients passing string keys/values. #977
 
 - Editor
+  - Fix powershell.exe command execution error on Windows
   - Fix drag in `are` when `clojure.test` is aliased. #967
   - Reduce time to calculate semantic tokens, reducing CPU usage in large files. #970
   - Fix JDK global cache load when jdk source were already analyzed on previous sessions, fixing finding definition of jdk classes not working sometimes.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,12 @@
 
 - General
   - Allow specify how many classes clojure-lsp should check before moving sorted package imported classes to next line. #966
+  - Allow `:linters` to be configured by LSP clients passing string keys/values. #977
 
 - Editor
   - Fix drag in `are` when `clojure.test` is aliased. #967
   - Reduce time to calculate semantic tokens, reducing CPU usage in large files. #970
+  - Fix JDK global cache load when jdk source were already analyzed on previous sessions, fixing finding definition of jdk classes not working sometimes.
 
 ## 2022.05.03-12.35.40
 
--- a/cli/src/clojure_lsp/clojure_coercer.clj
+++ b/cli/src/clojure_lsp/clojure_coercer.clj
@@ -3,10 +3,7 @@
    [clojure.spec.alpha :as s]
    [lsp4clj.coercer :as coercer])
   (:import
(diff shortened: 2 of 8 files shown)


cdzwm:
Using the following config in .lsp/config.edn can fix the error:
{:project-specs [{:project-path "shadow-cljs.edn"
                  :classpath-cmd ["pwsh" "-NoProfile" "-Command" "npx" "shadow-cljs" "classpath"]}]}

ericdallo:
@cdzwm if it makes sense to always use `-Command`, a PR for that is welcome, but I'm no windows user, so we would need to know if that would work for all windows users

cdzwm:
On Window 7 I can  successfully execute the command '**powershell.exe -NoProfile -Command npx shadow-cljs classpath**', but  '**powershell.exe -NoProfile npx shadow-cljs classpath**' caused error message: 'The argument 'npx' is not recognized as the name of a script file. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.'

Without -Command option powershell.exe will use 'npx' as a powershell script file to execute, but 'npx' is a shell command.
With -Command option powershell.exe will use 'npx' as a shell command to execute.

powershell.exe cand run on Windows 7+ and Windows Server 2003 + default.

So the sulution can work on almost all main Windows versions platform.

cdzwm:
I think it's more appropriate to use the command 'cmd.exe -NoProfile -Command npx shadow-cljs classpath' because cmd.exe can be used on more windows versions.

ericdallo (approved):
Looks good to me, but I'll ping @bpringe here as I know we had Calva users with issues on windows before, so I'd like to see what he thinks too

bpringe:
I haven't really been a Windows user in a while. I feel like there's a reason powershell was used over cmd though either in Calva or with this command in the past.

@PEZ Do you remember anything about that? It may have been something to do with quoting though, and that's not really of concern here.

bpringe:
Either way, it seems it's a simple thing to change later if there are issues.

PEZ:
Calva used powershell to execute the `clojure` command, which is a powershell module.

Here I think it is more a thing of ”Powershell started to get bundled with Windows 7” (or something, long time ago, anyway). That said, the cmd.exe command line looks good and should work. (I think powershell can be disabled on Windows via a setting, even.)

cdzwm:
1. The official command line tool CLJ & Clojure runs under Powershell.
2.  I think most developers should use Windows 7 and Windows 2003 Server or higher to develop. 
3. Calva also uses Powershell to run the command line. 
4. Powershell is less likely to be disabled.
Taking all these considerations together, PowerShell seems like a better fit.

ericdallo:
Yeah, I think pwsh is pretty common and never saw any user complain about it with clojure-lsp, and I wouldn't like to introduce a new bug in a release and need to fix it with another one because of a corner case which we are not sure, I think we can use the new `-Command` arg just to fix issues like yours @cdzwm , and keep using pwsh, WDYT?

ericdallo:
@cdzwm could you add a entry to change log please?

ericdallo (approved):
Thanks for the contribution!

PR #977, opened 2022-05-08
Support string config for :linters to support lsp clients

dhleong:
If not using a file for global config, and instead using non-clojure
language clients (eg: neovim's native LSP in lua) there's no way to
provide keywordized keys for the `:linters` config. This PR updates the
settings "cleaning" to process the `:linters` config similarly to how we
do for `:cljfmt` to support such clients.


ericdallo (approved):
Looks good, thanks!

PR #987, opened 2022-05-19
Fix several links in the docs

mainej:
1. Fix a link from [clojure-lsp.io/features/#stub-generation](https://clojure-lsp.io/features/#stub-generation) to [clojure-lsp.io/settings/#stub-generation](https://clojure-lsp.io/settings/#stub-generation).
1. Make the links in the doc files work on both GitHub and clojure-lsp.io.

The change has been verified with `make local-webpage`.

This patch uses the lowest-common denominator for both site generators:
* Paths are relative to `docs/`, as that works for both mkdocs and GitHub
* File extension is included, as that's required for GitHub, and correctly munged by mkdocs.
* No image sizes, as GitHub-flavored markdown doesn't support it. Sizing images is less important than it used to be—most of the screen-caps are in a table now, so they're automatically sized.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -42,4 +42,4 @@ It's possible to load clojure-lsp as a babashka pod giving access to the `clojur
 
 clojure-lsp will check for `.lsp/config.edn` in the project or home dir, but it's possible to force override the settings via the `:settings` option of the API or `--settings` option of the CLI.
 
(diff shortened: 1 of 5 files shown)


ericdallo (approved):
Awesome! thanks for that

PR #999, opened 2022-05-30
Add flake to build clojure-lsp

jlesquembre:
I was able to build clojure-lsp with [clj-nix](https://github.com/jlesquembre/clj-nix)

The dependencies with a version range are still tricky, that's the reason for the `maven-extra` attribute. Not sure if it's good idea to generalize that approach in clj-nix.

Anyways, I'm opening this PR to share the approach, you can try to run clojure-lsp from the flake with 

```
nix run github:jlesquembre/clojure-lsp/flake#cli -- --version
```


ericdallo:
Amazing @jlesquembre ! I'll test it and let you know!

ericdallo:
> I don't think we can do anything on clojure-lsp, it will require a change on clojure tools.deps. I'm not sure about the best solution for this one. I could update the lock file to save that information (but it could be tricky) or just add those files if needed (easiest solution as far as we have just 1-2 dependencies with version ranges.

I see now, no problem, your changes seem enough for now

> Does it fails too with nix run github:jlesquembre/clojure-lsp/flake#cli -- --version?

Yes, same error with latest commit:

```
$ nix --version
nix (Nix) 2.8.1
```

```
                       55.4s (15.1% of total time) in 152 GCs | Peak RSS: 6.70GB | CPU load: 5.33
------------------------------------------------------------------------------------------------------------------------
Produced artifacts:
 /build/clj-kondo (executable)
 /build/clj-kondo.build_artifacts.txt
========================================================================================================================
Finished generating 'clj-kondo' in 6m 7s.
buildPhase completed in 6 minutes 8 seconds
@nix { "action": "setPhase", "phase": "installPhase" }
installing
total 115096
drwx------ 2 nixbld nixbld       100 May 31 15:58 .
drwxr-x--- 9 nobody nixbld      4096 May 31 15:52 ..
-rwxr-xr-x 1 nixbld nixbld 117839600 May 31 15:58 clj-kondo
-rw-r--r-- 1 nixbld nixbld        24 May 31 15:58 clj-kondo.build_artifacts.txt
-rw-r--r-- 1 nixbld nixbld      6998 May 31 15:58 env-vars
install: cannot stat 'clojure-lsp': No such file or directory

```

Is there any kind of cache I should purge?

jlesquembre:
um, I'm quite confused with the error, looks like it's building clj-kondo, but the size of the generated binary I see on your output is almost the same  that my clojure-lsp binary.

I'm trying to reproduce it, but not sure how. I don't think there is any cache you should purge, especially if you tried the `nix run github:...` command.

For the record, my nix version is 2.8.1 too.

I'll report any findings

ericdallo:
@jlesquembre I saw this issue some time ago!
It was happening to me when building the clojure-lsp binary on the nixpkgs.. it seems it's finding the name of the binary name from [here](https://github.com/clj-kondo/clj-kondo/blob/master/resources/META-INF/native-image/clj-kondo/clj-kondo/native-image.properties#L1), probably the correct fix would be remove that imageName from the classpath and use it only on the clj-kondo build

jlesquembre:
@ericdallo ok, I think I know what's happening, the argument order to the `native-image` command is important, and the uberjar can override the options we pass on the command line. I guess that setting the name as the last argument should fix the issue. I also guess that we get different results because the uberjar generation is not fully deterministic.

I updated the PR, could you try once again? I hope that now it works

ericdallo:
It works @jlesquembre ! I'll merge since this is working, we can now work on docs and recommending using this flakes, thank you very much for the help!

jlesquembre:
Nice to hear @ericdallo ! 

The main reason I marked the PR as WIP is that I thought about updating the lock file format to support maven version ranges, but I'll not do it for now. I can always create a new PR if I decide to update the lock format.

If you decide to merge it, keep in mind that we'll need to update the lock file after a change in any of the `deps.edn` files. We could automate that with a github action, since we just need to execute this command: `nix run github:jlesquembre/clj-nix#deps-lock`

ericdallo:
Oh, makes sense, a GitHub action would be really handy

ericdallo (approved):

Issue #980, opened 2022-05-15
Find definition doesn't work on cljs macros that definition comes from clj/cljc files


ericdallo:
https://clojurians.slack.com/archives/CPABC1H61/p1652531342831589

Repro : Try to find definition of `reagent/with-let` [here](https://github.com/lassemaatta/macro_repro/blob/master/src/cljs/macro_repro/core.cljs#L13)

It seems related with clj-kondo not returning a var-definition for that macro on the `reagent.core` cljs ns which probably makes sense since the macro exists only on JVM side.

We could:
- make clj-kondo return any info that would say that macro is defined on other ns
- make clojure-lsp try to find the definition on JVM file (the reagent.core clj ns)

PR #1003, opened 2022-06-03
Avoid finding clj def from cljs usage

mainej:
```clojure
;; a.cljs
(ns a)
(def f)

;; a.clj
(ns a)
(def f)

;; b.cljs
(ns b
  (:require [a]))
|a/f
```

In the above code, find-definition should always go to `a.cljs`, but it could go to `a.clj`. Which one it picked was arbitrary, because both definitions match.

This bug was accidentally introduced in d5e1ef39bc17cd8bb7f94bf6c9f611335dd6b2af, which was developed to fix https://github.com/clojure-lsp/clojure-lsp/issues/980. It might have been caught earlier because there are tests around this. Unfortunately, the tests only fail if the cljs file is loaded before the clj file, which isn't how they were written originally.

This patch also addresses #980, but avoids the incorrect navigation.

### DIFF
--- a/lib/src/clojure_lsp/queries.clj
+++ b/lib/src/clojure_lsp/queries.clj
@@ -135,14 +135,23 @@
 
 (defmethod find-definition :var-usages
   [analysis element]
-  (find-last-order-by-project-analysis
-    #(and (identical? :var-definitions (:bucket %))
-          (= (:name %) (:name element))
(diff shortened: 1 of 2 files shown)


ericdallo (approved):
Looks good, a little bit more complexity on the queries side which makes me wonder if we will need to handle that in the future for more queries/cases, but for now looks a efficient/good enough fix. Thanks!

Issue #1028, opened 2022-06-08
performance: `notify-references` could split linting into dependents and dependencies


mainej:
**Describe the bug**
When notifying references that a file ha changed, dependents should respond to definition changes, and so need clj-kondo's unresolved-var and invalid-arity lints. Dependencies should respond to usage changes, and so need clojure-lsp's unused-public-var lint. It might be more efficient to process the two separately, asking for different kinds of linting of each. We'd have to confirm that this actually helps performance.

ericdallo:
Yes, not sure what we should do in code exactly, but sounds like a good way to distinguish linting

mainej:
`file-management/reference-filenames` currently calculates separate sets of dependent and dependency files, then merges them together. But instead of merging them, we could run different jobs on each set. For the dependency files, we could calculate, save and publish their unused-public-var lint, bypassing kondo. For the dependent files, we could send them through kondo, but without a custom-lint-fn.

ericdallo:
I see now, sounds like a good move indeed

PR #1033, opened 2022-06-09
Improve performance of notify references on file change.

### LINKED ISSUE #1027: performance: `notify-references` could skip analysis
**Describe the bug**
When we call `notify-references`, we don't expect that the reference files' contents have changed, only that their lint may have changed. This needs verification. It's possible that the analysis _does_ actually change, even if the contents don't change. And it needs confirmation that it improves performance. If either assumption is wrong, we can leave it as is.

ericdallo:
Fixes #1027

@mainej WDYT about this? I think it'd be nice to have tests about notify-references.

I didn't test performance yet about this change too, but it's probably faster

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Editor
   - Add support to rename namespace of namespaced keywords like re-frame events/subs. #978
   - Improve performance of find-declaration feature. #1021
+  - Improve performance of notify references on file change. #1027
 
 ## 2022.05.31-17.35.50
 
--- a/lib/src/clojure_lsp/crawler.clj
+++ b/lib/src/clojure_lsp/crawler.clj
@@ -30,10 +30,10 @@
     (swap! db* lsp.kondo/db-with-results result)))
 
 (defn analyze-reference-filenames! [filenames db*]
-  (let [result (lsp.kondo/run-kondo-on-reference-filenames! filenames db*)]
+  (let [result (lsp.kondo/run-kondo-lint-only-on-reference-filenames! filenames db*)]
     (swap! db* lsp.kondo/db-with-results result)))
 
-(defn lerp "Linear interpolation" [a b t] (+ a (* (- b a) t)))
+(defn ^:private lerp "Linear interpolation" [a b t] (+ a (* (- b a) t)))
 
 (defn ^:private analyze-external-classpath! [root-path source-paths classpath settings progress-token {:keys [db* producer]}]
   (let [ignore-directories? (get settings :ignore-classpath-directories)]
--- a/lib/src/clojure_lsp/feature/file_management.clj
+++ b/lib/src/clojure_lsp/feature/file_management.clj
@@ -118,22 +118,24 @@
     (set/union (or incoming-filenames #{})
                (or outgoing-filenames #{}))))
 
-(defn ^:private notify-references [filename old-local-analysis new-local-analysis {:keys [db* producer]}]
-  (async/go
-    (shared/logging-task
-      :notify-references
-      (let [filenames (shared/logging-task
-                        :reference-files/find
-                        (reference-filenames filename old-local-analysis new-local-analysis @db*))]
-        (when (seq filenames)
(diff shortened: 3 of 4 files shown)


mainej:
@ericdallo before you merge this can I test performance a little? I tried something similar and didn't really see any improvement. That said... I had a lot of other changes going on at the same time, and so I wasn't convinced I was testing the right thing.

mainej:
I did some tests in metabase. These aren't the most rigorous, but here are the results anyway.

Before the patch was applied, I did 3 experiments. 

1. I removed and restored a var usage, which affected one ref file.
2. I removed and restored a var definition, which affected eight ref files.
3. I removed and restored a different var definition, which affected sixteen ref files.

| change applied                | reference files affected | time analyzing+linting |
|-------------------------------|-------------------------:|-----------------------:|
| remove/restore var usage      |                        1 |                  300ms |
| remove/restore var definition |                        8 |                 1300ms |
| remove/restore var definition |                       16 |                 2850ms |

Then I applied the patch and re-ran the same experiments.

| change applied                | reference files affected | time linting |
|-------------------------------|-------------------------:|-----------------------:|
| remove/restore var usage      |                        1 |                  275ms |
| remove/restore var definition |                        8 |                 1150ms |
| remove/restore var definition |                       16 |                 2250ms |

So, I'd say this is a performance improvement, but a very modest one. WDYT @ericdallo? Is this worth it? At the moment, I'm leaning toward having the best possible analysis rather than saving a few hundred milliseconds every once in a while.

ericdallo:
> At the moment, I'm leaning toward having the best possible analysis rather than saving a few hundred milliseconds every once in a while.

Yeah, usually I think about the complexity tradeoff, it seems that calling kondo ignoring analysis doesn't add that much complexity, the main change is on the `normalize` function which is basically a new if.
I'm ok proceeding with this PR because of that, it's an improvement and the analysis are not really outdated, all that we care is updated so maybe it's ok?

On the other hand, metabase is a huge project, most other projects will not suffer that much improvement, so I'm kind of still wondering if we should merge this 😅

ericdallo:
Yeah, you are right, I think the note is a good idea but postpone seems better, if we start using any of those fields on var-usages we will hardly remember of this outdated analysis even with a note on that function.
I think we can leave this PR here for now for some time and see how #1028 goes

mainej:
@ericdallo I think I understand reference notification better now...

First and foremost, I don't think we have a great reason to work on improving the performance of reference notification. It seems reasonably fast. From this perspective, I'm inclined to close #1027 and maybe #1028, without any patches.

Second, I think #1027 was misleading... I suggested that we should run kondo for its lint, but skip its analysis. But there's not much to gain from that. When kondo is linting, it doesn't cost it too much to do analysis at the same time, and it doesn't cost us too much to incorporate that analysis. That's why this patch didn't show big improvements. And there is something to lose. We could end up with invalid analysis, as discussed above.

What #1027 _should_ have said is that we should skip kondo entirely—both analysis and lint—but only on dependencies. But the details of that idea are better captured by #1028.  By doing #1028 we would skip unnecessary work, and we wouldn't have invalid analysis.

So, in summary I think we should close #1027 and this PR. Thank you for taking my half-baked idea and experimenting with it! 🙏  But I no longer think it's the right direction to go.

Do you want to leave #1028 open? Like I said, the performance of reference notification isn't bothering me right now. There are some pathological things you can do which will cause many files to be re-analzyed, like deleting a helper from a `utils` namespace. But those are pretty rare. Most often only one or two references are analyzed and that happens asynchronously. So to me it's OK if that's not optimally efficient.

If you *do* want to have #1028 implemented, it would be easier if we had already moved custom lint after kondo, as [outlined here](https://github.com/clojure-lsp/clojure-lsp/pull/1053#issuecomment-1172610962). It's not necessary, but by doing that work first, we would already have tools to add and remove custom lint, independent of running kondo, which would be useful for #1028.

ericdallo:
Agreed, I do think things changed since this PR, and I see deps-graph helping with that performance boost soon.
I'm ok closing both, though the idea of removing custom-lint if possible is something to consider indeed

PR #1035, opened 2022-06-09
Fix suggesting alias to wrong lang

### LINKED ISSUE #1024: bug: `add-missing-libspec` will suggest an alias to the wrong language
**Describe the bug**

`add-missing-libspec` will suggest an alias to the wrong language.

```clojure
;; foo/s.cljs
(ns foo.s)
;; a.cljs
(ns a (:require [foo.s :as s]))
;; b.clj
(ns b)
|s/thing
```

This will suggest `(:require [foo.s :as s])` even though `foo.s` isn't a clj namespace.

This is because `find-alias-ns-pairs` and `find-missing-ns-alias-require` should consider file lang but don't. There is a test for this "aliases > Do not add from wrong language" but it can be made to fail if `foo.s` is defined in a cljs file. That is, the tests pass because the namespace isn't defined, not because it's in the wrong language.

mainej:
This prevents `add-missing-libspec` from suggesting an alias to a file in the the wrong language.

```clojure
;; foo/s.cljs
(ns foo.s)
;; a.cljs
(ns a (:require [foo.s :as s]))
;; b.clj
(ns b)
|s/thing
```

This would suggest `(:require [foo.s :as s])` even though `foo.s` isn't a clj namespace.

Fixes #1024

- [x] I created an issue to discuss the problem I am trying to solve or an open issue already exists. #1024
- [X] I added a new entry to [CHANGELOG.md](https://github.com/clojure-lsp/clojure-lsp/blob/master/CHANGELOG.md)

### DIFF
--- a/lib/src/clojure_lsp/queries.clj
+++ b/lib/src/clojure_lsp/queries.clj
@@ -510,34 +510,42 @@
                      (= (:name-end-col %) (:name-end-col keyword-element))))
        first))
 
-(defn find-all-ns-definition-names-xf []
+(def find-all-ns-definitions-xf
   (comp
     (mapcat val)
-    (filter #(identical? :namespace-definitions (:bucket %)))
+    (filter #(identical? :namespace-definitions (:bucket %)))))
+
+(def find-all-ns-definition-names-xf
+  (comp
+    find-all-ns-definitions-xf
     (map :name)))
 
 ;; TODO remove it, use only transducer one?
 (defn find-all-ns-definition-names [analysis]
   (into #{}
-        (find-all-ns-definition-names-xf)
+        find-all-ns-definition-names-xf
         analysis))
 
-(defn find-all-aliases
-  [analysis uri db]
-  (let [langs (shared/uri->available-langs uri)]
-    (into #{}
-          (comp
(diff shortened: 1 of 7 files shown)


lib/test/clojure_lsp/feature/add_missing_libspec_test.clj
```diff
@@ -211,6 +211,7 @@
     (testing "Do not add from wrong language"
```
mainej, line 211: This test was passing not because of the mismatch between languages as its name suggests, but because `foo.s` wasn't defined and because undefined namespaces aren't suggested. Adding the definition of `foo.s` in a cljs file made it fail and the rest of this patch is the fix.

lib/src/clojure_lsp/queries.clj
```diff
@@ -510,34 +510,42 @@
                      (= (:name-end-col %) (:name-end-col keyword-element))))
        first))
 
-(defn find-all-ns-definition-names-xf []
+(def find-all-ns-definitions-xf
```
ericdallo, line 513: 👍🏻

ericdallo (approved):

Issue #990, opened 2022-05-20
Lazy analysis


mainej:
Inspired by [this article](https://rust-analyzer.github.io/blog/2020/07/20/three-architectures-for-responsive-ide.html), I'm researching a new approach to analyzing and linting. I've been calling this approach "lazy analysis".

Lazy analysis, which I'll explain in a moment, has the potential to reduce the amount of analysis that clojure-lsp needs to do, and to limit the number of elements that must be inspected when querying the analysis. I believe this will improve performance at startup and during use, in projects large and small.

The core insight is that when a user is looking at a file, **most clojure-lsp features only need analysis from that file and its immediate neighbors**.

But that's it! Anything further away—transitive dependencies and dependents—won't affect these features.

To be more precise, features that need var _definitions_ (like go-to definition, view docstring, etc.) need to look only in the current namespace and its direct dependencies. Features that need var _usages_ (like show references, rename, incoming call hierarchy, etc) need only the current namespace and its direct dependents.

See the table below for a more complete picture of which features need which files. See the [Cons](#cons) section for features that need broader analysis.

| Features                         | self | dependencies: var def | dependents: var usage |
|----------------------------------|------|-----------------------|-----------------------|
| call hierarchy: incoming         | x    |                       | x                     |
| call hierarchy: outgoing         | x    | x                     |                       |
| code lens references             | x    |                       | x                     |
| completion                       | x    | x                     |                       |
| diagnostics: general             | x    |                       |                       |
| diagnostics: unresolved var      | x    | x                     |                       |
| diagnostics: unused public var   | x    |                       | x                     |
| docs: docstring                  | x    | x                     |                       |
| docs: signature                  | x    | x                     |                       |
| format                           | x    |                       |                       |
| goto: declaration                | x    |                       |                       |
| goto: definition                 | x    | x                     |                       |
| goto: implementations            | x    |                       | x                     |
| goto: references                 | x    |                       | x                     |
| refactor: coll: change-coll      | x    |                       |                       |
| refactor: coll: cycle-coll       | x    |                       |                       |
| refactor: create-test            | x    |                       | ? (test ns)           |
| refactor: cycle-privacy          | x    |                       |                       |
| refactor: drag: drag-backward    | x    |                       |                       |
| refactor: drag: drag-forward     | x    |                       |                       |
| refactor: fn: create-function    | x    |                       |                       |
| refactor: fn: demote-fn          | x    |                       |                       |
[...]

PR #1040, opened 2022-06-10
refactor: Query signatures

mainej:
This PR refactors the `queries` namespace so that its functions accept a `db` rather than `analysis`. In namespaces that do their own analysis filtering, it brings the db closer to the filters.

This change feels somewhat unmotivated, but it's the first step to storing more metadata in the db. With additional metadata, for example about the dependency graph [ref](https://github.com/clojure-lsp/clojure-lsp/issues/990), we'll be able to make the analysis filters more efficient, but only if we have access to the full db. These changes also hide a few small details, such as that queries that assist `clean-ns` actually rely more on diagnostics than analysis.

I discussed this PR with @ericdallo before submitting it. We agreed it would help minimize noise in the upcoming dep-graph-based queries PR.

### DIFF
--- a/lib/src/clojure_lsp/feature/completion.clj
+++ b/lib/src/clojure_lsp/feature/completion.clj
@@ -492,23 +492,18 @@
                                           :name (symbol name)
                                           :to (symbol ns)
                                           :bucket :var-usages})]
-    (if definition
-      (-> item
-          (assoc :documentation (f.hover/hover-documentation definition db*)))
-      item)))
+    (cond-> item
+      definition (assoc :documentation (f.hover/hover-documentation definition db*)))))
 
 (defn ^:private resolve-item-by-definition
   [{{:keys [name filename name-row name-col]} :data :as item} db*]
   (let [db @db*
-        local-analysis (get-in db [:analysis filename])
-        definition (q/find-first #(and (identical? :var-definitions (:bucket %))
-                                       (= name (str (:name %)))
-                                       (= name-row (:name-row %))
-                                       (= name-col (:name-col %))) local-analysis)]
-    (if definition
-      (-> item
-          (assoc :documentation (f.hover/hover-documentation definition db*)))
-      item)))
+        element (q/find-element-under-cursor db filename name-row name-col)
+        definition (when (and (identical? :var-definitions (:bucket element))
+                              (= name (str (:name element))))
+                     element)]
+    (cond-> item
+      definition (assoc :documentation (f.hover/hover-documentation definition db*)))))
 
 (defn resolve-item [{{:keys [ns]} :data :as item} db*]
   (let [item (shared/assoc-some item
--- a/lib/src/clojure_lsp/feature/completion.clj
+++ b/lib/src/clojure_lsp/feature/completion.clj
@@ -411,7 +411,7 @@
             current-ns-elements (get analysis filename)
             support-snippets? (get-in db [:client-capabilities :text-document :completion :completion-item :snippet-support] false)
             all-other-ns-elements (mapcat val (dissoc analysis filename))
-            cursor-element (q/find-element-under-cursor analysis filename row col)
+            cursor-element (q/find-element-under-cursor db filename row col)
             cursor-value (if (= :vector (z/tag cursor-loc))
                            ""
                            (if (z/sexpr-able? cursor-loc)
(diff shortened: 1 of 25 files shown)


lib/src/clojure_lsp/feature/completion.clj
```diff
@@ -407,11 +411,10 @@
       []
       (let [filename (shared/uri->filename uri)
             settings (settings/all db)
-            analysis (:analysis db)
-            current-ns-elements (get analysis filename)
+            current-ns-elements (get-in db [:analysis filename])
```
ericdallo, line 414: nice

lib/src/clojure_lsp/feature/file_management.clj
```diff
@@ -80,12 +80,15 @@
...
+(defn reference-filenames [filename db-before db-after]
+  (let [old-local-analysis (get-in db-before [:analysis filename])
+        new-local-analysis (get-in db-after [:analysis filename])
+        changed-var-definitions (find-changed-var-definitions old-local-analysis new-local-analysis)
         changed-var-usages (find-changed-var-usages old-local-analysis new-local-analysis)
-        project-analysis (into {}
-                               q/filter-project-analysis-xf
-                               (dissoc (:analysis db) filename)) ;; don't notify self
+        project-db (-> db-after
+                       q/db-with-project-analysis
```
ericdallo, line 89: the need to use this function outside query.clj maybe tells that we should move the query used on this function to inside query, WDYT?

lib/src/clojure_lsp/queries.clj
```diff
@@ -15,14 +15,23 @@
...
 
+(defn ^:private project-analysis [{:keys [analysis]}]
+  (medley/remove-vals
+    (fn [elems]
+      (:external? (first elems)))
+    analysis))
+
+(defn db-with-project-analysis [db]
+  (assoc db :analysis (project-analysis db)))
+
```
ericdallo, line 34: I think this is the only change that concerns me a little bit as we are creating a new concept of a db but where the analysis are not "full", and relies only on calling it as a `project-db`, I don't think it's a big deal, but maybe avoiding using it outside this namespace may reduce this concern? kind of related with https://github.com/clojure-lsp/clojure-lsp/pull/1040/files?diff=unified&w=1#r894925391

ericdallo (approved):
Looks good!

lib/src/clojure_lsp/queries.clj
mainej: I see what you mean @ericdallo, but I also think it's very similar to what we were doing before—filtering the analysis and passing that around, relying only calling it `project-analysis`.

I understand what you're saying about moving the usages into `queries`, but that'd be tricky in several situations. For example, consider custom-linting. We want to do all of our custom linting based on project files, excluding external files. That algorithm needs to run 4 queries on the analysis: to find the used vars and keywords, and then to find unused vars and keywords. We could move each of those queries into the `queries` namespace, but then we'd be filtering to the project analysis 4 times instead of once. Filtering to project analysis is much faster than it used to be, but still, I'd rather do it as few times as possible.

It's a similar situation for `file-management/find-references`, one of the other places that uses `queries/db-with-project-analysis`. We run a couple of queries on the project-db, and it'd be nice to filter only once. The final usage is in `internal-api/rename!*`. That one could pretty easily be moved into `queries`. I'll work on that.

But in general, this is hard to get away from. There are other places where we used to filter the analysis to a single file or a few files, and pass that around calling it `local-analysis` or something like that. This PR is doing the same thing, but with a filtered version of the db.

Anyway, you have a point—this is sort of a code smell, where perhaps we're exposing more than we want to. I'll keep that in mind as I work on dep-graph queries, which will include changes to many of the same locations. If I can, as I'm working on that, I'll move things into `queries`.

mainej: One other thought... I think this code is also a little surprising because it introduces a new way to filter to project analysis, something we've been doing with `filter-project-analysis-xf` so far. I plan to revisit that eventually, and hopefully I'll be able to get back to having just one way of doing things. But for awhile the project will be in a weird state, where we have older parts of `queries` mixed with newer parts. 

I need to keep both parts in the code at the same time so that we can have a flag to toggle dep-graph queries. But when the flag is removed, I'll plan to clean things up.

ericdallo: > filtering the analysis and passing that around, relying only calling it project-analysis.

Good point

> Anyway, you have a point—this is sort of a code smell

Yeah, exactly, this is not a new problem, it's something we are already doing in multiple places as you mentioned, I think it's good to keep that in mind during those refactors, for now I guess it's ok

PR #1045, opened 2022-06-11
Improve hover element selected

### LINKED ISSUE #995: function signatures in the Emacs minibuffer do not reflect cursor position or are misleading
**Describe the bug**
There is something odd with what is displayed in the Emacs minibuffer regarding to cursor position. Taking a Cider as a reference there should be (mostly) a current function signature.

Discussion on Slack: https://app.slack.com/client/T03RZGPFR/C099W16KZ

**To Reproduce**

- commented out LSP configuration in my [init.el](https://gist.github.com/genmeblog/fbec9d790ffd46aafdbf0cef7c69471d) (lines 315-327) open any clj file and call `cider-jack-in-clj`

Comments below describe minibuffer content:

```clojure
(map|) ;; map doc
(map |) ;; map doc
(map identity|) ;; map doc with `f` argument highlighted
(map identity |) ;; map doc with `coll` argument highlighted
(map identity (repeatedly|)) ;; repeatedly doc
(map identity (repeatedly 10|)) ;; repeatedly doc, first arg highlighted
(map identity (repeatedly 10 rand|)) ;; repeatedly doc, second arg highlighted
(map identity (repeatedly 10 rand)|) ;; map doc, second arg highlighted
(map identity|) ;; map doc with `f` argument highlighted (need to move a cursor left/right to have doc visible)
(map|) ;; as above
(|) ;; nothing
```

- LSP turned on

```clojure
(map|) ;; map doc
(map |) ;; map doc
(map identity|) ;; identity doc (?)
(map identity |) ;; identity doc (?)
(map identity (repeatedly|)) ;; repeatedly doc
(map identity (repeatedly 10|)) ;; repeatedly doc, no highlight
(map identity (repeatedly 10 rand|)) ;; rand doc (?)
(map identity (repeatedly 10 rand)|) ;; still rand doc (?)
(map identity|) ;; identity doc (?)
(map|) ;; map doc
(|) ;; map doc (?)
```

Additionally, when you repeat above a couple of times, below things happen sometimes:
- cider's doc appears, replaced by lsp (or not)
- corrected typo (eg. writing `identy` and corrected to `identity`) displays only symbol before correction (eg. `identy`)
- `map` -> `mapv` -> `map`, `mapv` docs persist
- when moving a cursor around existing code, docs is shown or not, eg:
    - `(let| [...])` - no `let` docs
	- `(let |[...])` - `let` docs is displayed
- other things like `(/ var 2.|7)` - `var` is displayed, depending what was visited by cursor

Seems like some caching issue, which is really visible after some time of coding and when heavy (and fast) editing is performed.

**Expected behavior**
Cider behavior as much as possible, ie. the doc context should be current sexp not current symbol

 - OS: Ubuntu 22.04 on WSL2 (Windows 11)
 - Editor: Emacs 29.0.50 with native compilation (custom build), newest Cider
 - Version: clojure-lsp 2022.05.23-13.18.11, clj-kondo 2022.04.26-SNAPSHOT
 - init.el: https://gist.github.com/genmeblog/fbec9d790ffd46aafdbf0cef7c69471d

ericdallo:
Fixes #995

This is an attempt to improve the element selected during hover, the current behavior select any valid analysis element found on the current row/col, if no one is found, it tries to find a valid element until the start of the row decreasing the col.
This works for  "most" cases, but it fails for some lisp cases where we want to have the hover information about the current function being called (Check the issue for more details).

### Impl

The approach done on this PR is to:
- Parse the code with rewrite-clj
- find the function being called (op) (same algh used on signature-function feature)
- use its location meta to find the element
- if something is not found/is nil, fallback to current algh.

### Concerns

1. The hover feature usually is quite fast as it only query analysis, now we are introducing the parse of a file with rewrite-clj (should we think about caching the text/version of rewrite-clj nodes?)

2. do we always want to show the current function being called on hover and not the current cursor element? I think so, but not sure I am missing some corner case

--- 

~- [ ] I updated documentation if applicable (`docs` folder)~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
   - Fix to avoid suggesting an alias from a clj file to a cljs file. #1024
   - Find references of namespace usages now find all namespace usages on project, not only the definition. #1022
   - Fix drag from quoted symbols and other special nodes #969
+  - Improve element selected on `textDocument/hover`, showing the function being called instead of the closest element found backwards. #995
 
 ## 2022.05.31-17.35.50
 
--- a/lib/src/clojure_lsp/feature/hover.clj
+++ b/lib/src/clojure_lsp/feature/hover.clj
@@ -1,10 +1,14 @@
 (ns clojure-lsp.feature.hover
   (:require
    [clojure-lsp.feature.clojuredocs :as f.clojuredocs]
+   [clojure-lsp.feature.file-management :as f.file-management]
+   [clojure-lsp.parser :as parser]
    [clojure-lsp.queries :as q]
+   [clojure-lsp.refactor.edit :as edit]
    [clojure-lsp.settings :as settings]
    [clojure-lsp.shared :as shared]
-   [clojure.string :as string]))
+   [clojure.string :as string]
+   [rewrite-clj.zip :as z]))
 
 (set! *warn-on-reflection* true)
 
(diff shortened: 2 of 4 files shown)


ericdallo:
@mainej It'd be nice to hear your thoughts on this one :)

mainej:
@ericdallo first, nice! This approach makes sense. It's always been weird that hover disappears exactly when you need it, right as you start to enter arguments.

As for performance, I see what you mean. It means re-parsing with every key stroke. We have a similar problem with completion.

> should we think about caching ... rewrite-clj nodes

That's one option. After the dep-graph branch is merged (assuming it is) the db will have a new key called `:file-meta`, which is updated after every analysis. If the file is open we could parse a rewrite-clj tree and cache it in that part of the db. Then everything else (completion, hover, all the refactors, etc.) could use that parse. That'd be a nice global optimization, at the expense of adding 40ms-100ms (the time to parse) to every analysis.

Another option, which might be easier in the short term, is to debounce the hover invocations. That'd limit the number of extra parses.

ericdallo:
Yeah, I like the file-meta containing the rewrite-clj node caching, sounds like a good improvement for the future, for now I think it's ok to keep this way and see how it goes

PR #1046, opened 2022-06-13
Update to latest clj-nix

jlesquembre:
In latest clj-nix, to add better support for maven snapshot version, I had to update the lock file format. Unfortunately, next time you run the update deps-lock workflow, the lock file and the clj-nix version will get out of sync. This PR solves that.
 
Also pins clj-nix to a specific version to avoid that problem in the future


ericdallo (approved):
Thank you!

PR #1050, opened 2022-06-14
Waits for client to do rename before moving db data

### LINKED ISSUE #1049: rename mangles db
**Describe the bug**
When sending the client a namespace rename, the server [preemptively modifies the db](https://github.com/clojure-lsp/clojure-lsp/blob/e92919647521dd756eb82b1760e9bb44d0054ed1/lib/src/clojure_lsp/feature/rename.clj#L228-L230) to move the old file's data to the new file. But, it does this naively. The document is moved, but its contents aren't updated, which could break future rewrite-clj parses. The analysis is moved, but the elements aren't updated with the new ns or filename, which will break future queries. (It also doesn't move findings, but that's a separate bug.)

Doing a preemptive modification is arguably incorrect anyway. The server is telling the client that it _should_ do a rename, but we don't have a guarantee that the client _will_ do the rename. I don't think there's anything that says the client can't ignore the edits, for example to let the user abort.

When a client actually does a rename, it's [supposed to](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didRename) do a `did-close` on the old file and a `did-open` on the new file. That should be more reliable. The `did-close` should remove the old file from the db, and the `did-open` should save the correct document text and trigger a fresh analysis.

So, I propose we remove the preemptive db modification.

mainej:
Instead of preemptively moving analysis and document within db, wait for the client to actually do the rename. When it does, it [should](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didRename) call `did-close` on the old file and `did-open` on the new file, which will remove the old file and add the document text and analysis for the new file.

Fixes #1049

- [x] I created an issue to discuss the problem I am trying to solve or an open issue already exists. #1049

### DIFF
--- a/lib/src/clojure_lsp/feature/rename.clj
+++ b/lib/src/clojure_lsp/feature/rename.clj
@@ -3,7 +3,6 @@
    [clojure-lsp.queries :as q]
    [clojure-lsp.settings :as settings]
    [clojure-lsp.shared :as shared]
-   [clojure.set :as set]
    [clojure.string :as string]))
 
 (set! *warn-on-reflection* true)
@@ -204,9 +203,8 @@
           result
           (shared/->range element))))))
 
-(defn rename-element [uri new-name db* filename element source]
-  (let [db @db*
-        references (q/find-references db element true)
+(defn rename-element [uri new-name db element source]
+  (let [references (q/find-references db element true)
         definition (q/find-definition db element)
         source-paths (settings/get db [:source-paths])
         client-capabilities (:client-capabilities db)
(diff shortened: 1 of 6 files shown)


lib/src/clojure_lsp/feature/rename.clj
```diff
@@ -225,9 +223,6 @@
                  (not (identical? :namespace-alias (:bucket element)))
                  (not= :rename-file source))
           (let [new-uri (shared/namespace->uri replacement source-paths (:filename definition) db)]
-            (swap! db* (fn [db] (-> db
-                                    (update :documents #(set/rename-keys % {filename (shared/uri->filename new-uri)}))
-                                    (update :analysis #(set/rename-keys % {filename (shared/uri->filename new-uri)})))))
```
ericdallo, line 230: I know there was some issue with ns renaming when you do it multiple times like: a -> b then b -> a and then a -> b, probably related with this, even so not sure why I added that swap 😅 anyway, this looks correct

mainej: You mean this was originally a fix for multiple renames, or you suspect it was the source of a problem with multiple renames?

One thing I noticed since writing this PR, is that the rename for `:documents` was actually never working because `:documents` is keyed by uri, not filename.

>  this looks correct

👍 I'll fix conflicts and merge.

ericdallo: Yeah, I meant that swap was probably causing that issue when renaming the same file multiple times, it should be fixed now

Issue #1023, opened 2022-06-08
bug: create public function doesn't consider unaliased namespace-usages


mainej:
**Describe the bug**

```clojure
(ns a
  (:require [clojure-lsp.shared :as shared))

|shared/foo
|clojure-lsp.shared/foo
```
The action menu offers to create the `foo` function in `clojure-lsp.shared` for the first cursor, which is correct. But for the second cursor it offers create both the namespace `clojure-lsp.shared` and the function `foo`, which is incorrect. 

To fix this, create-function should be expanded to recognize not just aliases but also the full names of required namespaces.

Issue #1020, opened 2022-06-08
bug: `find-definition` from a var defined by `potemkin/import-vars` goes to any definition with the same name


mainej:
**Describe the bug**

If you have `(import-vars |my-ns/q)`, calling find-definition could take you to any definition of `q`, in any ns, not just in `my-ns`.

It should go to the var from the imported ns. To fix this we should use `:imported-ns` and `:imported-var` to convert the var-definition into a var-usage, and call `find-definition :var-usages`.

mainej:
> quite specific

To say the least... this is a really rare use case. Still, pretty easy to fix now that imported-ns is available.

Issue #1018, opened 2022-06-08
bug: `file-management/reference-filenames` neglects keywords


mainej:
**Describe the bug**

`file-management/reference-filenames` doesn't try to detect changes to the usages of registered keywords, and therefore doesn't notify the files where they are registered. Therefore, their unused-public-var lint gets out of date.

Fixing this could cause a serious performance degradation, since so many changes add or remove keywords, and keywords can be registered in any namespace. However, see #1019.

mainej:
> making clojure-lsp analysis differ from kondo a little bit more

That's a valid concern... it's hard to keep track of how our data differs from the [clj-kondo documentation](https://github.com/clj-kondo/clj-kondo/tree/master/analysis). What I find particularly difficult to remember is which keys we've renamed. I assume those renames make  `shared/inside?` and other similar tools a little more generic. But if it were possible to do less of that, it'd be easier to read the clj-kondo docs and trust that it applied to our elements too.

Moving elements around to different buckets is also a bit confusing, but I've memorized that `:namespace-alias` is our custom bucket, and is mostly a copy of `:namespace-usages`. (I've actually thought about proposing that we get rid of `:namespace-alias`, and instead re-calculate whether a `:namepsace-usage` is an alias and what the alias's range is whenever we need it.)

Anyway, I agree, separating keyword definitions will be another thing to remember. :/

ericdallo:
Yes, I didn't even remember that `:namespace-alias` was a custom bucket :)

PR #1053, opened 2022-06-15
performance: Maintain and use a dependency graph

mainej:
Maintains a dependency graph and file meta-data to be able to more efficiently filter the clj-kondo analysis.

The dep-graph is a map from namespaces to the namespaces they depend on (their dependencies) and the namespaces which depend on them (their dependents). Along with file meta-data, which records which namespaces a file defines, this makes it easy to quickly filter to a much narrower part of the analysis. For example, when finding references of a var, you need to look only in the analysis of the var's namespace's dependent's files.

Since so many of the queries inspected every element of every file in the whole analysis, performance was tied to the size of the project. Now, in many cases performance is tied instead to the combined sizes of a file and its dependencies or dependents.

This makes many parts of the code faster, but is most noticeable in a few places.

In metabase, a very large project, find-references typically took 100-120ms, which was the time to look at every var usage. The time hasn't improved when finding references to widely used vars like `defn` because most namespaces depend on clojure.core, so we still have to search through all the var usages. But for namespaces that have fewer dependents, find-references is much faster, often around 20ms and sometimes less than 10ms.

Similarly, find-definition went from 60ms to 20ms.

But find-references and find-definition are both invoked by the user relatively rarely. The big payoff is in completion, which is invoked regularly, sometimes on every keystroke.

In metabase, before:

```
completing 'comp' "Elapsed time: 330.945051 msecs"
completing 'config/e' "Elapsed time: 251.061651 msecs"
completing '::lmf/d' "Elapsed time: 282.454571 msecs"
```

After:

```
completing 'comp' "Elapsed time: 23.971551 msecs"
completing 'config/e' "Elapsed time: 13.562191 msecs"
completing '::lmf/d' "Elapsed time: 96.987585 msecs"
```

These are just using `time`, not something more rigorous, but the effect is still apparent. In a medium size project, clojure-lsp, before:

```
completing 'comp' "Elapsed time: 105.267534 msecs"
completing 'shared/f' "Elapsed time: 94.15348 msecs"
completing '::coercer/c' "Elapsed time: 103.5377 msecs"
```

After:

```
completing 'comp' "Elapsed time: 8.83796 msecs"
completing 'shared/f' "Elapsed time: 9.592488 msecs"
completing '::coercer/c' "Elapsed time: 32.093406 msecs"
```

The biggest improvement in completion comes from minimizing time in `with-elements-from-alias` and `cursor-full-ns?`, finding aliases and namespace names, respectively. Previously these were a major contributor to completion time. Reducing time in these speeds up even things which can't use the dep graph, like keyword completion.

Notes on automated tests
========

You can run the test suites with dep-graph-queries turned on by un-commenting the `:dep-graph-queries` setting in `common-test/src/clojure_lsp/test_helper.clj` and `cli/integration-test/sample-test/.lsp/config.edn` for unit and integration tests respectively.

A few lower-level unit tests about analysis-narrowing return different results depending on whether `:dep-graph-queries` is on. This is expected, and proves that the feature is working. But otherwise this optimization should be transparent to the rest of the code, so other tests haven't changed.

It would be sort of nice to have a "test matrix", so that the full suite was run twice, once with `:dep-graph-queries` on and once with it off. But assuming that's a pain to set up, I'm OK with running them both manually until this experiment ends.

Notes for beta testers
========

With a build from this branch, you can test this feature by adding `:experimental {:dep-graph-queries true}` to your `.lsp/config.edn`. (If you work on clojure-lsp itself, that setting exists but is commented out.)

You can toggle the feature on and off by changing the setting, without restarting the server. (You have to wait at least 1 second for the new settings to be picked up. Sometimes this seems flaky.) This is useful to compare the results of certain actions and to compare performance.

Effects on all users
========

This patch will have some effect on all users, even if they aren't using the experimental feature.

In order for the dep-graph to keep track of the relationships between files in external dependencies (i.e. those outside of the project) it needs the clj-kondo analysis to include `:namespace-usages`. There isn't a way to turn off that part of clj-kondo so we've always received that when running the external analysis, but we used to dissoc it before putting it in the db and before caching it on disk. Now we need to keep that analysis around, including in the disk cache. For this reason, the db/version has been bumped, meaning that the external dependencies of all users' projects will be re-scanned, and that the size of cache files will be somewhat larger. (I haven't measured the difference in file size.)

Additionally, the dep-graph is built and maintained for all users, regardless of whether they're using its data. In metabase this adds about 1.5s to startup. In clojure-lsp it adds about 500ms. Ongoing maintenance of the dep-graph after startup is negligible. This approach means anyone can toggle the feature on without restarting the server (see above "Notes for beta testers"). But if the additional startup time isn't acceptable, this approach can be re-thought.

Notes for PR reviewers
===
* When looking at the diffs, I suggest hiding whitespace changes. 
* I'd love to have testers for this branch before its merged. After its merged, assuming it is, I hope we can appeal to the community for additional testing.
* And if you really want to stress test the feature, disable your editors' completion caching 😈 .

- [x] This performance improvement was discussed in #990


lib/src/clojure_lsp/crawler.clj
```diff
@@ -52,9 +52,7 @@
                                       (producer/publish-progress producer percentage "Analyzing external classpath" progress-token)))
             normalization-config {:external? true
                                   :filter-analysis (fn [analysis]
-                                                     (-> analysis
-                                                         (dissoc :namespace-usages)
```
mainej, line 56: This is the change that necessitates to bump to db/version.

lib/src/clojure_lsp/kondo.clj
```diff
@@ -54,14 +55,18 @@
...
 (defn db-with-analysis
   "Update `db` with normalized kondo analysis."
-  [db analysis]
-  (update db :analysis merge analysis))
+  [db {:keys [analysis external?]}]
+  (-> db
+      (update :analysis merge analysis)
+      (dep-graph/refresh-analysis (select-keys (:analysis db) (keys analysis))
+                                  analysis
+                                  (not external?))))
```
mainej, line 63: Whenever the analysis is updated, we refresh the dep graph. This is the only place this is done, so to keep the dep graph in sync with the analysis, everything that puts analysis in the db must either call this function or `db-with-results`, which calls this function.

lib/src/clojure_lsp/kondo.clj
```diff
@@ -296,15 +306,29 @@
       1 (run-kondo-on-paths! paths db* normalization-config (partial file-analyzed-fn 1 1))
       (->> batches
            (map (fn [{:keys [index paths]}]
-                  (logger/info "Analyzing" (str index "/" batch-count) "batch paths with clj-kondo...")
+                  (logger/info "Analyzing batch" (str index "/" batch-count))
                   (run-kondo-on-paths! paths db* normalization-config (partial file-analyzed-fn index batch-count))))
            (reduce shared/deep-merge)))))
 
 (defn run-kondo-on-reference-filenames! [filenames db*]
+  ;; TODO: We process the dependent and dependency files together, but we could
```
mainej, line 314: See #1027 and #1028

lib/src/clojure_lsp/refactor/transform.clj
```diff
@@ -774,17 +774,24 @@
...
-          (when-not (:external? ns-def)
-            {:ns (:name ns-def)
-             :name z-name})
-          (when-not (:external? ns-usage)
-            {:new-ns (:name ns-usage)
-             :name z-name}))
-        {:new-ns z-ns
-         :name z-name}))))
+          z-ns (namespace z-sexpr)
+          ;; TODO: shouldn't this also look for unaliased ns-usages?
```
mainej, line 778: See #1023

lib/src/clojure_lsp/refactor/transform.clj
```diff
@@ -798,40 +805,42 @@
 
 (defn ^:private create-function-for-alias
   [local-zloc ns-or-alias fn-name defn-edit uri db]
-  (let [ns-usage (q/find-namespace-usage-by-alias db (shared/uri->filename uri) (symbol ns-or-alias))
+  (let [filename (shared/uri->filename uri)
+        ;; TODO: shouldn't this also look for unaliased ns-usages?
```
mainej, line 809: See #1023

lib/src/clojure_lsp/queries.clj
```diff
@@ -175,18 +364,23 @@
...
                 (:reg %)
                 (= (:ns %) (:ns element)))
-          (:analysis db)))
+          db))
       element))
 
 (defmethod find-definition :var-definitions
   [db element]
   (if (= 'potemkin/import-vars (:defined-by element))
+    ;; FIXME: this is buggy... it goes to **any** definition with the same name,
```
mainej, line 373: See #1020

lib/src/clojure_lsp/feature/file_management.clj
```diff
@@ -67,59 +68,66 @@
...
-  [old-local-analysis new-local-analysis]
-  (let [old-var-usages (filter #(identical? :var-usages (:bucket %)) old-local-analysis)
-        new-var-usages (filter #(identical? :var-usages (:bucket %)) new-local-analysis)
+  [old-local-elements new-local-elements]
+  (let [old-var-usages (filter #(identical? :var-usages (:bucket %)) old-local-elements)
+        new-var-usages (filter #(identical? :var-usages (:bucket %)) new-local-elements)
         usage-signature (juxt :to :name)]
     (find-changed-elems-by usage-signature old-var-usages new-var-usages)))
 
+;; TODO: this neglects to detect changes to the usages of registered keywords,
```
mainej, line 84: See #1018

lib/src/clojure_lsp/refactor/transform.clj
ericdallo: Shouldn't we mention the issue on the comment to make it easier to understand the comment?

ericdallo (commented):
Thanks for this! I still need to review the code carefully but I'll already quote some comments here

> It would be sort of nice to have a "test matrix", so that the full suite was run twice, once with :dep-graph-queries on and once with it off. But assuming that's a pain to set up, I'm OK with running them both manually until this experiment ends.

We could probably set that up easily with a ENV var which we could check on test runner, sounds like a good move.

>  (If you work on clojure-lsp itself, that setting exists but is commented out.)

Sounds good for now, but we should leave it on as soon as we intend to rollout to clients

>  For this reason, the db/version has been bumped

I think it's ok, but we don't necessarily need to bump the version right now since no one except us want to enable this feature, right? we could bump it when we think about releasing to clients, have a outdated analysis needing to clear cache manually seems quite expected for experimental features/beta users. no hard feelings on this too, if you think consistency is even more important we can bump it.

> Additionally, the dep-graph is built and maintained for all users, regardless of whether they're using its data. In metabase this adds about 1.5s to startup. In clojure-lsp it adds about 500ms. 

Similar to the previous quote, we could check if the setting is enabled and don't calculate the dep-graph on start, I know we would lose the hot-reload setting on dep-graph, but it seems to be expected as an experimental feature and since it's related to performance improvements for all other users, I like the idea to add that check, WDYT?

> I'd love to have testers for this branch before its merged. After its merged, assuming it is, I hope we can appeal to the community for additional testing.

Sure, I can start testing it right now for some days.

lib/src/clojure_lsp/crawler.clj
ericdallo: Just a note for ourselves related to what was mentioned [here](https://github.com/clojure-lsp/clojure-lsp/pull/1053#pullrequestreview-1008373323):
if we agree on deploying a code that won't affect existing users, we could check the setting here as well to know if we should dissoc

lib/src/clojure_lsp/kondo.clj
ericdallo: I like it how it doesn't have side effects and just return the db with the dep-graph built-in

lib/src/clojure_lsp/crawler.clj
snoe: I'd vote for consistency, I could see namespace-usages being useful beyond the dep graph.

lib/src/clojure_lsp/refactor/transform.clj
mainej: Yeah, sorry... I had meant to do that but I wrote all these comments before writing the issues, and forgot to go back and add the cross references.

mainej:
Thanks for the feedback @ericdallo and @snoe! I'm busy with some other stuff, but later I'll take some time to process it all.

I wanted to give a quick update though. Originally I reported that find-definition went from 60ms to 20ms, but that seemed surprisingly long. I also noticed that find-definition was dominating the time in certain completion scenarios. I tracked down the problem and now find-definition is essentially instantaneous, 1ms in many cases.

Also, in the initial PR report I neglected a particular completion which is still slow. In metabase, if you type `met` you get almost 600 completions. Almost all of them are suggestions to complete to a namespace (there are a lot of namespaces that start with `met`). That takes 350-400ms. Looking at a flamegraph, 95% of that time is in `completion/with-elements-from-alias`, either in `feature.add-missing-libspec/add-known-alias` or in `refactor.transform/result`. We're calculating not just the completion, but also the edit that would add the namespace to the `:require`s. I think we could postpone that edit calculation to `completion/resolve-item`, so that we calculate only one edit, and only after the user chooses it. I think that'd be a nice optimization, but would rather postpone it to a separate PR.

lib/src/clojure_lsp/refactor/transform.clj
mainej: Cross-references added

lib/src/clojure_lsp/kondo.clj
mainej: I prototyped the dep-graph maintenance algorithm at the REPL and making it pure made it easier to verify. When I switched to unit tests I could have kept the purity, generating and passing analysis into this function, and checking the results. But I opted for the impure `h/load-code-and-locs` and `f.file-management/did-close` to simulate opening, changing and closing files. I think that's better because it tests that everything integrates with the dep-graph. I guess either way could have worked. Anyway, I'm glad that the core of this is pure... I think Rich's lessons might finally be sinking in. :)

mainej:
@snoe, I agree, performance tests—especially performance tests that integrate with real projects—would be handy. They'd be useful for developing this kind of feature but also for noticing regressions caused by non-performance related features. We've been collecting ideas in #909 for awhile. Would you mind adding your thoughts there about checking out projects at known versions? I really like that idea. I keep meaning to work on #909, but then get distracted by actually fixing performance problems :)

@ericdallo, I'll look at creating test matrices with an env var. For some reason the GitHub actions daunt me, so if this would be easy for you, I'd welcome help.

Also, you've figured out exactly what would be needed to limit the impact on all users: 
1. Figure out how to avoid the db version bump and/or conditionally dissoc the :namespace-usages.
2. Only conditionally run `dep-graph/refresh-analysis`.

Do you want me to look into that more, or leave it like it is?

ericdallo:
I think we can leave it like this for now, we are aware of what will be affected and we intend to take that risk in change of consistency.

I can help with the test matrix, do a PR on this one or even add commits directly to this branch.

I was thinking in release a new clojure-lsp version on Monday, do you think makes sense leave this PR to not this one but next release? I still need to review code and test it manually too

mainej:
> I can help with the test matrix, do a PR on this one or even add commits directly to this branch.

I'd appreciate that. Commits on this branch would be fine. I'll avoid force-pushing.

> I was thinking in release a new clojure-lsp version on Monday, do you think makes sense leave this PR to not this one but next release? I still need to review code and test it manually too

Yeah, probably best to postpone it to the next release cycle. No need to rush it through.

ericdallo:
@mainej last commit should do the trick, LMK what you think

mainej:
@ericdallo that's simpler than I was envisioning. I thought you'd use GitHub Action [env vars](https://docs.github.com/en/actions/learn-github-actions/environment-variables) and [test matrices](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs). But this way is nice and simple, and will get around messing with the `config.edn` files or the initial value of `:settings`. I kind of think we should run the CLI and integration tests with `CLOJURE_LSP_USE_DEP_GRAPH` too. WDYT?

Also... this is confusing... [This test](https://github.com/clojure-lsp/clojure-lsp/runs/6957662332?check_suite_focus=true) had an ERROR when running with the dep-graph, but the suite passed. Any idea why?

mainej:
> queries should affect lib only

Thanks for adding the CLI test... Queries are used in the CLI too, by `rename` and `clean-ns`.

> your link points just to a succeeded build

True, the build succeeded, but it shouldn't have. Search the test output for "ERROR".

mainej:
Ah, OK, thanks for figuring that out! That explains why the test had an error. But why did the build pass anyway?

ericdallo:
because it's a exception in a async go loop, the test doesn't catch this kind of exception

mainej:
So kaocha notices that an exception was thrown but doesn't fail the test?

ericdallo:
> So kaocha notices that an exception was thrown but doesn't fail the test?

Yeah, I think so, because it's a print to stdout, not really a `(is)` fail or a exception on main test thread

lib/src/clojure_lsp/db.clj
```diff
@@ -12,14 +12,16 @@
 (def ^:private db-logger-tag "[DB]")
 
 (def initial-db {:documents {}
-                 :processing-changes #{}})
+                 :processing-changes #{}
+                 :dep-graph {}
+                 :file-meta {}})
```
ericdallo, line 17: I remember when first time checking clojure-lsp one of the things that made me took some time to understand was what was inside the atom...
This change reminded me of that, maybe we could improve this code in the future making that more clear like with a schema or anything that would make that more clear.

lib/src/clojure_lsp/dep_graph.clj
```diff
@@ -0,0 +1,293 @@
...
+;; {(':dependencies' #<ns*>)?
+;;  (':dependents' #<ns*>)?
+;;  (':aliases' #<alias*>)?
+;;  (':files' #{filename*})?
+;;  (':internal?' boolean)?
+;;  (':from-internal?' boolean)?
+;;  (':from-langs' #{lang*})?}
+;; filename = string
+;; alias = (symbol | nil)
+;; lang = ':clj' | ':cljs' | ':edn'
```
ericdallo, line 37: maybe bikeshedding, but it took me a little bit to understand this is describing the db from top-bottom, being top a higher level, maybe just invert to bottom-top?  describing lang first and then the things that use it until describing the db? maybe it's personal too

mainej: I was sort of following the style of the Datomic [grammar](https://docs.datomic.com/cloud/query/query-data-reference.html#arg-grammar). But I've always found those docs a little hard to follow, partially because of your observation—the definitions are used before they're made. That's not how we do it in Clojure! :)

I'll take a stab at restructuring this. While I'm at it, do you want me to move it to `db/initial-db`? I kind of like it where it is, close to the code that needs to reference it most often, but I'd be OK with either location.

lib/src/clojure_lsp/db.clj
mainej: A schema would be nice, or even just a written description.

lib/src/clojure_lsp/dep_graph.clj
ericdallo: I'm ok with either too, let's keep it here for now and see how it goes

mainej: Grammar re-ordered; example added.

lib/src/clojure_lsp/dep_graph.clj
```diff
@@ -0,0 +1,347 @@
...
+                                        clojure-lsp.kondo        1
+                                        clojure-lsp.server       1
+                                        clojure-lsp.shared       1
+                                        clojure.core             1
+                                        clojure.edn              1
+                                        clojure.java.io          1
+                                        clojure.string           1
+                                        clojure.tools.cli        1
+                                        pod.clojure-lsp.api      1}
+                       :files          #{"~/code/clojure-lsp/cli/src/clojure_lsp/main.clj"}
```
ericdallo, line 111: Usually, I call `file` a java.io.file and let filename for strings, it'd follow the rest of the codebase naming I think too

ericdallo: One thing to keep in mind here is to maybe use always the absolute path instead of relative and in the future think about persists as URI instead of filename, there is an idea of kondo use URI for everything too for better compatibility with other features like remote LSP

lib/src/clojure_lsp/dep_graph.clj
```diff
@@ -0,0 +1,347 @@
...
+(comment
+  ;; Example
+  (require '[clojure-lsp.db :as db])
+  (-> @db/db*
+      :dep-graph
+      (select-keys '[clojure-lsp.main
+                     clojure.tools.cli]))
+  '{clojure-lsp.main  {:dependents     {clojure-lsp.main-test 1}
+                       :aliases        {main 1}
+                       :from-internal? true
```
ericdallo, line 98: Not sure it's clear that from-internal means that this ns used by a ns internally, maybe `used-internally?`

ericdallo: `internal-dependency`... naming is hard

lib/src/clojure_lsp/dep_graph.clj
ericdallo: looks great

lib/src/clojure_lsp/dep_graph.clj
```diff
@@ -0,0 +1,344 @@
...
+                       :internal?      true}
+    clojure.tools.cli {:dependents     {clj-depend.main                                    1
+                                        cljfmt.main                                        1
+                                        clojure-lsp.main                                   1
+                                        clojure.tools.deps.alpha.script.generate-manifest2 1
+                                        clojure.tools.deps.alpha.script.make-classpath2    1
+                                        clojure.tools.deps.alpha.script.print-tree         1
+                                        clojure.tools.deps.alpha.script.resolve-tags       1
+                                        kaocha.runner                                      1}
+                       :aliases        {cli 7, nil 1}
```
ericdallo, line 121: Should we have some way to differ `[require foo]` from `[require foo :refer :all]` ? would both be just `nil`?

lib/src/clojure_lsp/dep_graph.clj
```diff
@@ -0,0 +1,344 @@
...
+(defn ms-distinct [ms] (some-> ms keys))
+
+(defn ms-overlaps-set? [ms s]
+  ;; Conceptually set overlap is symmetric, but with our implementation of
+  ;; multiset, in practice we must check whether the items of the set are in the
+  ;; multiset, not vice-versa.
+  (boolean (and ms (some ms s))))
+
+(def ^:private ms-conj (fnil mulitset-conj empty-multiset))
+(def ^:private ms-disj (fnil multiset-disj empty-multiset))
```
ericdallo, line 165: Doing this now would be overengineering but I think it's worth to mention so we can be aware of this if need more performance: Since this is critical performance code for clojure-lsp, we could consider having a `Multiset` defrecord so we could have more performance probably to work on that data structure and a place for those functions

lib/src/clojure_lsp/dep_graph.clj
```diff
@@ -0,0 +1,344 @@
...
+  (reduce #(update-usage %1 f %2) db namespace-usages))
+
+(defn ^:private remove-usages [db usages] (update-usages db ms-disj usages))
+(defn ^:private add-usages [db usages]    (update-usages db ms-conj usages))
+
+(defn ^:private update-definition [db f {:keys [name filename]}]
+  (let [in-file (get-in db [:file-meta filename])]
+    (-> db
+        (update-in [:dep-graph name :files] f filename)
+        ;; TODO: is it wrong that once an internal ns, always an internal ns?
```
ericdallo, line 204: I think we can assume it's always internal, I can't see any case where this is not true

lib/src/clojure_lsp/dep_graph.clj
```diff
@@ -0,0 +1,344 @@
...
+                                                      (:to-cljs result)))]
+                         (reduce result-with-element result implicit-elements)))
+                     {:defs [] :usages []}
+                     analysis)]
+      [defs usages])))
+
+(defn refresh-analysis [db old-analysis new-analysis internal?]
+  ;; TODO: should this be wrapped in use-dep-graph? If it is, startup will be
+  ;; faster (500ms in medium projects 1200ms in v. large). But if it isn't, you
+  ;; can toggle the setting live.
```
ericdallo, line 282: We agreed on keep it for consistency, so we can remove the TODO and maybe just leave as a NOTE?

lib/src/clojure_lsp/dep_graph.clj
```diff
@@ -0,0 +1,344 @@
+(ns clojure-lsp.dep-graph
```
ericdallo, line 1: Pretty good functions on this file, each one with its own responsibility and very composable 👏🏻

lib/src/clojure_lsp/feature/completion.clj
```diff
@@ -207,64 +207,60 @@
...
+                                    :priority :required-alias
+                                    :kind :property
+                                    :detail detail}
+                             (seq require-edit) (assoc :additional-text-edits (mapv #(update % :range shared/->range) require-edit)))))))
+                aliases))
+        (let [alias-namespaces (->> aliases
+                                    (filter #(= (-> % :alias str) cursor-alias))
+                                    (map :to)
+                                    seq
+                                    set)]
```
ericdallo, line 263: nice

lib/src/clojure_lsp/queries.clj
```diff
@@ -1,5 +1,8 @@
 (ns clojure-lsp.queries
```
ericdallo, line 1: Lots of new ifs here, but expected.
After merge, let's try to fix all missing points to let dep-graph be the default soon, any fix on queries between that will probably be harder since we will need to keep testing both, glad we have the tests running for both states.

lib/src/clojure_lsp/dep_graph.clj
mainej: > filename for strings

I thought about that a couple of times as I was working on this—it's a good point. I'll refactor. 

> the absolute path

In the DB it's an absolute path. I replaced `/Users/jmaine` with `~` so that my name wasn't all over the example, but I'll find another way so that it isn't ambiguous. Maybe `/Users/me`.

I thought about using URI as a key. I'd rather do that too. But the dep graph is mostly about speeding up access to kondo data, so for now we need file names. Maybe those things should be changed in tandem.

lib/src/clojure_lsp/dep_graph.clj
mainej: Yes, both would be `nil`. So far there hasn't been any need to distinguish situations that use `:refer`. I think that's mostly because this code is about namespace definitions and usages, not var definitions and usages.

Actually, the presence of the `nil`s is mostly an accident. Allowing `nil` to be present makes the dep graph maintenance code a little simpler. None of the other code _needs_ to know which namespaces are required without an alias. Avoiding the `nil`s would be possible, but would add an `if` to handle an edge case we don't actually care about.

lib/src/clojure_lsp/dep_graph.clj
mainej: There are a few libraries that implement multisets as defrecords. But those libraries do a lot of other stuff—multiset versions of union, intersection, etc.—which we don't need. I didn't want to add a dependency if we weren't going to use much of its code.

FWIW, those libraries use defrecords so that they can implement various protocols. For example, by implementing `clojure.lang.IPersistentCollection`, you can call `(clojure.core/conj ms item)`, instead of what I have here `(mulitiset-conj ms item)`. But under the hood the defrecords in those libraries wrap a hashmap, which maps keys to frequencies, just like I've done here. So that extra layer of indirection might actually be slower. Anyway, if our goal was performance, we'd have test defrecords carefully.

lib/src/clojure_lsp/dep_graph.clj
mainej: Yeah, it'd be very unusual for a namespace to switch from internal to external. The only way it would happen is if you re-opened a dependency's namespace momentarily (external -> internal), then removed the code that did that (internal -> external). It's such a weird scenario that I'll just remove the comment.

lib/src/clojure_lsp/dep_graph.clj
mainej: Oh, right... thanks for the reminder.

lib/src/clojure_lsp/queries.clj
mainej: Yep, lots of new ifs, but I think they're all contained to this namespace. And I think most new scenarios will be able to use the existing helpers, so there won't be too many additional test cases. But still, I agree, I don't want to live in this limbo world for too long.

lib/src/clojure_lsp/feature/completion.clj
mainej: I still have a hard time reasoning about what this code does. Maybe I'll write myself a little note describing what it returns and when.

lib/src/clojure_lsp/queries.clj
mainej: 👍 on the tests... thanks for figuring those out. I'm glad we have them running for both states too.

lib/src/clojure_lsp/dep_graph.clj
ericdallo: I realized that later 😅 , I'm ok with something simple like `/path/to/clojure-lsp-project/...` or even the `/Users/me`, feel free to choose

lib/src/clojure_lsp/dep_graph.clj
ericdallo: yeah, makes sense, looks good for now

lib/src/clojure_lsp/dep_graph.clj
ericdallo: Sure, if we need in the future, we could start with out own defrecord simple implementation just encapsulating those functions, but something not important for now

lib/test/clojure_lsp/dep_graph_test.clj
```diff
@@ -0,0 +1,259 @@
...
+      (is (= '{ccc 1} (get-in db [:dep-graph 'aaa :dependents])))
+      (is (seq (get-in db [:dep-graph 'aaa :filenames])))
+      (is (seq (get-in db [:dep-graph 'bbb :dependents])))
+      (is (not (nil? (get-in db [:file-meta "/aaa.clj"])))))
+    #_(alter-var-root #'db/diagnostics-chan (constantly (async/chan 1)))
+    (with-redefs [shared/file-exists? (constantly false)]
+      (f.file-management/did-close "file:///aaa.clj" db/db*))
+    (let [db @db/db*]
+      (is (empty? (get-in db [:dep-graph 'aaa :dependencies])))
+      (is (= '{ccc 1} (get-in db [:dep-graph 'aaa :dependents]))) ;; <-- no change, because ccc stil depends on aaa, even though aaa is now undefined
```
ericdallo, line 164: good corner case, I wonder how that would affect us, like how bad is to store a ns dependent/dependency that its definition doesn't exist anymore

ericdallo: I know this is not the case this `is` is checking, but it made me thing about that

lib/src/clojure_lsp/dep_graph.clj
```diff
@@ -0,0 +1,344 @@
...
+;; other namespaces.
+
+;; :from-langs is a multiset of the langs of the :namespace-usages or files (if the
+;; :namespace-usages don't have langs) that use this namespace. May be either
+;; empty or absent if the namespace isn't required by other namespaces.
+
+;; file-meta-item =
+;; {(:namespaces #{ns*})?
+;;  :internal? boolean
+;;  :langs #{lang+}}
```
ericdallo, line 75: I think `file-metas` is similar to a already existing `:documents` in the db? would make sense to use `:documents` but add those extra infos there?

ericdallo (approved):
Looks really good!
LMK what you think about missing comments and feel free to merge it when it's done, looking forward to this feature

ericdallo:
FYI: tested on one of Nubank biggest clojure only projects (> 180K loc) and it's awesome to see something like:
```
:maintain-dep-graph 262ms
```

lib/test/clojure_lsp/dep_graph_test.clj
mainej: I've thought a little about what would happen if the dep graph gets corrupted. Obviously that _shouldn't_ happen, but if it does, what would be the result?

There are probably too many cases to consider fully. I don't think anything bad would happen if we store extra data, like relationships that don't actually exist anymore, or files that don't actually define a namespace. The reason is that whenever we have a namespace and a file that defines it, we can look up that file's analysis, but there's always the possibility that the file defines two or more namespaces. So, when we iterate through that file's elements we always (I think) double check that their `:from` is the namespace we're interested in. If we were storing extra data, I think this second check of namespace would eliminate extraneous elements.

If we store too _little_ data, i.e. miss a relationship or a file, the queries would return bad results. Queries wouldn't pull in the right files' analysis, and we'd return too few elements.

Of course I don't _think_ that the dep-graph can get out of sync with the analysis, and so I think we're safe. But, if it does, that's one of the things I'm hoping would turn up in beta testing.

lib/src/clojure_lsp/dep_graph.clj
mainej: I hadn't thought of that, but maybe... let me think about it more. One of the differences between `:file-meta` and `:documents` is that `:documents` is keyed by URI. But I'm considering your suggestion about re-keying by URI more seriously. I'll detail those thoughts in a separate comment.

As an aside... I'm really excited about your idea of caching the rewrite-clj parse. We all spend much more time reading code than writing it, so the parses aren't changing most of the time. I think having the parse cached would speed up things like `:hover`, which need a parse but which are invoked as the user navigates, not as they type. Now that I'm thinking about `:documents`, that's probably a good place to cache the parse. You could experiment with putting it there independently of when I merge dep-graph.

mainej:
> `:maintain-dep-graph 2119ms`

I'm pleased it doesn't slow down startup _too_ much. I think it's fast enough for now, but it's still a little longer than I'd like. A follow up PR will make it a tiny bit faster. But the real bummer is that whenever you run internal analysis, the code has to update the dep-graph twice, ending up with the same results both times. So, if you look at the logs again, I'd expect you to  see the 2 seconds of maintenance appear twice.

We've had this double calculation problem for a long time—normalization happens twice too—but now we have it with dep-graph maintenance as well. The reason is custom-lint-fn, which is why I was asking about it a few days ago. The custom lint fns receive raw clj-kondo analysis. To find unused public vars they need to query the db with that additional analysis, so they have to normalize it, merge it, and update the dep-graph. They do this speculatively, without actually touching the db atom. With this temp db they figure out the custom diagnostics, then return them to clj-kondo via `reg-finding!`. `reg-finding!` filters a few of them out based on the "ignores" that it knows about, and then returns the diagnostics to us along with the other diagnostics it found. It also returns the raw analysis again. Then we have to take that raw analysis, re-normalize, re-update the dep-graph, and finally swap everything into the db. In pictures, it looks like this:

```
clojure-lsp: change -> 
kondo: pass raw analysis to custom-lint-fn -> 
clojure-lsp: normalization+dep-graph -> (speculatively)
clojure-lsp: custom lint -> 
kondo: filter custom diagnostics by level and ignores ->
kondo: return raw analysis+full diagnostics ->
clojure-lsp: normalization+dep-graph -> (duplicate work!)
clojure-lsp: swap db
```

What I'd like to do is remove the custom lint fns and instead do this:

```
clojure-lsp: change -> 
kondo: return raw analysis+kondo diagnostics -> 
clojure-lsp: normalization+dep-graph -> 
clojure-lsp: custom lint -> 
kondo: filter custom diagnostics by level and ignores ->
clojure-lsp: concat kondo and custom diagnostics ->
clojure-lsp: swap db
```
The only thing making this difficult is the "ignores" filtering that `reg-finding!` does. That filtering isn't public, and it depends on the full `:ctx`, which holds the "ignores" data. So, in theory we could do it with a little refactoring to expose the filtering publicly, but we'd also need kondo to return the `:ctx`, so that we could use it _after_ `clj-kondo/run!` returns. We could work with @borkdude on the details, but if I were him I wouldn't be excited about leaking all the implementation detail of `:ctx`.

Again, it's not a huge problem—the duplicate work only amounts to a few seconds at startup and a few milliseconds as individual files are updated—but it still seems wasteful.

lib/src/clojure_lsp/dep_graph.clj
mainej: Let me think about using URIs a bit more...

URI pros:
- It'd be nice to avoid refactoring the dep-graph when kondo starts returning URIs.

neutral:
- When kondo starts returning URIs, we'll need a db/version bump, anyway.

URI cons:
- To use the dep-graph to access analysis, we'd need an extra pass to convert URIs to filenames. We'd cache filename by URI, limiting the expense. Still, needs performance testing.
- We won't have a way to cache the other direction, filename to URI. We've had performance problems when we accidentally do this conversion in a loop. Need to audit code, to see if we often need to convert a bunch of filenames to URIs.

mainej: I moved the file metadata into `:documents`. Looks fine from a performance perspective.

Thanks for the suggestion... I think it's nicer this way. It'll make it easier to refactor to URIs later.

To keep things fast, we still have a `:file-meta` key in the db, but all it does is map filename to uri. We'll be able to remove it when we use URIs consistently.

lib/src/clojure_lsp/dep_graph.clj
mainej: See other comment for details, but I moved the file metadata into :documents, as suggested. Much nicer. 💯

lib/src/clojure_lsp/dep_graph.clj
mainej: I want to keep `:from-internal?` consistent with `:from-langs`. I'm going to go with `:dependents-internal?` and `:dependents-langs`. It's still not perfectly clear, but it emphasizes that it's additional data about the dependents. And as you said naming is hard.

PR #1061, opened 2022-06-20
Ignore increase/decrease in usages when linting references

### LINKED ISSUE #1019: performance: `file-management/reference-filenames` could return fewer results
**Describe the bug**

Currently we re-lint dependencies whenever the number of usages is increased or decreased. This means we re-lint when the number of usages goes from 5 to 6 or 7 to 2. But technically we need to re-lint only when the number increases from or decreases to 0. Those are the cases where we might need to remove or add (respectively) unused-public-var lint. Fixing this might help with the performance concerns in #1018 because it is much more common to use a keyword for a second, third or nth time in a given file than to start using it for the first time.

mainej:
Instead, only lint when usage added/removed. Fixes #1019.

- [x] I created an issue to discuss the problem I am trying to solve or an open issue already exists. #1019

### DIFF
--- a/lib/src/clojure_lsp/crawler.clj
+++ b/lib/src/clojure_lsp/crawler.clj
@@ -40,10 +40,6 @@
                      (lsp.kondo/db-with-results kondo-result)
                      (lsp.depend/db-with-results depend-result))))))
 
-(defn analyze-reference-filenames! [filenames db*]
-  (let [result (lsp.kondo/run-kondo-on-reference-filenames! filenames db*)]
-    (swap! db* lsp.kondo/db-with-results result)))
-
 (defn lerp "Linear interpolation" [a b t] (+ a (* (- b a) t)))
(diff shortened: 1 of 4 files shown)


lib/src/clojure_lsp/crawler.clj
```diff
@@ -40,10 +40,6 @@
                      (lsp.kondo/db-with-results kondo-result)
                      (lsp.depend/db-with-results depend-result))))))
 
-(defn analyze-reference-filenames! [filenames db*]
-  (let [result (lsp.kondo/run-kondo-on-reference-filenames! filenames db*)]
-    (swap! db* lsp.kondo/db-with-results result)))
```
ericdallo, line 45: nice

ericdallo (approved):

ericdallo (approved):

PR #1062, opened 2022-06-20
Extracted functions are private

### LINKED ISSUE #1039: feat: extracted functions should be private
`promote-fn` and `create-function` both create private functions when possible. For consistency, `extract-function` should do the same.

mainej:
For consistency with promote-fn and create-function.

Fixes #1039

- [x] I created an issue to discuss the problem I am trying to solve or an open issue already exists. #1039


ericdallo (approved):
Nice!

Issue #728, opened 2022-01-29
:additional-text-edits don't get returned on completionItem/resolve


Cyrik:
Left over from #720 

The coercion for this case needs to be handled. Should probably be based on #727

mainej:
This group of changes was all for Calva, right? Does that mean additionalTextEdits aren't applied to Calva at all? Or does it apply the the edits it gets before starting completion/resolve?

mainej:
Or I guess a better question is, does that mean additionalTextEdits are currently broken for all editors? I can check that myself for Emacs.

ericdallo:
I don't think they are broken since we have a feature of adding the require when completing a not required symbol, you can test from that feature I think

Cyrik:
They are broken for editor's that use resolve as the source of truth, so anything based on top of official libs. In those additional edits won't be used at all because resolve does not return them.
I think emacs wrote the whole client? They don't seem to follow the spec on this bit.
I think postponing them would actually fix it for calva since then the resolve would return it. Currently it doesn't 😁

ericdallo:
@mainej Almost sure you are absolutely correct, thanks for summarizing.

There are few features that I know that emacs-lsp is smart and tries to be safe from server misinformation like this one.
Even so, we should try to follow the spec to fix the issue for most(all?) clients that don't do the same, and probably improving in-request coercer would be a good start.

BTW calva uses vscode-languageclient node lib, which already has most built-in capabilities enabled, for the rest, Calva enable manually on the lsp.ts file on the server object initialization.

mainej:
@ericdallo ok, thanks. I'll move forward with looking at postponing additionalTextEdits. Doesn't seem like it'll hurt anything. I know the input coercion is more important, but I'm intimidated by that project. :)

For my reference in the future [here](https://github.com/microsoft/vscode-languageserver-node/blob/6c689d22501aa2bac9990f3a41139d47428b5f40/client/src/common/client.ts#L1745) are the Calva capabilities.

mainej:
I've decided not to move additionalTextEdits to resolve, at least for now. If anyone is interested, the reasons are [here](https://github.com/clojure-lsp/clojure-lsp/pull/764#issuecomment-1053684207). So, in regards to this bug, nothing has changed. As discussed above, moving additionalTextEdits to resolve would partially solve the bug, but A) I'm not going to work on that right now, and B) the better fix is still to change input coercion so that it can copy additionalTextEdits during resolve.

[...]

PR #1069, opened 2022-06-21
Postpone calculation of ns edits to speed up completion

### LINKED ISSUE #1068: performance: postpone require edits to completionItem/resolve
Sometimes completion knows that for a completion item to be valid the `ns` form will have to be edited to require a new namespace. It calculates the edit for each completion item and includes that in its response. When there are many of these completion items, the edit calculations slow down completion. These calculations could be postponed to `completionItem/resolve`, so that only one edit is calculated, and only after the user selects its item. See #1053 for background.

mainej:
Sometimes completion knows that for a completion item to be valid the `ns` form will have to be edited to require a new namespace. It calculates the edit for each completion item and includes that in its response. When there are many of these completion items, the edit calculations slow down completion.

Now these calculations are postponed to `completionItem/resolve`, so that only one edit is calculated, and only after the user selects its item. See #1053 for background.

This also stops asking clients to resolve documentation if they don't support that. And it preserves "snippet-kind", which is a secret backdoor to slip additional data to some clients.

Fixes #1068. I expect this to have conflicts with the dep-graph-queries branch, but I'll resolve them, whichever one gets merged first.

- [x] I created an issue to discuss the problem I am trying to solve or an open issue already exists. #1068


snoe:
Currently, clients can show the user that an extra edit will occur if they choose a completion (I use 🍦 to indicate). Do we lose that with this change? I find it very useful to have the indication because it lets me know if I'll be pulling in a namespace I didn't expect to.
image

ericdallo:
@snoe I'm not sure if resolveItem is called on all visible items or only when the selected one is highlighted

snoe:
> @snoe I'm not sure if resolveItem is called on all visible items or only when the selected one is highlighted

@ericdallo Ok, checked out the branch to test. Looks like neither to me.
image
No indication, but the namespace is added once chosen. Is this a client issue? Otherwise I think we shouldn't do this by default.

ericdallo:
Looks like client issue, but we can check the protocol to know when clients should call resolveItem

mainej:
The [spec says](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion), "If computing full completion items is expensive, servers can additionally provide a handler for the completion item resolve request (‘completionItem/resolve’). This request is sent when a completion item is selected in the user interface."

To me, these instructions to clients are ambiguous. What does "when a completion item is selected" mean? I'm pretty sure it doesn't mean that resolve should be called for all visible items. But does it mean it should be called when you highlight (mouse over/arrow down) an item? Or does it mean it should be called when you choose (mouse click/type enter) an item?

The spec goes on, "A typical use case is for example: the [textDocument/completion](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion) request doesn’t fill in the documentation property for returned completion items since it is expensive to compute. When the item is selected in the user interface then a ‘completionItem/resolve’ request is sent with the selected completion item as a parameter. The returned completion item should have the documentation property filled in."

I don't know the VSCode UI very well, but I imagine that it shows you documentation when you highlight an item. Showing you documentation after you've chosen an item would be too late. That is, I think the spec authors intended "selected" to mean "highlighted".

So, I think we can say that what this PR is doing is acceptable behavior, and that you're seeing a bug in coc.nvim—either that it isn't calling `resolve` when you highlight an item, or that it it is, but it's not noticing that the return of that call includes `additionalTextEdits` and that it should therefore add a 🍦.

As far as showing 🍦 for every thing that _will eventually_ do additional text edits, I think this is a gap in the spec, a gap I noticed when I was writing this code. If you're going to do resolution, you really want to be able to tell the client which things you plan to resolve, and you want to have separate params for each one instead of a single opaque `data` field. (For example, documentation resolution needs a ns or position, while alias resolution needs a ns name and alias.) Having to stick everything in the opaque `data` field means A) that the client's can't predict what will happen when they call resolve (and therefore can't show 🍦 on every item that can resolve additional text edits) and B) every server has to come up with their own format for how to pass itself the resolution params.

mainej:
@snoe I checked Calva and ... lol. 

Before this patch, Calva doesn't show a 🍦 or anything like that (or at least it doesn't with my "Suggestions" settings, which I think are the defaults.) So I don't think this patch changes that aspect of the Calva experience. Do you know if there's a setting that would show that kind of indicator?

More interesting, and much funnier to me, is that before this patch Calva doesn't add the namespace at all. After the patch, it does. So it seems there's a bug in Calva which this patch unintentionally works around. If I had to, I'd guess that the VS Code LSP client assumes that because it's advertised to the server that it _can_ do additionalTextEdits during resolution that it will _only_ get additionalTextEdits during resolution, and then it (incorrectly) ignores additionalTextEdits in the primary completion request.

mainej:
Actually, @PEZ, I should have you weigh in on this... See above for details, but clojure-lsp provides additional data with some completions that would add a missing alias. But in Calva, those edits don't get applied when the completion is chosen. Is that a bug report you'd like on Calva, or is it an upstream problem?

If this patch gets merged, then I think Calva users would start getting the alias additions, although the bug would still be present in VS Code.

mainej:
If you want to reproduce the VS Code bug, open clojure-lsp/lib/src/clojure_lsp/feature/format.clj, and somewhere in the file type `q`. clojure-lsp should suggest "q alias to: clojure-lsp.queries", as well as public functions like "q/db-with-project-analysis". If you select any of those, the `:require` section of the ns form should get a new  `[clojure-lsp.queries :as q]` alias, but doesn't.

mainej:
I rebased this onto master, which has dep-graph-queries merged. With that branch and this one combined, completing `met` in metabase, which [was taking](https://github.com/clojure-lsp/clojure-lsp/pull/1053#issuecomment-1158274811) 350-400ms, now takes 20ms. 🚀

mainej:
I was just reading back through some old PRs and bug reports. Back in February I spent some time researching whether postponing edits would improve completion performance—see #764. At the time I concluded it wouldn't help much, but I hadn't run across this motivating case of trying to complete `met` in metabase.

A little before that time a bug in completion had been noticed #728. Apparently in trying to understand the relationship between the bug and the new research, I had [worked out](https://github.com/clojure-lsp/clojure-lsp/issues/728#issuecomment-1052847800) that Calva wasn't applying text edits and that postponing would probably fix it. So what I said before was wrong... I did know that bug existed and that postponing would fix it. I just completely forgot! February was a long time ago, I guess.

In #728 I categorized the different clients:

1. those that don't support resolve at all
2. those that support resolve, but not additionalTextEdits
3. those that support resolve, including additionalTextEdits

After this PR, type 1 and 2 clients will get the edits in textDocument/completion. Type 3 will get the edits in completionItem/resolve.

I think that will make type 1 and type 3 work, which is good, because that covers all the actual clients I know of. Type 2 clients are hypothetically possible, but we don't know of any. But if there are any, they might still have a problem...

The spec says that additionalTextEdits are supposed to be copied from the input to the output of completionItem/resolve. Because of some ugly debeaning problems our code still doesn't do that. That may mean that we would send additionalTextEdits to type 2 clients in textDocument/completion, but lose them in completionItem/resolve, and that therefore those clients would still see the bug described in #728. So, I won't claim that this fixes #728.

This PR does however fix #728 for type 3a clients (see the bug for the meaning of "a"), including Calva simply by excluding the additionalTextEdits from textDocument/completion.

I expect that the debeaning problems would go away with https://github.com/clojure-lsp/lsp4clj/issues/8, which would finally fix #728 for all clients.

mainej:
@snoe I submitted a coc.nvim [issue](https://github.com/neoclide/coc.nvim/issues/3911) for the missing snippet indicator.

ericdallo:
Ping @snoe, is this PR ok to be merged for vim clients?

snoe:
@ericdallo I mean, I don't like it as as it feels like it's going to degrade the experience too much for me to use it til it's fixed upstream. I'll leave it up to you and @mainej to make the call though.

ericdallo:
I'm ok waiting for upstream to avoid a bad experience for vim users

snoe:
I'll also say, I'm working on metabase which the linked issue uses as a reason for this pr, and while the speed of completions could be better I feel they are generally usable day-to-day.

mainej:
I don't mind maintaining this branch until the upstream fix. @snoe perhaps you could add a 👍 to that Issue or an explanation of why it's important to you.

I agree, generally completion is quite fast already. I just got so used to it being nearly instant that I was surprised to find a case where it felt draggy. It's not just metabase—it's noticeable in a more medium-sized project like clojure-lsp too.

But, I admit, it's not _that_ common to try to complete `met` in metabase or `clo` in clojure-lsp, and those are really the problem cases. You run across it if you do a lot of manual editing of the ns form, although personally, I rarely do that anymore. I almost exclusively rely on the "add require" refactoring lately.

Calculating the ns edits also slows down one of the most common completion scenarios in any project, completing `foo/b`, where you're completing a var in an aliased namespace. But, it's often less noticeable then because the time is proportional to the number of var definitions in the aliased namespace, instead of proportional to the number of namespaces that begin with the project's name. To be thorough, there's an alternative fix for this particular scenario. If an aliased namespace defines 100 vars, we currently (pre-PR) calculate the ns edit 100 times, but we could memoize that calculation. I considered this but decided it'd be better to fix both scenarios the same way, hence this PR.

mainej:
@snoe it looks like that coc.nvim issue has been closed without a fix. The reason is a bit cryptic, but I think they're saying that it's not possible to alter the pop-up menu after it's been created. Their suggestion, I think, is to have the server modify the documentation it returns during `completionItem/resolve` to indicate when an additionalTextEdit is being applied. I'm not sure whether to be appreciative of that as a clever hack or disappointed that the hack would affect all users, not just coc.nvim users.

If we wanted to get fancy, we could let users pick what icon 🍦 they wanted to see in the documentation, and only show it if they have a selection. Or we could hard-code one particular icon.

I don't know what's best. Coming back to whether this PR is worth it... It changes your flow @snoe. It improves performance for all users, though in a narrow set of completion scenarios. And it fixes automatic requires for all Calva users. I still think it's worth it, but I could be convinced to abandon it. What do you think @ericdallo?

snoe:
@mainej looking at https://github.com/clojure-lsp/clojure-lsp/issues/728#issuecomment-1052847800 

Would this change break type 1 clients that do not support resolve? e.g. nvim core
It seems that what we should be doing is looking at capabilities, then clients (and users through client config) can decide which they want.

mainej:
@snoe, if I understand your question, no, type 1 clients won't be broken.

The PR already takes into account client capabilities (see the `resolve-support` check in `completion-item-with-unresolved-alias-edit`). As such type 1 clients will get the additionalTextEdits in the first (and only) `textDocument/completion` request. So this change wouldn't break their ability to apply the edits. If they were doing some sort of 🍦 thing, that'll continue to work too, although I don't know if nvim core was.

Type 1 clients also don't support documentation resolution. (This was true before this PR, and remains true after it. It'd be pretty easy to fix. I'm not sure whether it was an oversight originally, or if it was intentional as a way to avoid an expensive documentation lookup for every edit.) Therefore, they wouldn't get documentation that has been modified to include the 🍦, but they weren't getting any documentation before either.

snoe:
> Type 1 clients also never get documentation. (This was true before this PR, and remains true after it. It'd be pretty easy to fix. I'm not sure whether it was an oversight originally, or if it was intentional as a way to avoid an expensive documentation lookup for every edit.) 

Ah sounds like there was a regression, I'm guessing when clojure.docs came into play (@ericdallo ?). There could be a faster but worse documentation we send for type 1 clients.

Let's ship this as it seems to be the way forward and deal with the fallout, maybe it's just me who will miss 🍦 or we go with their suggestion to add it on the backend (or something else).

ericdallo:
We should add docs for type 1 clients, probably I missed that, let's do it in a different issue/PR though

mainej:
https://github.com/clojure-lsp/clojure-lsp/pull/1069/commits/62684d014c6c7727d3938386886b4f7989fc953c implements the [suggestion](https://github.com/neoclide/coc.nvim/issues/3911#issuecomment-1200101179) from the coc.nvim team.

If the setting `{:completion {:additional-edits-warning "🍦"}}` is present and a completion will perform additional edits—such as requiring a new alias—then we show 🍦 in the completion item's documentation.

I think that's the best we're going to get, for now.

@ericdallo this is ready for a final review, unless you still have objections @snoe.

docs/all-available-settings.edn
```diff
@@ -22,6 +22,7 @@
  :hover {:hide-file-location? false
          :arity-on-same-line? false
          :clojuredocs true}
+ :completion {:additional-edits-warning nil}
```
ericdallo, line 25: Maybe make clear this should be a string?
```suggestion
 :completion {:additional-edits-warning-text nil}
```

ericdallo (approved):

docs/all-available-settings.edn
mainej: Yep, will change.

PR #1079, opened 2022-06-23
Removes several blocking puts from file-management

mainej:
Uses put! (async) instead of >!! (blocking). We don't need these calls to block, so better to let them be async.

Also refactors tests to have better tools for testing what gets put on chans.

Still to do:

- [ ] Create integration test for `did-change`
- [ ] Consider replacing `(go (>! ...))` with `(put! ...)` for all diagnostics publishing.
- [ ] Alternatively, the opposite, replace all `(put! ...)` with `(go (>! ...))`

### DIFF
--- a/.clj-kondo/config.edn
+++ b/.clj-kondo/config.edn
@@ -5,4 +5,5 @@
                                       rewrite-clj.node]}}
  :lint-as {rewrite-clj.zip.subedit/edit->    clojure.core/->
            rewrite-clj.zip.subedit/subedit-> clojure.core/->
-           clojure-lsp.refactor.edit/my-defn clojure.core/defn}}
+           clojure-lsp.refactor.edit/my-defn clojure.core/defn
+           clojure-lsp.test-helper/let-mock-chans clojure.core/let}}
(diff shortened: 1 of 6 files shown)


mainej:
@ericdallo now that we know that `>!!` isn't the true root cause of #1080, this branch isn't very relevant anymore. I don't think we need to replace `>!!` with `put!`. In fact, `put!` would just mean that any problems would show up much later, when the pending put queue finally filled up.

That said, there are a lot of things in this branch worth saving. I'll plan to do some cleanup and resubmit as a different PR. But before I do that, I want to ask you whether you want to keep a few things...

1. This wires lsp4j, which uses java.utils.logging, to SLF4J, and wires SLF4J to timbre. This has two effects. First, it gets rid of the SLF4J warnings during builds. Second, in theory at least, it means that lsp4j logs would appear in our logfile. I've never seen any of these logs, so I'm not sure. Either things still aren't wired correctly or lsp4j simply doesn't log very much. The downside of this is that it adds two new dependencies.
2. It wraps `System/out` in a way that would throw IOExceptions if that stream were ever closed. This is a very speculative change. I've never seen it do anything, and I don't know if it's necessary. There's a chance that it might fix certain things (like it might prevent zombie servers if a client disconnects abnormally), but again that's speculation.

I'm on the fence. I'd be OK with keeping or dropping either. WDYT?

mainej:
Here's a PR with the things that I think are definitely worth keeping: https://github.com/clojure-lsp/clojure-lsp/pull/1090

PR #1081, opened 2022-06-28
Install script: account for uname -m being arm64 on macos m1

borkdude:


ericdallo (approved):
thanks, I missed that one

ericdallo:
Actually, I remember why I didn't change that yet, that won't work with previous versions right @borkdude ? we would need to compare versions just like babashka does right?

borkdude:
Oh, right

ericdallo:
I mean, I think most people install latest release which wouldn't be a issue, so not sure

borkdude:
Maybe we can support the `--arch` option so people can override it for older versions.

PR #1082, opened 2022-06-28
Rollback critical changes

ericdallo:


mainej:
@ericdallo I think most of this branch can be discarded. But what are your thoughts on whether to rollback the `async/put!` for diagnostics?

Here are some different perspectives:
1. We don't care about responses, so there's no need to block. `put!` is better. In fact, all of the variations of publish-diagnostics should use `put!`, even in tests. (The tests are being changed by #1090 to handle the asynchrony.)
2. Same as 1) we don't care about responses, but we're worried that the async put queue will fill up. When this happens it means that the take end of one of the channels has stopped. That's the real problem, but it manifests on the put end. Both tools—`>!!` and `put!`—have difficulty, either blocking or eventually throwing an error. Both are hard to debug because the problem is elsewhere, but maybe blocking is better because it happens a little sooner. `>!!` is better.
3. This shouldn't use core.async at all. A `future` would work fine. We want to submit a job (send the diagnostics notification) to a thread pool, but as soon as we have, we're done. We don't care about the response. We expect the thread to be busy the whole time it's processing the job. The job doesn't need to relinquish the thread so that other work can be done (i.e., it doesn't need to use `go` to park.) `future` is better. The downside is that there's one thing that occasionally _does_ care about an unfinished job—the tests. It's convenient to provide a mock chan in the tests to see what's getting submitted as jobs.

Do any of these perspectives make you want to follow one approach or another? I think 3 is off the table because of the tests. But I could be convinced to go with 1 or 2.

ericdallo:
I'd say 3 could cause performance issues as well AFAICS, there is a occasion where we call publish diagnostics frequently in a short period of time: when clojure-lsp finish the project analysis and start to publish all kondo/clj-depend findings for the whole project, meaning for projects like metabase we may have hundreds or even thousands of diagnostics being published.
For those cases, using `future` for each one would probably not work because of multiple threads spawned right? so clojure.async.core may fit better here

I think I'd stick with 1 because of the fact we don't care about the response, just fire and forget, **but** since `>!!` can help a little bit better if any issues, 2. sounds better in the long term.

mainej:
> using `future` for each one would probably not work because of multiple threads spawned right

Good point, `(chan 1)` is another benefit of using core.async.

> stick with 1... 2. sounds better in the long term

OK, I think we're fine for now... this PR can be closed. I'll leave that to you.

Issue #1080, opened 2022-06-24
clojure-lsp silently stops working in calva after a while


calherries:
Note: duplicated issue in calva repo: https://github.com/BetterThanTomorrow/calva/issues/1782

**Describe the bug**
After about 10 minutes after starting clojure-lsp and using the editor, it’s functionality in the editor stops. I can solve it temporarily by restarting clojure-lsp, but it soon fails again.

The issue persists after disconnecting with the REPL.

Things that stop working:
- Any new syntax errors I create in the editor aren’t underlined with linting errors (existing ones before it stops working persist)
- Any new errors I create aren’t showing in the “Problems” tab (but existing ones persist)
- Go to definition
- All code navigation and refactoring commands

Things that are still working:

The clojure-lsp log file still logs handlers, publishing diagnostics, etc. If I intentionally create a syntax error I can see the logs update with a correctly updated “Publishing X diagnostics” message
```
2022-06-24T09:45:14.936Z DEBUG [clojure-lsp.server:82] - Publishing 11 diagnostics for file:///Users/callumherries/meta/metabase/src/metabase/query_processor/card.clj
```
VS code still shows clojure-lsp as working, and when I click the button, it shows the relevant quick actions.

**To Reproduce**
It happens reliably on my machine after jumping between files, definitions, etc., so I don’t know the specific cause. I haven’t tried with a minimal example but it happens while I’ve been working on https://github.com/metabase/metabase, which is open source.

<details>
 <summary><b>Log - client <-> server</b></summary>
<pre>
[Trace - 10:23:59 PM] Sending notification '$/cancelRequest'.
Params: {
    "id": 808
}

Jun 24, 2022 10:23:59 PM org.eclipse.lsp4j.jsonrpc.RemoteEndpoint handleCancellation
WARNING: Unmatched cancel notification for request id 808
[Trace - 10:23:59 PM] Sending request 'textDocument/codeAction - (809)'.
Params: {
    "textDocument": {
        "uri": "file:///Users/callumherries/meta/metabase/src/metabase/query_processor/card.clj"
    },
    "range": {
        "start": {
            "line": 249,
            "character": 23
        },
        "end": {
            "line": 249,
            "character": 23
        }
    },
    "context": {
        "diagnostics": []
    }
}

[Trace - 10:23:59 PM] Sending notification '$/cancelRequest'.
Params: {
    "id": 807
}

[...]

PR #1087, opened 2022-06-29
Revert "Revert "Improve hover element selected (#1045)" (#1083)"

mainej:
This reverts commit 012323126d0a9bf5aee4a8a2aa2c1663be001f85, restoring #1045 while avoiding problems in Calva, problems documented in #1080 and discussed at length on [Slack](https://clojurians.slack.com/archives/CPABC1H61/p1656093769590509).

The new hover code introduced in #1045 was the first real use of `parser/safe-zloc-of-string`. (It was used only in tests before, though we had plans to use it elsewhere.)

But there was a subtle bug in `safe-zloc-of-string`. When there was a parse error, it didn’t log to the logfile, it **printed to stdout**.

So the new hover code accidentally caused a print to stdout when there was a syntax error in the file. This corrupted the server’s output stream permanently. From this moment on, Calva was waiting for the stream to contain a valid message, but it never would. In the meantime, clojure-lsp kept writing to the stream, but it didn’t matter, because Calva couldn’t process it.

Changing the `println` to `logger/warn` fixes the problem.

We suspect that this wasn't a problem in Emacs because lsp-mode does a better job detecting and discarding messages that don't conform to the JSON-RPC spec.

The longer term fix is to avoid ever printing to the stream that lsp4j is using. If we're going to use `System/out`, lsp4j needs to have sole control over when that stream is written to. To this end, we're planning a longer-term fix to lsp4clj. We'll wrap every call into clojure-lsp with `with-out-str`, to avoid having output leak through to the clients. We'll log any output we receive to the log file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Editor
+ - Restored #995 Improve element selected on `textDocument/hover` (previously reverted) with a fix that keeps it working for Calva even after a syntax error is introduced.
+
 ## 2022.06.29-19.32.13
 
 - Editor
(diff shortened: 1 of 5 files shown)


ericdallo (approved):
Tested with Calva and it works

PR #1088, opened 2022-06-29
Remove safe-zloc-of-string

mainej:
This removes `safe-zloc-of-string`. If code needs a zloc, but the file has a syntax error, it's acceptable to short-circuit by throwing an exception. None of the current code needs to continue anyway. This formalizes that notion by removing the 'safe' version of parsing, allowing exceptions to be raised. The 'safe' version wasn't used anyway except by tests and one new codepath, all of which are actually fine with using the 'unsafe' version.

Getting rid of this nearly-dead code is a post-mortem TODO item from #1087. As usual with dead code, it's best to remove so that nobody accidentally starts using something that hasn't been reviewed recently and which may not be fully up to date. That was the case with #1087—a long time ago a bug was introduced in `safe-zloc-of-string`, where it printed to stdout instead of logging. Nobody noticed that until some new code started using `safe-zloc-of-string`. The long passage of time between changing `safe-zloc-of-string` and finally starting to use it made it very hard to track down as the root cause of #1087.

### DIFF
--- a/lib/src/clojure_lsp/feature/hover.clj
+++ b/lib/src/clojure_lsp/feature/hover.clj
@@ -116,7 +116,7 @@
   (let [db @db*
         filename (shared/uri->filename uri)
         zloc (some-> (f.file-management/force-get-document-text uri db*)
-                     (parser/safe-zloc-of-string)
+                     (parser/zloc-of-string)
                      (parser/to-pos line column))
         {function-loc-row :row
          function-loc-col :col} (some-> (edit/find-function-usage-name-loc zloc)
--- a/lib/src/clojure_lsp/feature/signature_help.clj
+++ b/lib/src/clojure_lsp/feature/signature_help.clj
@@ -77,7 +77,6 @@
 (defn signature-help [uri row col db*]
   (let [filename (shared/uri->filename uri)
         zloc (some-> (f.file-management/force-get-document-text uri db*)
-                     ;; TODO: use safe-zloc-of-string and handle nils
                      (parser/zloc-of-string)
                      (parser/to-pos row col))
         function-loc (edit/find-function-usage-name-loc zloc)]
--- a/lib/src/clojure_lsp/handlers.clj
+++ b/lib/src/clojure_lsp/handlers.clj
@@ -275,7 +275,7 @@
      :client-settings (:client-settings db-value)
      :final-settings (settings/all db-value)
      :cljfmt-raw (binding [*print-meta* true]
-                   (with-out-str (pr (f.format/resolve-user-cljfmt-config db-value))))
+                   (pr-str (f.format/resolve-user-cljfmt-config db-value)))
      :port (or (:port db-value)
                "NREPL only available on :debug profile (`make debug-cli`)")
      :server-version (shared/clojure-lsp-version)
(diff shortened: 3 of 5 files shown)


lib/src/clojure_lsp/feature/hover.clj
```diff
@@ -116,7 +116,7 @@
   (let [db @db*
         filename (shared/uri->filename uri)
         zloc (some-> (f.file-management/force-get-document-text uri db*)
```
ericdallo, line 118: I think at least for hover feature, it could makes sense to have a safe or a try-catch it because if we don't find a correct zloc, hover can still be calculated from kondo analysis right? Although there is a chance to be outdated since we can't parse the code too.. so I think it's ok to throw the exception. WDYT @mainej ?

ericdallo: Another thought is that that kind of exceptions may log a lot on server log, in this case, logging just a line mentioning the parse didn't work sounds better to me too, I can see users that don't use paredit like tools checking server logs and complaining about those exceptions just because for a moment their code was in a invalid state before they fix the parens for example

mainej: When we can't parse, we don't have outdated analysis, we have no analysis. That is, we send the invalid text to clj-kondo and, along with a diagnostic about the syntax error, it returns an empty vector of analysis. So we never find any hover information to return. The net effect was that with `safe-zloc-of-string` we were creating a successful response with no results. With `zloc-of-string` we're creating an error response. From the user's perspective, it's the same–nothing happens in the editor. The only real difference is whether the logs contain a stacktrace.

That said, I agree, stacktraces in the logs are alarming even if they don't mean anything is wrong.

So maybe the better approach is to make `zloc-of-string` private, and instead use `safe-zloc-of-string` everywhere. Anywhere that relies on an Exception being thrown to short-circuit can be refactored to check for nils instead. There aren't actually many uses of either function, so it wouldn't be hard to refactor.

WDYT @ericdallo? Would that be better?

To be honest, the main reason I wanted to remove `safe-zloc-of-string` was that it was previously dead code and it caused #1080. But after the `println` was fixed there was really nothing wrong with it. So I'd be OK if we revive it.

ericdallo: > When we can't parse, we don't have outdated analysis, we have no analysis. That is, we send the invalid text to clj-kondo and, along with a diagnostic about the syntax error, it returns an empty vector of analysis. So we never find any hover information to return. The net effect was that with safe-zloc-of-string we were creating a successful response with no results. With zloc-of-string we're creating an error response. From the user's perspective, it's the same—nothing happens in the editor. The only real difference is whether the logs contain a stacktrace.

Agreed

> WDYT @ericdallo? Would that be better?

Yeah, I'd like just to avoid the unnecessary logs, so sounds good to me
I agree it was dead code, but now, it may be a good opportunity to start using it in the correct way avoiding the verbose exception logs I think

mainej:
Closing in favor of #1092.

lib/src/clojure_lsp/feature/hover.clj
mainej: 👍 I moved the switch to safe-zloc-of-string to #1092.

PR #1090, opened 2022-06-30
Extend tests of file-management

mainej:
Adds tests for the async messages created by did-open, did-change and did-change-watched-files. Improves the tools for testing async messages. Adds integration test for did-change.

### DIFF
--- a/cli/integration-test/entrypoint.clj
+++ b/cli/integration-test/entrypoint.clj
@@ -10,6 +10,7 @@
     integration.definition-test
     integration.declaration-test
     integration.implementation-test
+    integration.text-change-test
     integration.code-action-test
     integration.completion-test
     integration.diagnostics-test
@@ -36,8 +37,8 @@
       (future-cancel fut))
     ret))
 
-(defn log-tail []
-  (:out (sh/sh "tail" "-n" "300" "clojure-lsp.integration-test.out" :dir "integration-test/sample-test/")))
+(defn log-tail [file lines]
+  (:out (sh/sh "tail" "-n" (str lines) file :dir "integration-test/sample-test/")))
 
 (def first-print-log-tail?* (atom true))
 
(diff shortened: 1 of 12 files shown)


cli/integration-test/entrypoint.clj
```diff
@@ -36,24 +37,27 @@
...
   (when (medley/deref-reset! first-print-log-tail?* false)
     (binding [*out* *err*]
       (println "--- RECENT LOG OUTPUT ---")
-      (print (log-tail))
-      (println "--- END RECENT LOG OUTPUT ---"))))
+      (print (log-tail "clojure-lsp.integration-test.out" 300))
+      (println "--- END RECENT LOG OUTPUT ---")
+      #_(println "--- LSP TRACE ---")
+      #_(print (log-tail "clojure-lsp.lsp-trace.out" 500))
+      #_(println "--- END LSP TRACE ---"))))
```
ericdallo, line 53: Any other reason than debugging purpose to have these comments?

lib/test/clojure_lsp/features/diagnostics_test.clj
```diff
@@ -290,15 +290,18 @@
...
+          [mock-diagnostics-chan #'db/diagnostics-chan]
           (h/load-code-and-locs code)
-          (is (= ["user/foo is called with 3 args but expects 1 or 2"
-                  "user/baz is called with 1 arg but expects 3"
-                  "user/bar is called with 0 args but expects 1 or more"
-                  "user/foo is called with 3 args but expects 1 or 2"
-                  "user/foo is called with 0 args but expects 1 or 2"
-                  "user/foo is called with 4 args but expects 1 or 2"]
-                 (map :message (get @h/mock-diagnostics "file:///a.clj")))))))
+          (let [{:keys [uri diagnostics]} (h/take-or-timeout mock-diagnostics-chan 500)]
```
ericdallo, line 296: cool, looks better than use the old mock diagnostics indeed

cli/integration-test/entrypoint.clj
mainej: Correct, when paired with the commented out tracing launcher (in server.clj), these lines assist in debugging. While I'm still debugging the flaky tests, I'd prefer to have this launcher and logging on all the time in integration tests. I tried to add a `:trace-path` setting, similar to the `:log-path` setting, so that I could configure which launcher gets used. But I couldn't get it to work because settings aren't available during launch, for whatever reason. So I left this around to make it easier the next time I want to turn on these traces.

PR #1092, opened 2022-06-30
Use parser/safe-zloc-of-string

mainej:
Converts all usages of `parser/zloc-of-string` to `parser/safe-zloc-of-string`. Instead of letting exceptions short-circuit, we short-circuit on nils. This keeps stack traces out of the logs.

This patch also improves the performance of call hierarchies by parsing less. This was just an opportunistic refactoring—while looking at the parsing done in call-hierarchy I noticed were doing it a lot more frequently than necessary. There's more performance work that can be done here. See comments in call_hierarchy.clj.

### DIFF
--- a/lib/src/clojure_lsp/feature/call_hierarchy.clj
+++ b/lib/src/clojure_lsp/feature/call_hierarchy.clj
@@ -11,14 +11,25 @@
 
 (set! *warn-on-reflection* true)
 
+(defn ^:private safe-zloc-of-forced-uri [db* uri]
+  (some-> (f.file-management/force-get-document-text uri db*)
+          (parser/safe-zloc-of-string)))
+
+;; TODO: call-hierarchies could be made much faster if
+;; element-by-uri->call-hierarchy-item didn't re-parse so often.
+;; 1. Reason through whether it really needs the branch that parses. The tests
+;;    suggest it doesn't.
+;; 2. Failing that, when dep-graph is merged, use it to find ns name.
+;; 3. Failing that, gather URIs that need to be parsed up front, parse each one
+;;    once, then thread those root zlocs through the rest of the code.
 (defn ^:private element-by-uri->call-hierarchy-item
-  [{uri :uri
-    {:keys [ns filename name-row name-col arglist-strs deprecated] el-name :name :as parent-element} :parent-element
-    usage-element :usage-element} db*]
-  (let [project-file? (string/starts-with? uri "file://")
+  [{:keys [uri parent-element usage-element]} db*]
+  (let [{:keys [ns filename name-row name-col arglist-strs deprecated] el-name :name} parent-element
+        project-file? (string/starts-with? uri "file://")
         detail (if project-file?
-                 (some-> (f.file-management/force-get-document-text uri db*)
-                         (parser/zloc-of-string) ;; throws on invalid Clojure
+                 ;; TODO: tests pass if this branch is removed. Why can't we use
+                 ;; the else branch for project files?
+                 (some-> (safe-zloc-of-forced-uri db* uri)
                          (parser/to-pos name-row name-col)
                          edit/find-namespace-name)
                  (or (some-> ns str)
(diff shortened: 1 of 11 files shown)


lib/src/clojure_lsp/feature/call_hierarchy.clj
```diff
@@ -42,19 +53,27 @@
...
+  (when-let [{:keys [row col]} (some-> root-zloc
+                                       (parser/to-pos row col)
+                                       (edit/find-var-definition-name-loc)
+                                       z/node
+                                       meta)]
+    (when-let [definition (q/find-element-under-cursor db filename row col)]
+      (when (or (identical? :var-definitions (:bucket definition))
+                (and (identical? :var-usages (:bucket definition))
+                     (:defmethod definition)))
+        definition))))
```
mainej, line 69: This logic used to be in `edit`, even though it was only needed here. That was making `edit`, which is primarily about manipulating zlocs, too intimately familiar with the analysis. So, this moves the logic and implements it more efficiently.

lib/src/clojure_lsp/feature/call_hierarchy.clj
```diff
@@ -68,32 +87,35 @@
...
-               {:from-ranges []
-                :from (element-by-uri->call-hierarchy-item element-by-uri db*)}))))
+  (let [db @db*
+        references (q/find-references-from-cursor db (shared/uri->filename uri) row col false)
+        filenames (->> references (map :filename) distinct)
+        file-meta (zipmap filenames
+                          (map (fn [filename]
+                                 (let [uri (shared/filename->uri filename db)]
+                                   {:uri uri
+                                    :root-zloc (safe-zloc-of-forced-uri db* uri)}))
```
mainej, line 97: Parse once per filename, not once per reference.

lib/src/clojure_lsp/handlers.clj
```diff
@@ -275,7 +275,7 @@
      :client-settings (:client-settings db-value)
      :final-settings (settings/all db-value)
      :cljfmt-raw (binding [*print-meta* true]
-                   (with-out-str (pr (f.format/resolve-user-cljfmt-config db-value))))
+                   (pr-str (f.format/resolve-user-cljfmt-config db-value)))
```
mainej, line 278: Equivalent, but more idiomatic. Found during a post-mortem of #1080, while I was looking for places we might accidentally print to stdout.

lib/src/clojure_lsp/refactor/edit.clj
```diff
@@ -112,33 +112,11 @@
...
-                                     (get-in db [:analysis filename]))]
-        (when (seq var-definition-ops)
-          {:fn-name-loc fn-name-loc
-           :var-definitions var-definition-ops})))))
-
-;; TODO Move the query part to queries.clj ?
-(defn find-var-definition-name-loc [loc filename db]
-  (:fn-name-loc (find-var-definition loc filename db)))
+(defn find-var-definition-name-loc [loc]
+  (some-> loc to-top z/next var-name-loc-from-op))
```
mainej, line 116: This function is a little more permissive now. It no longer confirms that we have analysis of a var def at the loc. But that's OK, since this only has two usages. First, create-test uses it, but create test only needs the var definition name, not an analysis element. And second, call-hierarchy uses it. Call-hierarchy does need an analysis element, but it actually prefers to find that element itself.

lib/src/clojure_lsp/feature/call_hierarchy.clj
```diff
@@ -11,14 +11,25 @@
...
-    usage-element :usage-element} db*]
-  (let [project-file? (string/starts-with? uri "file://")
+  [{:keys [uri parent-element usage-element]} db*]
+  (let [{:keys [ns filename name-row name-col arglist-strs deprecated] el-name :name} parent-element
+        project-file? (string/starts-with? uri "file://")
         detail (if project-file?
-                 (some-> (f.file-management/force-get-document-text uri db*)
-                         (parser/zloc-of-string) ;; throws on invalid Clojure
+                 ;; TODO: tests pass if this branch is removed. Why can't we use
+                 ;; the else branch for project files?
```
ericdallo, line 31: Good question, I can't recall why I added that at the time, but if we can always get the namespace reliable and fallback for filenames when not seems already good enough, so the else branch makes sense.
Also, parsing the loc just to find the namespace name sounds too much

lib/src/clojure_lsp/feature/call_hierarchy.clj
ericdallo: Great!

lib/src/clojure_lsp/refactor/edit.clj
ericdallo: 💯

ericdallo (approved):
LGTM

lib/src/clojure_lsp/feature/call_hierarchy.clj
mainej: Looks like the parsing was added all the way back in bc39d0ed7df803bde86dbcc7638e9ef6754f5484. It was introduced before we started using `ns`, which was added in 2e154faccf38652351bfeff1717a719818f9397a. I think everything could have been switched to `ns` at that time. I'll make the switch now, keeping the filename fallback. 🎉

ericdallo: That makes sense! Nice

ericdallo (approved):

PR #1096, opened 2022-07-02
Index analysis by file and bucket

### LINKED ISSUE #1018: bug: `file-management/reference-filenames` neglects keywords

mainej:
Index analysis by file _and_ bucket, instead of just file as we did before. This is helpful because the majority of queries need to look in only select buckets.

@ericdallo as we discussed, this is the second of three PRs that bump the db/version. I'd like to get them all in this release, if possible. The third one will be much smaller.

This PR improves performance a little, but as with the third PR, it's mostly preparation for fixing #1018. To fix #1018 we'll need very fast access to keyword definitions, which could be in any file. This PR will help with that by letting us skip many analysis elements (especially var usages and locals). The third PR will split keywords into usages and definitions which will be the last step to being able to access keyword definitions quickly.

- [x] This is prep for #1018

### DIFF
--- a/lib/src/clojure_lsp/kondo.clj
+++ b/lib/src/clojure_lsp/kondo.clj
@@ -74,9 +74,19 @@
       (update :findings merge findings)
       (assoc :kondo-config config)))
 
-(defn entry->normalized-entries [{:keys [bucket arglist-strs] :as element}]
-  (cond
-    (identical? :var-definitions bucket)
+;;;; Normalization
+
+;; The analysis we get back from clj-kondo is indexed by bucket
+;; {(bucket [element+])*}
+;; We re-index by filename then bucket
+;; {(filename {(bucket [element+])+})*}
+;; We also normalize elements somewhat, changing their keys, and even spliting
+;; some into two elements. As such, the buckets we get from clj-kondo aren't
+;; exactly the same as the buckets we have in our db.
+
+(defn ^:private element->normalized-elements [{:keys [bucket arglist-strs] :as element}]
+  (case bucket
+    :var-definitions
     [(-> element
          (shared/assoc-some :arglist-kws (->> arglist-strs
                                               (keep (fn [arglist-str]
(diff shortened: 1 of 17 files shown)


lib/src/clojure_lsp/kondo.clj
```diff
@@ -74,9 +74,19 @@
...
-    (identical? :var-definitions bucket)
+;;;; Normalization
+
+;; The analysis we get back from clj-kondo is indexed by bucket
+;; {(bucket [element+])*}
+;; We re-index by filename then bucket
+;; {(filename {(bucket [element+])+})*}
+;; We also normalize elements somewhat, changing their keys, and even spliting
+;; some into two elements. As such, the buckets we get from clj-kondo aren't
+;; exactly the same as the buckets we have in our db.
```
ericdallo, line 85: This may be interesting for @borkdude, I wonder if clj-kondo could do that for us already in a way we don't need breaking changes, like a new flag or something, I can see this being very convenient for clj-kondo analysis clients for performance reasons

ericdallo (approved):
This is something we should do a long time I think, thanks!

Looks good codewise, a little bit concern if we missed any place which we are still accessing via the old filename index only, but since we have good tests for queries, we just need to seek those features that query manually analysis, which I think you covered them all.

mainej:
> something we should do a long time

I tested this code by itself, without dep-graph-queries, and it provides some of the same benefit. But dep-graph-queries is still quite a bit faster, and together they're faster still.

> concern if we missed any place

That's always a worry in these big refactorings. There aren't too many usages of analysis outside of queries and I've touched them all recently, so I think I found them all. Another grep for `analysis` confirms. Still—good to keep an eye out for.

PR #1097, opened 2022-07-02
Separate analysis of keyword definitions

mainej:
Some performance sensitive things like completion and custom linting need to find all keyword definitions. But to get to them they had to search through every keyword usage. Since there are orders of magnitude more keyword usages than keyword definitions, this was unnecessarily slow.

This splits clj-kondo keywords into keyword-usages and keyword-definitions during analysis normalization. Now it's much faster to get to the definitions.

Before this patch completing a keyword in metabase took about 100ms. Now it takes 30ms.

This will assist in fixing #1018 and is the last of the three PRs that bump db/version.

- [x] I created an issue to discuss the problem I am trying to solve or an open issue already exists. #1018 

### DIFF
--- a/lib/src/clojure_lsp/queries.clj
+++ b/lib/src/clojure_lsp/queries.clj
@@ -363,13 +363,12 @@
              rf-some
              (get-in db [:analysis filename :locals])))
 
-(defmethod find-definition :keywords
+(defmethod find-definition :keyword-usages
   [db element]
   (or (when (:ns element)
         (find-last-order-by-project-analysis
-          :keywords
+          :keyword-definitions
           #(and (= (:name %) (:name element))
-                (:reg %)
                 (= (:ns %) (:ns element)))
           db))
       element))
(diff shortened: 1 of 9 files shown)


ericdallo (approved):
Looks good, I will test this a little bit today then we can merge.

I was wondering if this should be an option on kondo as well as it's one more normalization we are doing on clojure-lsp side differing from kondo.

lib/src/clojure_lsp/queries.clj
```diff
@@ -363,13 +363,12 @@
              rf-some
              (get-in db [:analysis filename :locals])))
 
-(defmethod find-definition :keywords
+(defmethod find-definition :keyword-usages
   [db element]
   (or (when (:ns element)
```
mainej, line 368: It's possible to register keywords without a namespace, but this `when` means we wouldn't find those registrations. I think this is a good time to fix that. WDYT @ericdallo?

ericdallo: Yes, To be hones I don't recall why we did that 🤔

mainej: Confirmed, this is pretty easy to fix... I'll submit a follow-up PR after this one is merged.

ericdallo (approved):

PR #1101, opened 2022-07-05
Fix: update unused-public-var lint on keywords as usages change

### LINKED ISSUE #1018: bug: `file-management/reference-filenames` neglects keywords

mainej:
This fixes #1018. I was concerned that fixing 1018 would cause performance problems in find-references, because we have to find the files where keywords are defined. But, with other recent changes (#1096 and #1097), performance of find-references is still pretty fast.

This PR also improves performance in a later stage, in the code that actually finds unused public keywords. We no longer generate a list of all the keyword usages in the whole project if the files we're linting don't have any keyword definitions.

And finally, it makes it slightly easier to test linting, by testing only the pure part. See lib/test/clojure_lsp/features/diagnostics_test.clj. In retrospect, I'm not sure this was worth it—I didn't end up adding new tests in that area. In fact, since the tests no longer check that `reg-finding!` is called, technically coverage is reduced. I don't find it particular worthwhile to test `reg-finding!`, but if you'd like, I'll revert that part of this PR.

- [x] Fixes #1018


lib/src/clojure_lsp/feature/diagnostics.clj
```diff
@@ -185,47 +184,64 @@
...
                     (some var-usages (q/var-definition-signatures var-def)))
-        kw-definitions (remove (partial exclude-public-diagnostic-definition? config) kw-defs)
-        ;; TODO: skip if no kw-definitions
-        kw-usages (into #{}
-                        (comp
-                          q/xf-all-keyword-usages
-                          (map q/kw-signature))
-                        (:analysis project-db))
+        kw-definitions (remove (partial exclude-public-diagnostic-definition? kondo-config) kw-defs)
+        kw-usages (if (seq kw-definitions) ;; avoid looking up thousands of keyword usages if these files don't define any keywords
```
ericdallo, line 188: Nice improvement

lib/src/clojure_lsp/feature/diagnostics.clj
```diff
@@ -185,47 +184,64 @@
...
   (let [project-db (q/db-with-internal-analysis db)]
-    (lint-defs! (file-var-definitions project-db filename)
-                (file-kw-definitions project-db filename)
-                project-db kondo-ctx)))
+    (unused-public-vars (file-var-definitions project-db filename)
+                        (file-kw-definitions project-db filename)
+                        project-db kondo-config)))
+
+(defn ^:private finalize-findings! [findings reg-finding!]
+  (run! reg-finding! findings)
```
ericdallo, line 231: Looks better, if we stop using custom-lint-fn in the future, it's easier to remove this code here

ericdallo (approved):
Looks good! 
About stop testing `reg-finding!`, I know code is way easier to read now, since we decoupled retrieving findings and applying `reg-finding!`, even so, `reg-finding!` is critical for diagnostics work properly, without it, clj-kondo won't notice and will not consider those diagnostics when finishing linting, so I still think it's a good idea to test it, does that makes sense for you?

lib/src/clojure_lsp/feature/diagnostics.clj
mainej: Yeah, this was a simple fix that should help a lot.

mainej:
> reg-finding! is critical ... so I still think it's a good idea to test it

That's true. OK, I'll leave the refactoring which separates the pure and impure parts, but revert the tests.

PR #1103, opened 2022-07-05
Fix: navigate to definition of potemkin var

### LINKED ISSUE #1020: bug: `find-definition` from a var defined by `potemkin/import-vars` goes to any definition with the same name

mainej:
Always navigate to the correct definition of a var imported by potemkin, instead of any var with the same name.

- [x] Fixes #1020

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
   - Fix: find definition will find registration of unnamespaced keyword.
   - Fix to update unused-public-var lint on registered keywords as usages change in other files. #1018
   - Fix to navigate to var defined by declare, when there aren't any later defs. #1107
+  - Fix to always go to the definition of the correct var imported by potemkin. #1020
 
 ## 2022.06.29-19.32.13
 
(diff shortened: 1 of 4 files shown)


mainej:
Note that this depends on a SNAPSHOT version of clj-kondo. Let me know if you'd rather wait for a regular release.

ericdallo:
@mainej it's good to keep in sync with stable kondo release but we already released lots of versions using specific master versions, so it's ok.
The issue is that you are using a SNAPSHOT, not a commit specific one, so each time we build clojure-lsp it will get latest clj-kondo version, could you change to a specific clj-kondo one? you can find from the circle-ci build deploy task on clj-kondo, it should be something like `clj-kondo-12345233-231231923923`

mainej:
> each time we build clojure-lsp it will get latest clj-kondo version

Right, good point. Fixed now.

ericdallo (approved):
Looks good! feel free to fix conflicts and merge

PR #1106, opened 2022-07-06
Log stdout, instead of sending it to clients

mainej:
See https://github.com/clojure-lsp/lsp4clj/issues/9 and https://github.com/clojure-lsp/lsp4clj/pull/11

I'm submitting this as a draft, so I can see whether the integration tests capture any stdout.

- [x] I created an issue to discuss the problem I am trying to solve or an open issue already exists. See https://github.com/clojure-lsp/lsp4clj/pull/11

### DIFF
--- a/lib/src/clojure_lsp/kondo.clj
+++ b/lib/src/clojure_lsp/kondo.clj
@@ -288,18 +288,14 @@
         (with-additional-config (settings/all db)))))
 
 (defn ^:private run-kondo! [config err-hint db]
-  (let [err-writer (java.io.StringWriter.)
-        out-writer (java.io.StringWriter.)]
+  (let [err-writer (java.io.StringWriter.)]
     (try
       (let [result (if (:api? db)
                      (kondo/run! config)
-                     (binding [*err* err-writer
-                               *out* out-writer]
+                     (binding [*err* err-writer]
                        (kondo/run! config)))]
         (when-not (string/blank? (str err-writer))
           (logger/warn "Non-fatal error from clj-kondo:" (str err-writer)))
-        (when-not (string/blank? (str out-writer))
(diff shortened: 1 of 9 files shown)


lib/src/clojure_lsp/kondo.clj
```diff
@@ -288,18 +288,14 @@
         (with-additional-config (settings/all db)))))
 
 (defn ^:private run-kondo! [config err-hint db]
-  (let [err-writer (java.io.StringWriter.)
-        out-writer (java.io.StringWriter.)]
+  (let [err-writer (java.io.StringWriter.)]
     (try
       (let [result (if (:api? db)
```
mainej, line 293: Hmm... when run with `:api?` we weren't capturing stdout. But now we are, because `shared/capturing-stdout` wraps this call. So the kondo output, which was going to stdout, will now go to the logs. Is this a problem though? `kondo/run!` normally doesn't put anything on stdout.

mainej:
This needs https://github.com/clojure-lsp/lsp4clj/pull/11 before it can be taken out of draft.

lib/src/clojure_lsp/kondo.clj
ericdallo: This is a fix!
For CLI at least this is good as we were logging unnecessary kondo logs even without --verbose flag, for API I think it's acceptable to lose that log too

mainej:
Thanks for approving this @ericdallo, but I think we may want to abandon this branch. It's reimplemented in #1117, and if we merged this branch to master, it would complicate the merge of 1117 slightly. WDYT? I don't have strong feeling either way.

ericdallo:
Yeah, makes sense, let's close this

PR #1110, opened 2022-07-07
Re-model LSP JSON-RPC with core.async channels

mainej:
This gives clients a uniform interface to conform to. They model messages as Clojure hashmaps, and put those messages to and take them from a pair of channels.

This cleans up the integration tests slightly, but is actually prep for https://github.com/clojure-lsp/lsp4clj/issues/8. Here, we're using this code to model a client, but we'll be able to use the same code to model a server. With tools to convert stdio to channels, we get a few benefits.

1. It will be easier to implement socket communication https://github.com/clojure-lsp/lsp4clj/issues/1 as a complement to stdio communication. We'll need to write the code that converts socket i/o to channels, but the client (or actually server) won't change.
2. It will be easier to write mocks for tests. The tests (even unit tests) can read from and write to a server's channels to make assertions about how it responds without having to understand the wire level of the LSP JSON-RPC protocol.
3. There will be a natural mechanism for concurrency. Servers can choose to read several messages from a channel, distributing each to a pool of workers.

### DIFF
--- a/cli/integration-test/integration/client.clj
+++ b/cli/integration-test/integration/client.clj
@@ -1,9 +1,8 @@
 (ns integration.client
   (:require
-   [cheshire.core :as json]
-   [clojure.java.io :as io]
-   [clojure.string :as string]
-   [integration.helper :as h])
+   [clojure.core.async :as async]
+   [integration.helper :as h]
+   [integration.lsp-json-rpc :as lsp-json-rpc])
   (:import
    [java.time LocalDateTime]
    [java.time.format DateTimeFormatter]))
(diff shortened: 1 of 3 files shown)


ericdallo (approved):
Looks good! Handling that communication with Clojure instead of raw messages makes things easier

mainej:
I'm going to merge this @ericdallo, but while working on https://github.com/clojure-lsp/lsp4clj/issues/8 I discovered that this code has a bug. It reads `Content-Length` characters but `Content-Length` is actually in _bytes_. It's working because the integration tests are all in ASCII (# characters == # bytes), but would break if they used non-ASCII characters.

Before I merge, I'll add a note about that. The long term fix is to wait for the https://github.com/clojure-lsp/lsp4clj/issues/8 work to be done. Then we can use the i/o -> channel helpers from lsp4clj in the integration test client.

PR #1113, opened 2022-07-10
Query xf

mainej:
@ericdallo what do you think of this refactoring? The idea was to extract the common things queries do into transducers. I think it's interesting how it makes the queries more declarative. It's also a nice way to encourage the correct use of `identical?` and `safe-equal?` But I can also see how it could make them less approachable for someone who was just scanning the code for the first time. Up to you whether to go this direction.


ericdallo (approved):
I think it's ok as it clarifies what some `comps` are doing, but it probably makes it a little bit trickier as well for people not used to the code to understand it? 

We can merge it and check how easy we/other people will think about this in the future

Issue #826, opened 2022-03-15
lsp4clj features to move


Cyrik:
This is just a collection of functions / features we should think about moving out of the cli/lib folder:

1. most parts of server/run-server!
2. server/client-settings
3. shared/debound-by (and sub functions)
4. db/ all the channels
5. a protocol version of f.file-management/analyze-changes and analyze-watched-created-files!
6. file_management/did-change + handle-change

ericdallo:
Not sure I agree with most suggestions, IMO lsp4clj should be agnostic of language implementations, we could use lsp4clj to implement a python language server for example, and not sure those servers would make sense use db/channels, debounced-by, and things like that, so we should probably discuss each one of them separately

ericdallo:
Check lsp4j, we should be as agnostic as them IMO

Cyrik:
Yeah, I do want it to be agnostic. But all of them need shared/debounce-by
The ones that call Clojure specific stuff would need to be made generic/turned into protocols. But that's only only parts of 2 and 5

ericdallo:
I'm closing this since this is actually a mirror of https://github.com/clojure-lsp/lsp4clj/issues/14 which has some good discussion there

Issue #818, opened 2022-03-12
Lsp4clj left-overs


Cyrik:
Leftover comments from the PR and issue. Should probably be cleaned up into 3+ issues, just don't want to lose the for now.

---
Yeah, the problem is that the server needs to implement the function and the lib needs to call it. And we do have it on the API, `ClojureLspProducer` in server.clj implements it. I'm also not happy with the way this part is working.

The good thing is that this is all purely Clojure, with no lsp4j or java in the way. So I think we can come up with something better here.
I do like the producer design, so maybe we should just have two of those. One is provided by lsp4clj with a default implementation for built-in LSP things and a second one for custom messages. The custom one can look however you want since the implementation and the calling side is fully outside of lsp4clj. lsp4clj would just handle construction and clean up for you.
_Originally posted by @Cyrik in https://github.com/clojure-lsp/clojure-lsp/pull/805#discussion_r824773595_

- I'd like ILSPFeature to already have the actual params as input, so that it's easier to know what to implement and easier to write a spec for it.
I think this will be easier when we run the incoming messages through the coercion as well instead of only going through java->clj, so I'm happy to wait.

- Extending the server with custom messages means you have to write Java classes (because of attributes). @ericdallo started building a more generic way to do it for dependencyContents messages, but seems to have abandoned it. Was there a reason? Do we want to find an easier way to extend the server before we push this out?
Do we pull more stuff from lib, like the transit caching? (agree with @ericdallo, lets do this in a separate PR)

ericdallo:
Closing this since we merged #1117, Talking with @mainej, we think we can open specific issues for what #1117 didn't fix

PR #1117, opened 2022-07-14
Remove dependency on lsp4j

### LINKED ISSUE #727: Fix incoming request coercion
Based on the problems left in #720 

I'd like to do the camelCase <-> kebab-case conversion by hand in the clojure.java.data/from-java multimethod, but clojure-lsp.server/sync-request pipes the clojure-lsp.coercer/java->clj call directly into the handler.
I think the cleanest solution would be to also pipe this through a request-spec spec before calling the handler. That way we can change the casing or any other stuff that might come up at the "right" location.

I'm not sure when I'll have time to work on this but I wanted to keep the design idea alive.

### LINKED ISSUE #728: :additional-text-edits don't get returned on completionItem/resolve

mainej:
I've submitted a [PR](https://github.com/clojure-lsp/lsp4clj/pull/12) to lsp4clj that removes the dependency on lsp4j, as [proposed](https://github.com/clojure-lsp/lsp4clj/issues/8) earlier. This is the clojure-lsp side of that conversion.

I believe this also fixes or addresses most of the issues left over from the initial extraction of lsp4clj from clojure-lsp.

/cc @Cyrik: See more notes addressed to you in the lsp4clj [PR](https://github.com/clojure-lsp/lsp4clj/pull/12).

- Fixes #727
- Fixes #728 
- Addresses (or fixes) #818 (Input is exactly what the client provided, coerced to kebab-case keywords. Servers choose which LSP methods to implement.)
- Addresses #826. It is now much less code to start a server, and none of it involves Java interop. I agree with @ericdallo that most of the other aspects of 826 shouldn't be extracted to lsp4clj.

To test, follow the usual procedure... check out this branch and run `make`. ~Until lsp4clj is released, the only way to test this branch is to clone and check out [this branch](https://github.com/mainej/lsp4clj/tree/lsp2clj) of lsp4clj. Put it at the same level as clojure-lsp, then `make` clojure-lsp.~ (edit: this clojure-lsp branch's dep on lsp4clj is pinned to a `:git/sha` in the lsp4clj branch until it's merged and released.)

Here's the new dependency tree of clojure-lsp:

<details>
 <summary><code>cd cli && clj -Stree</code></summary>
<pre>
org.clojure/clojure 1.11.1
  . org.clojure/spec.alpha 0.3.218
  . org.clojure/core.specs.alpha 0.2.62
org.clojure/tools.cli 1.0.206
borkdude/dynaload 0.2.2
nrepl/bencode 1.1.0
com.taoensso/timbre 5.2.1
  . com.taoensso/encore 3.21.0
    . org.clojure/tools.reader 1.3.6
    . com.taoensso/truss 1.6.0
  . io.aviso/pretty 1.1.1
com.github.clojure-lsp/lsp4clj /Users/jmaine/workspace/opensource/lsp4clj
  . org.clojure/core.async 1.5.648
    . org.clojure/tools.analyzer.jvm 1.2.2
      . org.clojure/tools.analyzer 1.1.0
      . org.clojure/core.memoize 1.0.253
        . org.clojure/core.cache 1.0.225
          . org.clojure/data.priority-map 1.1.0
      . org.ow2.asm/asm 9.2
      . org.clojure/tools.reader 1.3.6
  . camel-snake-kebab/camel-snake-kebab 0.4.3
  . cheshire/cheshire 5.11.0
    . com.fasterxml.jackson.core/jackson-core 2.13.3
    . com.fasterxml.jackson.dataformat/jackson-dataformat-smile 2.13.3
      . com.fasterxml.jackson.core/jackson-core 2.13.3
    . com.fasterxml.jackson.dataformat/jackson-dataformat-cbor 2.13.3
      . com.fasterxml.jackson.core/jackson-core 2.13.3
    . tigris/tigris 0.1.2
clojure-lsp/lib /Users/jmaine/workspace/opensource/clojure-lsp/lib
  . org.benf/cfr 0.152
  . com.github.clj-easy/stub 0.2.3
  X com.github.clojure-lsp/lsp4clj /Users/jmaine/workspace/opensource/lsp4clj :use-top
  . cljfmt/cljfmt 0.8.0
    X org.clojure/tools.cli 1.0.206 :use-top
    X org.clojure/tools.reader 1.3.5 :older-version
    X rewrite-clj/rewrite-clj 1.0.605-alpha :older-version
  . rewrite-clj/rewrite-clj 1.1.45
    . org.clojure/tools.reader 1.3.6
  . clj-kondo/clj-kondo 2022.06.22
    X com.cognitect/transit-clj 1.0.324 :older-version
    . io.lambdaforge/datalog-parser 0.1.9
    . cheshire/cheshire 5.11.0
    X nrepl/bencode 1.1.0 :use-top
    . org.babashka/sci 0.3.2
      . borkdude/sci.impl.reflector 0.0.1
      . borkdude/edamame 0.0.19
        X org.clojure/tools.reader 1.3.4 :older-version
    X babashka/fs 0.1.2 :older-version
    . org.ow2.asm/asm 9.2
  . com.googlecode.java-diff-utils/diffutils 1.3.0
  . com.cognitect/transit-clj 1.0.329
    . com.cognitect/transit-java 1.0.362
      X com.fasterxml.jackson.core/jackson-core 2.8.7 :older-version
      . org.msgpack/msgpack 0.6.12
        . com.googlecode.json-simple/json-simple 1.1.1
        . org.javassist/javassist 3.18.1-GA
      . javax.xml.bind/jaxb-api 2.3.0
  . anonimitoraf/clj-flx 1.2.0
  . borkdude/rewrite-edn 0.2.0
  . babashka/fs 0.1.6
  . com.fabiodomingues/clj-depend 0.6.1
    . org.clojure/tools.namespace 1.2.0
      . org.clojure/java.classpath 1.0.0
      . org.clojure/tools.reader 1.3.6
    X org.clojure/tools.cli 1.0.206 :use-top
  . medley/medley 1.4.0
  . org.clojure/core.async 1.5.648
</pre>
</details>

- [x] Addresses issues listed above


mainej:
I've pinned clojure-lsp's dep on lsp4clj to a commit in the `v1` branch, so it's possible to test this with a regular `make`. I'll update the PR message to reflect that.

Also, CI is _almost_ passing. There's an intermittent failure where `$/progress` messages arrive at the client out of order. I don't understand yet how that's happening... will require some research.

mainej:
I found and fixed what I think was causing the integration test to be flaky. So, this should be all ✅

mainej:
Spoke too soon... GraalVM compiles but then doesn't run. It also doesn't create any log files. I thought GraalVM was working at some point, so I'll have to figure out what happened.

mainej:
I forgot some type hints when I was adding tracing. Fixed and ✅ now, for real.

cli/integration-test/integration/diagnostics_test.clj
```diff
@@ -107,8 +105,6 @@
                                                                               :accessed-by-layers #{:c}}}})}))
   (lsp/notify! (fixture/initialized-notification))
   (lsp/notify! (fixture/did-open-notification "diagnostics/depend/a.clj"))
-  ;; TODO remove after integration timeout fixes
-  (Thread/sleep 200)
```
ericdallo, line 111: 💯

cli/integration-test/integration/client.clj
```diff
@@ -28,111 +32,128 @@
...
+                     output
+                     ;; TODO: return error until initialize request is received? https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize
+                     ;; `keep` means we do not reply to responses and notifications
+                     (keep #(receive-message this context %))
+                     input)]
+      (async/go
+        ;; wait for pipeline to close, indicating input closed
+        (async/<! pipeline)
+        (deliver join :done)))
+    ;; invokers can deref the return of `start` to stay alive until server is
```
ericdallo, line 87: 👍🏻

cli/resources/META-INF/native-image/clojure-lsp/clojure-lsp/reflect-config.json
```diff
@@ -17,49 +17,6 @@
   "name":"clojure.lang.Var",
   "allPublicMethods":true
 },
-{
```
ericdallo, line 20: Graal build should be faster without all those reflection configs :)

Makefile
```diff
@@ -19,9 +19,6 @@ clean:
            docs/README.md \
            docs/CHANGELOG.md
 
-classes:
```
ericdallo, line 22: I'm so glad we don't need to handle java on the project anymore 🚀

cli/src/clojure_lsp/nrepl.clj
```diff
@@ -12,10 +13,12 @@
...
 
-(defn setup-nrepl [db]
+(defn setup-nrepl [db*]
   (try
     (when-let [port (repl-port)]
       (logger/info "====== LSP nrepl server started on port" port)
-      (swap! db assoc :port port))
+      ;; db/db* only available in dev. Don't use it otherwise.
+      (alter-var-root #'db/db* (constantly db*))
+      (swap! db/db* assoc :port port))
```
ericdallo, line 22: I don't understand this change, could you elaborate?

cli/src/clojure_lsp/server.clj
```diff
@@ -74,103 +77,286 @@
...
+        (lsp.server/send-request server "workspace/codeLens/refresh" nil))))
+
   (publish-workspace-edit [_this edit]
-    (some-> (producer/publish-workspace-edit lsp-producer edit)
-            deref))
+    (lsp.server/discarding-stdout
+      (let [request (->> {:edit edit}
+                         (conform-or-log ::coercer/workspace-edit-params)
+                         (lsp.server/send-request server "workspace/applyEdit"))
+            response (lsp.server/deref-or-cancel request 10e3 ::timeout)]
```
ericdallo, line 102: Quite useful this `deref-or-cancel`

cli/src/clojure_lsp/server.clj
```diff
@@ -74,103 +77,286 @@
...
+        (if (= ::timeout response)
+          (logger/error "No reponse from client after 10 seconds.")
```
ericdallo, line 104: Maybe mention on what kind of action client didn't respond?
```suggestion
          (logger/error "No reponse from client after 10 seconds while applying workspace-edit.")
```

cli/src/clojure_lsp/server.clj
```diff
@@ -74,103 +77,286 @@
...
+                         (conform-or-log ::clojure-coercer/publish-test-tree-params)
+                         (lsp.server/send-notification server "clojure/textDocument/testTree"))))))))))
 
-(defn client-settings [params]
-  (-> params
-      :initializationOptions
-      (or {})
-      shared/keywordize-first-depth
-      (settings/clean-client-settings)))
+;;;; clojure experimental features
```
ericdallo, line 154:
```suggestion
;;;; clojure extra features
```

cli/src/clojure_lsp/server.clj
```diff
@@ -199,18 +385,70 @@
...
+(defmethod lsp.server/receive-request "initialize" [_ {:keys [db* server] :as components} params]
+  (logger/info "Initializing...")
+  ;; TODO: According to the spec, we shouldn't process any other requests or
+  ;; notifications until we've received this request. Furthermore, we shouldn't
+  ;; send any requests or notifications (except for $/progress and a few others)
+  ;; until after responding to this request. However, we start sending
+  ;; diagnostics notifications during `handler/initialize`. This particular case
+  ;; might be fixed by delaying `spawn-async-tasks!` until the end of this
+  ;; method, or to `initialized`, but the more general case of being selective
+  ;; about which messages are sent when probably needs to be handled in lsp4clj.
```
ericdallo, line 418: Agreed

cli/src/clojure_lsp/server.clj
```diff
@@ -219,53 +457,57 @@
...
+          producer (ClojureLspProducer. server db*)
+          components {:db* db*
+                      :logger timbre-logger
+                      :producer producer
+                      :server server}]
+      (logger/info "[SERVER]" "Starting server...")
+      ;; TODO: if this were moved to `initialize`, after timbre has been
+      ;; configured, the server's startup logs and traces would appear in the
+      ;; regular log file instead of the temp log file. The downside would be that
+      ;; if anything bad happens before `initialize`, we wouldn't get any logs.
```
ericdallo, line 509: Yes, it happened before exceptions before we update the new log path, so makes sense

lib/src/clojure_lsp/settings.clj
```diff
@@ -53,20 +53,19 @@
...
-        (update :document-range-formatting? (fnil identity true)))))
+  (-> client-settings
+      (update :dependency-scheme #(or % "zipfile"))
+      (update :text-document-sync-kind kwd-string)
+      (update :source-paths parse-source-paths)
+      (update :source-aliases parse-source-aliases)
+        ;; (update :project-specs #(->> % (mapv kwd-keys) not-empty))
+      (update :cljfmt-config-path #(or % ".cljfmt.edn"))
+        ;; (update :cljfmt kwd-keys)
+        ;; (update :linters kwd-keys)
```
ericdallo, line 64: Why those are commented? I guess it's safe to remove

ericdallo (approved):
Looks amazing!
Will start manually testing it

ericdallo:
@mainej just found a issue on codelens/Resolve when starting server and opening `handlers.clj`:
```clojure
2022-07-24T03:16:30.773Z  ERROR [clojure-lsp.server:38] - Error receiving message: Internal error (-32603)
{:id 27540, :method "codeLens/resolve"}
com.oracle.svm.core.posix.thread.PosixJavaThreads.pthreadStartRoutine    PosixJavaThreads.java:  194
            com.oracle.svm.core.thread.JavaThreads.threadStartRoutine         JavaThreads.java:  597
                                                 java.lang.Thread.run              Thread.java:  829
                   java.util.concurrent.ThreadPoolExecutor$Worker.run  ThreadPoolExecutor.java:  628
                    java.util.concurrent.ThreadPoolExecutor.runWorker  ThreadPoolExecutor.java: 1128
                                                                  ...                               
                                    clojure.core.async/thread-call/fn                async.clj:  484
                                      clojure.core.async/pipeline*/fn                async.clj:  549
                                 clojure.core.async/pipeline*/process                async.clj:  534
                                                clojure.core.async/fn                async.clj:  172
              clojure.core.async.impl.channels.ManyToManyChannel/put!             channels.clj:  147
                             clojure.core.async.impl.channels/chan/fn             channels.clj:  304
                                              clojure.core/keep/fn/fn                 core.clj: 7393
                                         lsp4clj.server.ChanServer/fn               server.clj:  165
                                       lsp4clj.server/receive-message               server.clj:  108
                                    lsp4clj.server/receive-message/fn               server.clj:  113
                            lsp4clj.server.ChanServer/receive-request               server.clj:  214
                                                                  ...                               
                                                clojure-lsp.server/fn               server.clj:  268
                               clojure-lsp.handlers/code-lens-resolve             handlers.clj:  402
                      clojure-lsp.feature.code-lens/resolve-code-lens            code_lens.clj:   48
                                     clojure-lsp.shared/uri->filename               shared.clj:  166
                                          clojure.string/starts-with?               string.clj:  365
java.lang.NullPointerException: 
``` 

BTW, I'm using the static linux graal binary from CI

mainej:
Thanks for noticing that error @ericdallo. I'll look into it. Code lens gets me every time! I never remember to turn it on.

mainej:
Code lens resolution fixed in 28e6c558e412ad0be9966e49cae57dded4eabc39.

cli/resources/META-INF/native-image/clojure-lsp/clojure-lsp/proxy-config.json
```diff
@@ -1,3 +1,2 @@
 [
-  ["clojure_lsp.ClojureLanguageClient","org.eclipse.lsp4j.jsonrpc.Endpoint"]
 ]
```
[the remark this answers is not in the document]
ericdallo: Yes, but no problem keep it as if we need to add proxy configs the file is already there and one wouldn't need to understand how graal magically gets that from classpath

cli/resources/META-INF/native-image/clojure-lsp/clojure-lsp/reflect-config.json
```diff
@@ -17,49 +17,6 @@
   "name":"clojure.lang.Var",
   "allPublicMethods":true
 },
```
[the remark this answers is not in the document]
ericdallo: We certainly need to re-check that, we had many changes on deps since started using graal so there is a chance that we don't need more reflection configs in this file. Let's do in another PR though

cli/src/clojure_lsp/nrepl.clj
ericdallo: I see, I think we could be straightforward here and just have a `setup-dev-environment` for now.

lib/test/clojure_lsp/handlers_test.clj
```diff
@@ -277,3 +277,26 @@
                 {:text-document {:uri (h/file-uri "file:///b.clj")}
                  :context {:diagnostics []}
                  :range {:start {:line 1 :character 1}}})))))
+
+(deftest code-lens-can-be-resolved
```
ericdallo, line 281: Good!

ericdallo:
Got one more, not sure if already happens on master though

mainej:
@ericdallo do you have a reproduction for the drag exception? I've been playing with it but can't replicate the error. Even if it isn't related to this PR, it'd be good to make an Issue for it.

ericdallo:
@mainej I can't repro anymore on both master and this branch :( 
If I notice something again I let you know

ericdallo:
@mainej I've been testing this branch for some time and looks pretty stable and working!
I still got a exception related to didChangeWatchedFiles, could you take a look please?
image

I think we can merge this and keep testing on master until next release, it's working pretty well!

mainej:
I'm glad to hear it's been working @ericdallo. Thanks for testing it. Feel free to merge any time you want. It might be good to give the lsp4clj branch a tag (`1.0.0-rc1` or something), and then reference it with `:git/tag` in the 3 clojure-lsp deps.edn files. The sha works fine though, so that's optional.

Good find on that error. It's going to be hard to track down. For me it only happens in a Graal build when calling (elisp) `rename-file`. The error I get is slightly different than yours... for me it happens in `hover` not `did-change-watched-files`. But `hover` is trying to do the same thing—slurp a file that doesn't exist.

There are a lot of messages that go back and forth when a file is renamed. I'll plan to turn on tracing to see what's getting out of sync. Then I'll have to try to understand what's different about the new system. Anyway, I'll let you know what I find.

mainej:
A question about merging @ericdallo ... do you want to merge with a regular merge commit? Or squash into one commit? Or squash into a smaller number of larger commits? If it's the later, I can help.

ericdallo:
@mainej the latter sounds best to me

mainej:
> the latter sounds best to me

It sounds better to me too, but I was hoping you'd say it doesn't matter. 😅  It's might be hard to rebase. Probably worth it though... There's a narrative to the commits that explains some of the decisions and it would be a shame to lose. But there are also lots of commits that are mostly noise. I'll see what I can do.

> exception related to didChangeWatchedFiles

When you rename a file (say `old.clj` to `new.clj`) with elisp `rename-file`, lsp-mode sends clojure-lsp a `workspace/didChangeWatchedFiles` message that says the old file was deleted:

```json
{"changes": [{"uri": "file:///path/to/old.clj", "type": 3}]}
```

clojure-lsp [handles](https://github.com/clojure-lsp/clojure-lsp/blob/45ba142b53de7f0cdd02746c33c1e386a9bf73f4/lib/src/clojure_lsp/feature/file_management.clj#L281-L283) this by removing the file from the db, one step of which is deleting it from `:documents`. As will become clear in a moment, importantly this means we no longer have its text in `[:documents uri :text]`.

If you had the `old.clj` file open in an Emacs buffer, Emacs doesn't rename the buffer. (This surprises me, but I'm not trying to debug Emacs at the moment. 😄 ) So, on disk, `old.clj` has been deleted. `clojure-lsp` also has no record of `old.clj`. But Emacs still refers to `old.clj` in the buffer name. As you move around in that buffer, Emacs keeps generating LSP commands for that buffer's URI. If you do something that causes clojure-lsp to try to `slurp` the URI off disk, you get the exception.

One way, and for me the quickest way, to cause clojure-lsp to `slurp` is to invoke `hover`. That invokes `f.file-management/force-get-document-text` which, because the file's text isn't in the DB, tries to `slurp`. It looks like you were doing something else in Emacs besides `hover`, because you were hitting a [different `slurp`](https://github.com/clojure-lsp/clojure-lsp/blob/45ba142b53de7f0cdd02746c33c1e386a9bf73f4/lib/src/clojure_lsp/feature/file_management.clj#L278). Maybe `rename-file` again? Anyway, the problem is the same—trying to `slurp` a file that doesn't exist throws an exception.

So this leaves two questions. How are we going to avoid the exception? And why doesn't it happen on master?

To answer the second question first, I went back to testing on master. It turns out I _can_ reproduce the exception there, but only in a JVM build. (The exception looks different in the logs because lsp4j is printing it, instead of lsp4clj@v1, but it's there.) I hadn't seen it earlier because I was testing a Graal build, which doesn't produce the exception. That made me curious—why not? With some more tracing and logging I realized that in Graal by the time the `:changes` reach `f.file-management/did-change-watched-files`, they're always an empty vector. Therefore, `file-management` doesn't remove the file's text from the DB, and things like `hover` don't try to `slurp`. So, we don't get exceptions in a Graal build, but only because `did-change-watched-files` is broken in Graal.

`:changes` being an empty vector in Graal feels like a reflection/debeaning problem. Does that sound right to you? If so, then the good news is that removing `lsp4j` will fix `did-change-watched-files`. The bad news is that I think it means `did-change-watched-files` doesn't work, and probably never has. 😬

Anyway, in summary, we should have been seeing these exceptions all along, and we weren't only because of a bug in our Graal build.

This still leaves the other open question: once the Graal build is fixed (either by fixing reflection on master or by merging lsp2clj)—how are we going to avoid the exceptions? I think we have to be more careful about `slurp`ing. We can't assume that the file exists on disk. It might exist only as an Emacs buffer. Does that sound right to you? Do you want to fix it on master now, on lsp2clj now, or on master after lsp2clj is merged?

As a final observation, while watching the traces I noticed what might be another bug. When calling `rename-file`, first the client sends a `willRenameFiles`, saying it will rename `old.clj` to `new.clj`. clojure-lsp responds by saying that the `ns` should be renamed from `old` to `new` before the file is renamed. The client performs this change, and notifies the server that `old.clj` `didChange`. This kicks off an analysis of `old.clj`. While that's running, the client sends `didChangeWatchedFiles` to say that it deleted `old.clj`. The server removes the file from the db and publishes 0 diagnostics. But then the analysis completes and the diagnostics for `old.clj` (which are that the ns `new` doesn't match the file's name) are put into the db, and we publish 1 diagnostic for `old.clj`.

In the case of `rename-file` this is probably OK. The buffer for `old.clj` still exists (even though I think Emacs should have renamed it) so you see the diagnostic there. But if Emacs _had_ renamed the buffer (as I believe it does if you rename via `lsp-rename` on the ns) then the diagnostic would be on a non-existent buffer. I'm not sure what Emacs does in this case. Do you end up with a flycheck error that you can never resolve because it's for a file and buffer that don't exist? How do other editors behave in the same situation? What should clojure-lsp do to avoid the problem? Maybe wait to handle `didChangeWatchedFiles` until the analysis is done? Or maybe (!) make `didChange` synchronous?

Anyway, that's a separate bug for another day.

ericdallo:
> So, we don't get exceptions in a Graal build, but only because did-change-watched-files is broken in Graal.

Oh, that explains a lot why I noticed didChangeWatchedFiles is working only some times, probably only when I'm using a JVM build :)

> :changes being empty in Graal feels like a reflection/debeaning problem. Does that sound right to you? If so, then the good news is that removing lsp4j will fix did-change-watched-files. The bad news is that I think it means did-change-watched-files doesn't work, and probably never has. grimacing

Yep, makes sense, for graal probably never worked before, but it's a recent feature (some months ago was introduced properly)

> We can't assume that the file exists on disk. It might exist only as an Emacs buffer.

Correct. As LSP spec says, a URI not necessarily is on disk, as it can be a remote LSP connection, so we should improve that somehow on server side, if we manage to remove the slurps, that would be the best. I'd prefer to be fixed on master after lsp2clj is merged, WDYT?

> We can't assume that the file exists on disk. It might exist only as an Emacs buffer.

Yes, that's was a issue I thought we solved, but yeah, it's possible to happen.

> make didChange synchronous

This would be a critical change, had lots of issues in the past related to that, we would need to take care if we take this path.

Not sure yet what clojure-lsp would need to do, but a issue for that makes sense indeed.

ericdallo:
About the squash, feel free to make what is easiest for you, I don't care that much for that, it's just that if we need to debug a bug or something, finding a relevant commit with that change would be nice.

mainej:
> if we manage to remove the slurps, that would be the best.

It'd be nice to remove the slurps, but unfortunately we need them occasionally—for example when we're moving a function to a namespace that hasn't been opened yet with `didOpen`. 

> I'd prefer to be fixed on master after lsp2clj is merged, WDYT?

Sounds good.

> > make didChange synchronous

> This would be a critical change

I was mostly joking. Someday we could revisit the assumptions around why it needs to be asynchronous, but not now.

PR #1118, opened 2022-07-14
Document how to reset the project root.

thomascothran:
If the project root is not set correctly, lsp will ignore config files such as
`clj-kondo/config.edn`. Document how to check if this is the case and reset the
project root if necessary.


docs/troubleshooting.md
```diff
@@ -146,6 +146,7 @@ then those can be added at the project level.
 - clojure-lsp persist the external jars analysis in a `.lsp/.cache/` folder, if you have issues with some specific feature,
 try to remove that dir and restart the server.
 - If you have issues with macros, [double check your clj-kondo config](https://github.com/clj-kondo/clj-kondo/blob/master/doc/config.md#unrecognized-macros).
+- Check that the project root is set correctly with `lsp-clojure-server-info`. If it is incorrect, run `lsp-workspace-folders-remove`, then `lsp`.
```
ericdallo, line 149: `lsp-clojure-server-info` is specific for Emacs/lsp-mode, same for `lsp-workspace-folders-remove` and `lsp`.
Maybe we can be more generic here explaining to check the project root and remove and add the correct dir?

PR #1123, opened 2022-07-17
Fix namespace rename

### LINKED ISSUE #1121: Renaming a namespace results in the wrong name
**Describe the bug**
Request to rename `my.app.ui.foo` to `my.app.ui.reachability.foo`. This creates a new file named `my/app/ui/reachability/foo.clj` (correct) but sets the namespace to `my.app.ui.reachability.foochability.foo` (incorrect).

**Expected behavior**
The namespace should be `my.app.ui.reachability.foo`.

**Logs**
<details>
<pre>
[Trace - 11:43:29 AM] Sending request 'textDocument/prepareRename - (16327)'.
Params: {
  "textDocument": {
    "uri": "file:///Users/jmaine/workspace/opensource/lsp-rename-repro/src/cljs/my/app/ui/foo.cljs"
  },
  "position": {
    "line": 0,
    "character": 14
  }
}

[Trace - 11:43:29 AM] Received response 'textDocument/prepareRename - (16327)' in 13ms.
Result: {
  "start": {
    "line": 0,
    "character": 4
  },
  "end": {
    "line": 0,
    "character": 17
  }
}

[Trace - 11:43:41 AM] Sending request 'textDocument/rename - (16328)'.
Params: {
  "textDocument": {
    "uri": "file:///Users/jmaine/workspace/opensource/lsp-rename-repro/src/cljs/my/app/ui/foo.cljs"
  },
  "position": {
    "line": 0,
    "character": 14
  },
  "newName": "my.app.ui.reachability.foo"
}

[Trace - 11:43:41 AM] Received response 'textDocument/rename - (16328)' in 23ms.
Result: {
  "documentChanges": [
    {
      "textDocument": {
        "version": 128,
        "uri": "file:///Users/jmaine/workspace/opensource/lsp-rename-repro/src/cljs/my/app/ui/foo.cljs"
      },
      "edits": [
        {
          "range": {
            "start": {
              "line": 0,
              "character": 4
            },
            "end": {
              "line": 0,
              "character": 17
            }
          },
          "newText": "my.app.ui.reachability.foo"
        }
      ]
    },
    {
      "oldUri": "file:///Users/jmaine/workspace/opensource/lsp-rename-repro/src/cljs/my/app/ui/foo.cljs",
      "newUri": "file:///Users/jmaine/workspace/opensource/lsp-rename-repro/src/clj/my/app/ui/reachability/foo.cljs",
      "kind": "rename"
    }
  ]
}

[Trace - 11:43:41 AM] Sending notification 'textDocument/didSave'.
Params: {
  "textDocument": {
    "uri": "file:///Users/jmaine/workspace/opensource/lsp-rename-repro/src/cljs/my/app/ui/foo.cljs",
    "version": 130
  },
  "text": "(ns my.app.ui.reachability.foo)\n"
}

[Trace - 11:43:41 AM] Sending notification 'textDocument/didChange'.
Params: {
  "textDocument": {
    "uri": "file:///Users/jmaine/workspace/opensource/lsp-rename-repro/src/cljs/my/app/ui/foo.cljs",
    "version": 130
  },
  "contentChanges": [
    {
      "text": "(ns my.app.ui.reachability.foo)\n"
    }
  ]
}

[Trace - 11:43:41 AM] Sending notification 'textDocument/didClose'.
Params: {
  "textDocument": {
    "uri": "file:///Users/jmaine/workspace/opensource/lsp-rename-repro/src/cljs/my/app/ui/foo.cljs"
  }
}

[Trace - 11:43:41 AM] Sending request 'workspace/willRenameFiles - (16329)'.
Params: {
  "files": [
    {
      "oldUri": "file:///Users/jmaine/workspace/opensource/lsp-rename-repro/src/cljs/my/app/ui/foo.cljs",
      "newUri": "file:///Users/jmaine/workspace/opensource/lsp-rename-repro/src/clj/my/app/ui/reachability/foo.cljs"
    }
  ]
}

[Trace - 11:43:41 AM] Received response 'workspace/willRenameFiles - (16329)' in 12ms.
Result: {
  "documentChanges": [
    {
      "textDocument": {
        "version": 130,
        "uri": "file:///Users/jmaine/workspace/opensource/lsp-rename-repro/src/cljs/my/app/ui/foo.cljs"
      },
      "edits": [
        {
          "range": {
            "start": {
              "line": 0,
              "character": 4
            },
            "end": {
              "line": 0,
              "character": 17
            }
          },
          "newText": "my.app.ui.reachability.foo"
        }
      ]
    }
  ]
}

[Trace - 11:43:41 AM] Sending notification 'textDocument/didClose'.
Params: {
  "textDocument": {
    "uri": "file:///Users/jmaine/workspace/opensource/lsp-rename-repro/src/clj/my/app/ui/reachability/foo.cljs"
  }
}

[Trace - 11:43:41 AM] Sending notification 'textDocument/didChange'.
Params: {
  "textDocument": {
    "uri": "file:///Users/jmaine/workspace/opensource/lsp-rename-repro/src/cljs/my/app/ui/foo.cljs",
    "version": 131
  },
  "contentChanges": [
    {
      "text": "(ns my.app.ui.reachability.foochability.foo)\n"
    }
  ]
}

[Trace - 11:43:41 AM] Sending request 'textDocument/semanticTokens/range - (16330)'.
Params: {
  "textDocument": {
    "uri": "file:///Users/jmaine/workspace/opensource/lsp-rename-repro/src/clj/my/app/ui/reachability/foo.cljs"
  },
  "range": {
    "start": {
      "line": 0,
      "character": 0
    },
    "end": {
      "line": 1,
      "character": 0
    }
  }
}

[Trace - 11:43:41 AM] Sending notification 'textDocument/didOpen'.
Params: {
  "textDocument": {
    "uri": "file:///Users/jmaine/workspace/opensource/lsp-rename-repro/src/clj/my/app/ui/reachability/foo.cljs",
    "languageId": "clojurescript",
    "version": 131,
    "text": "(ns my.app.ui.reachability.foochability.foo)\n"
  }
}

</pre>
</details>

 - Editor Emacs

**Analysis**

I can see why this is happening, or at least partially. You can follow along in the logs above. It's an interaction between `textDocument/rename` and `workspace/willRenameFiles`.

In Emacs (other editors may follow a different sequence), first the client and server communicate to prepare for a rename.

Then, the client sends a `textDocument/rename` request, saying "I'm on the 14th character of the first line of 'my/app/ui/foo.cljs'. I want to rename the thing there to `my.app.ui.reachability.foo`."

The server responds by saying "OK, on the first line, replace chars 4-17 with `my.app.ui.reachability.foo`. Also rename the file 'my/app/ui/foo.cljs' to 'my/app/ui/reachability/foo.cljs'."

The client starts executing these commands. First it edits the old file to contain the new ns. It notifies the server of that change (twice, once with `textDocument/didSave` and once with `textDocument/didChange`). Then it starts the file rename. First it closes the old file. Then it sends the server a `workspace/willRenameFiles` request, saying that it is about to rename the old file to the new file.

And here's where the buggy text is introduced... The server responds by telling the client "on the first line of the old file, replace chars 4-17 with `my.app.ui.reachability.foo`". But this is wrong. The server seems to think that characters 4-17 are `my.app.ui.foo`, when they are actually `my.app.ui.rea` (followed by `chability.foo`). Effectively, the server is telling the client to apply the same edit for a second time. When the client follows the server's directions, it edits the old file. After replacing characters 4-17, the namespace ends up being `my.app.ui.reachability.foochability.foo`.

To finish off the sequence, the client tells the server about the change to the old file, including the buggy text. Then it renames the file on the filesystem (I presume; the logs don't actually say when the client does this). Then it opens the new file, and notifies the server, again including the buggy text.

So, the question is, why? During `workspace/willRenameFiles`, the server has looked in the analysis for the namespace of the old file and found `my.app.ui.foo`. That's why it think characters 4-17 should be replaced. Though the client told the server that the old file contains the new ns via `textDocument/didChange`, the server is still analyzing that change asynchronously when it receives `workspace/willRenameFiles`.

@ericdallo you rolled back the `swap!` which fixed it... I haven't watched the logs to analyze why, but I have a guess. That swap moves the analysis of 'my/app/ui/foo.cljs' to 'my/app/ui/reachability/foo.cljs' during the first request, `textDocument/rename`. Then in the second request `workspace/willRenameFiles` we have no analysis for 'my/app/ui/foo.cljs', and so the server doesn't generate any edits. This works, but not for the right reasons. That swap is definitely wrong, so it needs to be removed.

So what's a better fix? We should probably wait for analysis to finish before handling `workspace/willRenameFiles`. But maybe there's a better fix. In the first request `textDocument/rename`, instead of responding with both the text edits and the file rename, we could respond with just the file rename. Then we can rely on `workspace/willRenameFiles` to apply the edit. Or maybe not... I'm not sure all clients have will-rename capabilities. I'll think about it more.

mainej:
This fixes #1121. It also restores #1049, which was temporarily reverted as a quick fox for #1121.

When a user asks to rename a namespace, we instruct the client to edit the namespace and to rename the file. The client makes the edit (in the old file), then tell us via didChange. We start an asynchronous analysis. Then, the client sends us willRenameFiles. We respond by (incorrectly) asking the client to make the same edit to the old file we asked for before. The reason we do this is we haven't waited for the analysis to finish, so we think the file still has its old namespace.

This changes will-rename-files to wait for the analysis to finish. It also changes it to return no edits if the namespace has already been edited (which it has in this scenario, though it wouldn't if the file had been renamed because of something besides a rename refactoring).

Finally, it fixes another bug where we weren't clearing the diagnostics of deleted files. In my Emacs, rename refactorings still cause erroneous lint (we ask clj-kondo to analyze #.foo.clj, which doesn't exist, and it complains about that) but that might just be because of my Emacs config.

### DIFF
--- a/lib/src/clojure_lsp/feature/file_management.clj
+++ b/lib/src/clojure_lsp/feature/file_management.clj
@@ -263,6 +263,10 @@
       (shared/dissoc-in [:analysis filename])
       (shared/dissoc-in [:findings filename])))
 
+(defn ^:private file-deleted [db* uri filename]
+  (swap! db* db-without-file uri filename)
+  (f.diagnostic/publish-empty-diagnostics! uri @db*))
+
 (defn did-change-watched-files [changes db*]
   (doseq [{:keys [uri type]} changes]
     (case type
@@ -276,16 +280,14 @@
                                db*)))
       :deleted (shared/logging-task
                  :delete-watched-file
-                 (let [filename (shared/uri->filename uri)]
-                   (swap! db* db-without-file uri filename))))))
+                 (file-deleted db* uri (shared/uri->filename uri))))))
 
 (defn did-close [uri db*]
   (let [filename (shared/uri->filename uri)
         source-paths (settings/get @db* [:source-paths])]
     (when (and (not (shared/external-filename? filename source-paths))
                (not (shared/file-exists? (io/file filename))))
-      (swap! db* db-without-file uri filename)
-      (f.diagnostic/publish-empty-diagnostics! uri @db*))))
+      (file-deleted db* uri filename))))
 
 (defn force-get-document-text
   "Get document text from db, if document not found, tries to open the document"
@@ -303,7 +305,9 @@
        (keep (fn [{:keys [oldUri newUri]}]
                (let [old-filename (shared/uri->filename oldUri)
                      new-ns (shared/uri->namespace newUri db)
-                     ns-definition (q/find-namespace-definition-by-filename db old-filename)]
-                 (when ns-definition
-                   (f.rename/rename-element oldUri new-ns db ns-definition :rename-file)))))
+                     old-ns-definition (q/find-namespace-definition-by-filename db old-filename)]
+                 (when (and new-ns
+                            old-ns-definition
+                            (not= new-ns (name (:name old-ns-definition))))
+                   (f.rename/rename-element oldUri new-ns db old-ns-definition :rename-file)))))
        (reduce #(shared/deep-merge %1 %2) {:document-changes []})))
--- a/lib/src/clojure_lsp/feature/file_management.clj
+++ b/lib/src/clojure_lsp/feature/file_management.clj
@@ -298,13 +298,12 @@
 (defn did-save [uri db*]
   (swap! db* #(assoc-in % [:documents uri :saved-on-disk] true)))
 
-(defn will-rename-files [files db*]
-  (let [db @db*]
-    (->> files
-         (keep (fn [{:keys [oldUri newUri]}]
-                 (let [old-filename (shared/uri->filename oldUri)
-                       new-ns (shared/uri->namespace newUri db)
-                       ns-definition (q/find-namespace-definition-by-filename db old-filename)]
-                   (when ns-definition
-                     (f.rename/rename-element oldUri new-ns db* old-filename ns-definition :rename-file)))))
-         (reduce #(shared/deep-merge %1 %2) {:document-changes []}))))
+(defn will-rename-files [files db]
+  (->> files
+       (keep (fn [{:keys [oldUri newUri]}]
+               (let [old-filename (shared/uri->filename oldUri)
+                     new-ns (shared/uri->namespace newUri db)
+                     ns-definition (q/find-namespace-definition-by-filename db old-filename)]
+                 (when ns-definition
+                   (f.rename/rename-element oldUri new-ns db ns-definition :rename-file)))))
+       (reduce #(shared/deep-merge %1 %2) {:document-changes []})))
--- a/lib/test/clojure_lsp/features/file_management_test.clj
+++ b/lib/test/clojure_lsp/features/file_management_test.clj
@@ -5,7 +5,9 @@
    [clojure-lsp.shared :as shared]
    [clojure-lsp.test-helper :as h]
    [clojure.test :refer [are deftest is testing]]
-   [medley.core :as medley]))
+   [medley.core :as medley]
+   [clojure-lsp.feature.rename :as f.rename]
+   [clojure-lsp.queries :as q]))
 
 (h/reset-db-after-test)
 
@@ -117,7 +119,16 @@
         [{:type :created
           :uri h/default-uri}]
         db/db*)
-      (is (= h/default-uri (h/take-or-timeout mock-created-chan 500))))))
+      (is (= h/default-uri (h/take-or-timeout mock-created-chan 500)))))
+  (testing "deleted file"
+    (h/let-mock-chans
+      [mock-diagnostics-chan #'db/diagnostics-chan]
+      (f.file-management/did-change-watched-files
+        [{:type :deleted
+          :uri h/default-uri}]
+        db/db*)
+      (is (= {:uri h/default-uri, :diagnostics []}
+             (h/take-or-timeout mock-diagnostics-chan 500))))))
 
 (deftest var-dependency-reference-filenames
   (swap! db/db* medley/deep-merge {:settings {:source-paths #{(h/file-path "/src")}}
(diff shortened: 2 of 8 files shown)


lib/test/clojure_lsp/features/file_management_test.clj
```diff
@@ -5,7 +5,9 @@
    [clojure-lsp.shared :as shared]
    [clojure-lsp.test-helper :as h]
    [clojure.test :refer [are deftest is testing]]
-   [medley.core :as medley]))
+   [medley.core :as medley]
+   [clojure-lsp.feature.rename :as f.rename]
+   [clojure-lsp.queries :as q]))
```
ericdallo, line 10: I wonder how lint didn't catch this on `clean-ns` 🤔 maybe it's not running for tests...

lib/src/clojure_lsp/feature/file_management.clj
```diff
@@ -263,6 +263,10 @@
       (shared/dissoc-in [:analysis filename])
       (shared/dissoc-in [:findings filename])))
 
+(defn ^:private file-deleted [db* uri filename]
+  (swap! db* db-without-file uri filename)
+  (f.diagnostic/publish-empty-diagnostics! uri @db*))
```
ericdallo, line 268: That's a good pratice, we were really missing that, I noticed some time ago when I would do lots of renames and see old diagnostics

ericdallo: oh, we were just missing for file watch, I see

lib/test/clojure_lsp/features/file_management_test.clj
mainej: Good question... I'll run clean-ns manually, but it'd be nice for the lint CI to catch that.

ericdallo (approved):
Looks good, tested manually as well

One other approach would be to handle differently if client supports willRenameFiles since it's a pretty recent feature from LSP 3.17, probably only main editos have support for it, but, your approach looks better, we don't need to handle differently checking the capability, it's a little bit more complex though

lib/test/clojure_lsp/features/file_management_test.clj
mainej: Nm... you're too fast!

mainej:
Thanks for doing another test. If you'd like, I can research special behavior for clients that support willRenameFiles, but as you said, it's working so maybe isn't worth more effort. 🤷

ericdallo:
Yeah, it's not necessary, current approach looks good

lib/src/clojure_lsp/feature/file_management.clj
mainej: Should we implement [`workspace/didDeleteFiles`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didDeleteFiles)?

PR #1129, opened 2022-07-19
Add flake package default and overlay default

weihua-lu:
Just add a package default and overlay default in the flake output for easy flake consumption.


ericdallo (approved):
Looks good, thanks!

Issue #767, opened 2022-02-16
[graalvm] Support 22.0.0.2 for Windows


ericdallo:
For some reason, clojure-lsp compiled with latest graalvm that we are already using on Linux/Macos/aarch64, doesn't work for Windows.
We should debug and understand why since other projects (babashka for example) work normally with that version.

One thing we should probably do that should help is to make integration tests work with Windows, which is commented/not running ATM.

[GraalVM Build instructions](https://clojure-lsp.io/building/#graalvm)
Places we need to change after the fix: [release](https://github.com/clojure-lsp/clojure-lsp/blob/master/.github/workflows/release.yml#L358), [nightly build](https://github.com/clojure-lsp/clojure-lsp/blob/master/.github/workflows/nightly.yml#L284)

PR #1133, opened 2022-07-21
Choose PowerShell exec program on Windows based on availability

### LINKED ISSUE #1132: Classpath lookup failure error with Calva on MS-Windows
**Describe the bug**
Calva throws an error message when loading Clojure-lsp up on MS-Windows:

```
Classpath lookup failed when running `push -NoProfile -Command ...`
```

**To Reproduce**
On a stock Windows environment that comes with standard PowerShell 5.1 version and does not have any newer PowerShell version installed.

1. Open Calva in Visual Studio Code.
2. Open a workspace with an empty deps.edn file (or any other type of Clojure project file).
3. The above error will popup in Visual Studio Code.

**Expected behavior**
There should be no error message popping up.

 - OS: Windows 10
 - Editor VSCode (Calva)
 - Version: latest calva update

**Additional context**
There has been a recent change with #976 that appears to have forced all external program invocations on MS-Windows to go via the `pwsh` command which is only available with the latest version of PowerShell, but does not come with any MS-Windows version preinstalled.

The error is thrown when trying to scan the class path.

This is likely to occur with any editor, but I do not know how to run clojure-lsp standalone.

PR to follow.

ikappaki:
Hi,

can you please consider patch to fix an issue on MS-Windows that always tries to invoke `pwsh.exe` to scan the classpath, even if it is not installed. It fixes #1132.

`clojure-lsp` invokes one of `Clojure`, `lein`, `boot` or `bb` executables on *nix to retrieve a project's class path.  Invoking these programs on MS-Windows by this name is not always successful, if their extension is not `.exe` or `.bat`. In that case they have either to be invoked with their name+extension (e.g. `npx.cmd`) or if they are PowerShell `.ps1` scripts (as with Clojure cli tools or some installation of `lein`) they have to be invoked with a PowerShell executable (either `powershell.exe` that is preinstalled on Windows or `pws.exe` if newer versions of PowerShell is installed by the user).

Currently `clojure-lsp` tries to invoke all project executables using `pwsh` which is not preinstalled with MS-Windows, and thus fails.

This patch either adds the extension to the project executable so that can be invoked, or, in the case of a powershell script, tries to locate the PowerShell executable installed on the system and uses that to invoke it.

For demo purposes, I've created tests that confirm the good usage across all different type of projects and architectures, though I understand that this is quite a lot. The tests will be skipped if the project executable is not installed (e.g. `boot`), but will always run on GitHub. I've named the action  `cross platform tests`, although it runs only these new tests under `classpath`. Currently the clojure-lsp test suite does not pass under MS-Windows, and I plan at some point to make everything work on Win, and perhaps add them in this action. Feel free to ask it to be removed.

Thanks

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -77,40 +77,6 @@ jobs:
           path: clojure-lsp
           name: clojure-lsp
 
-  build-cli-jar-win:
-    if: startsWith(github.ref, 'refs/tags/2')
-    name: Build JVM cli win
-    runs-on: windows-2022
-    steps:
-      - uses: actions/checkout@v2.2.0
-        with:
-          fetch-depth: 0
-      - name: Prepare java
-        uses: actions/setup-java@v1
-        with:
-          java-version: 11
-
-      - name: Install Clojure
-        uses: DeLaGuardo/setup-clojure@master
-        with:
-          cli: '1.10.3.1013'
-
-      - name: Get latest tag
-        id: latest-tag
-        uses: WyriHaximus/github-action-get-previous-tag@v1
-
-      - name: Generate jar for native
-        run: |
-          cd cli
-          clojure -T:build prod-jar-for-native
-          mv target/clojure-lsp-standalone.jar ../clojure-lsp-standalone-win.jar
-
-      - name: Upload jar
-        uses: actions/upload-artifact@v2
-        with:
-          path: clojure-lsp-standalone-win.jar
-          name: clojure-lsp-standalone-win.jar
-
   linux-amd64:
     name: Build native linux amd64 binary
     needs: [build-lib-jar, build-cli-jar]
@@ -356,13 +322,13 @@ jobs:
 
   windows:
     name: Build native Windows binary
-    needs: [build-lib-jar, build-cli-jar-win]
+    needs: [build-lib-jar, build-cli-jar]
     runs-on: windows-2022
     steps:
       - uses: actions/checkout@v2
       - uses: actions/download-artifact@v2
         with:
-          name: clojure-lsp-standalone-win.jar
+          name: clojure-lsp-standalone.jar
 
       - name: configure Pagefile
         uses: al-cheb/configure-pagefile-action@v1.2
@@ -403,7 +369,7 @@ jobs:
 
       - name: Build Windows native image
         env:
-          CLOJURE_LSP_JAR: ..\clojure-lsp-standalone-win.jar
+          CLOJURE_LSP_JAR: ..\clojure-lsp-standalone.jar
           CLOJURE_LSP_XMX: "-J-Xmx7g"
         run: |
           cd cli
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -77,6 +77,40 @@ jobs:
           path: clojure-lsp
           name: clojure-lsp
 
+  build-cli-jar-win:
+    if: startsWith(github.ref, 'refs/tags/2')
+    name: Build JVM cli win
+    runs-on: windows-2022
+    steps:
+      - uses: actions/checkout@v2.2.0
+        with:
+          fetch-depth: 0
+      - name: Prepare java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+
+      - name: Install Clojure
+        uses: DeLaGuardo/setup-clojure@master
+        with:
+          cli: '1.10.3.1013'
+
+      - name: Get latest tag
+        id: latest-tag
+        uses: WyriHaximus/github-action-get-previous-tag@v1
+
+      - name: Generate jar for native
+        run: |
+          cd cli
+          clojure -T:build prod-jar-for-native
+          mv target/clojure-lsp-standalone.jar ../clojure-lsp-standalone-win.jar
+
+      - name: Upload jar
+        uses: actions/upload-artifact@v2
+        with:
+          path: clojure-lsp-standalone-win.jar
+          name: clojure-lsp-standalone-win.jar
+
   linux-amd64:
     name: Build native linux amd64 binary
     needs: [build-lib-jar, build-cli-jar]
@@ -87,13 +121,13 @@ jobs:
         with:
           name: clojure-lsp-standalone.jar
 
-      - name: Install Clojure
+      - name: Install clojure build tools
         uses: DeLaGuardo/setup-clojure@master
         with:
           cli: '1.10.3.1013'
-
-      - name: Install Babashka
-        run: curl -s https://raw.githubusercontent.com/borkdude/babashka/master/install | sudo bash
+          lein: '2.9.1'
+          boot: '2.8.3'
+          bb: '0.9.161'
 
       - name: Install GraalVM
         uses: DeLaGuardo/setup-graalvm@master
@@ -145,13 +179,13 @@ jobs:
         with:
           name: clojure-lsp-standalone.jar
 
-      - name: Install Clojure
+      - name: Install clojure build tools
         uses: DeLaGuardo/setup-clojure@master
         with:
           cli: '1.10.3.1013'
-
-      - name: Install Babashka
-        run: curl -s https://raw.githubusercontent.com/borkdude/babashka/master/install | sudo bash
+          lein: '2.9.1'
+          boot: '2.8.3'
+          bb: '0.9.161'
 
       - uses: graalvm/setup-graalvm@v1
         with:
@@ -277,13 +311,13 @@ jobs:
         with:
           name: clojure-lsp-standalone.jar
 
-      - name: Install Clojure
+      - name: Install clojure build tools
         uses: DeLaGuardo/setup-clojure@master
         with:
           cli: '1.10.3.1013'
-
-      - name: Install Babashka
-        run: curl -s https://raw.githubusercontent.com/borkdude/babashka/master/install | sudo bash
+          lein: '2.9.1'
+          boot: '2.8.3'
+          bb: '0.9.161'
 
       - name: Install GraalVM
         uses: DeLaGuardo/setup-graalvm@master
@@ -322,13 +356,13 @@ jobs:
 
   windows:
     name: Build native Windows binary
-    needs: [build-lib-jar, build-cli-jar]
+    needs: [build-lib-jar, build-cli-jar-win]
     runs-on: windows-2022
     steps:
       - uses: actions/checkout@v2
       - uses: actions/download-artifact@v2
         with:
-          name: clojure-lsp-standalone.jar
+          name: clojure-lsp-standalone-win.jar
 
       - name: configure Pagefile
         uses: al-cheb/configure-pagefile-action@v1.2
@@ -341,14 +375,13 @@ jobs:
         with:
           java-version: 11
 
-      - name: Install Clojure
-        run: |
-          iwr -useb download.clojure.org/install/win-install-1.10.3.1013.ps1 | iex
-
-      - name: Install Babashka
-        uses: turtlequeue/setup-babashka@v1.3.0
+      - name: Install clojure build tools
+        uses: DeLaGuardo/setup-clojure@master
         with:
-          babashka-version: 0.7.8
+          cli: '1.10.3.1013'
+          lein: '2.9.1'
+          boot: '2.8.3'
+          bb: '0.9.161'
 
       - name: Install MSVC
         uses: ilammy/msvc-dev-cmd@v1
@@ -370,7 +403,7 @@ jobs:
 
       - name: Build Windows native image
         env:
-          CLOJURE_LSP_JAR: ..\clojure-lsp-standalone.jar
+          CLOJURE_LSP_JAR: ..\clojure-lsp-standalone-win.jar
           CLOJURE_LSP_XMX: "-J-Xmx7g"
         run: |
           cd cli
@@ -383,10 +416,10 @@ jobs:
         with:
           file: clojure-lsp.exe
 
-      # Windows return the json in the same line but for some reason clients seems to accept that.
-      # - name: Run integration tests
-      #   run: |
-      #     bb integration-test/run-all.clj .\clojure-lsp.exe
+      - name: Run integration tests
+        run: |
+          cd cli
+          bb integration-test ../clojure-lsp.exe
 
       - name: Zip binary
         run: |
--- a/cli/build.clj
+++ b/cli/build.clj
@@ -1,8 +1,11 @@
 (ns build
   (:require
+   [babashka.fs :as fs]
+   [babashka.process :as p]
    [clojure.java.io :as io]
    [clojure.string :as string]
-   [clojure.tools.build.api :as b]))
+   [clojure.tools.build.api :as b]
+   [clojure.xml :as xml]))
 
 (def lib 'com.github.clojure-lsp/clojure-lsp)
 (def clojars-lib 'com.github.clojure-lsp/clojure-lsp-standalone)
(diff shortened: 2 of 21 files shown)


lib/test/clojure_lsp/classpath_test.clj
```diff
@@ -46,3 +48,116 @@
              {:project-path "bb.edn"
               :classpath-cmd ["bb" "print-deps" "--format" "classpath"]}]
             (classpath/default-project-specs #{:something :otherthing}))))))
+
+(defmacro never-skip-exec-tests
```
ericdallo, line 52: I don't understand why we need all those macros and a new test suite, can't we just have integration tests for that? we already run integration tests on all platforms

ericdallo (approved):
The source change looks good, but I think we can change those complex tests to behave like integration ones

ericdallo:
Could you add a entry to changelog as well?

ericdallo:
FYI @bpringe

lib/test/clojure_lsp/classpath_test.clj
ikappaki: Sure, could you please point me to the relevant yaml file where integration test occur across all architectures so I can have a look and fit this in?

My logic for a separate file and the use of macros was this:
1. I considered these tests to be testing a single function in `lib`, thus I categorised them as unit tests.
2. I couldn't find anything relevant in `ci.yml` that would accommodate all existing unit tests + this one to run across all architectures; if I run the unit tests on MS-Windows (e.g. `cd lib && Clojure -M:test`), some tests fail (I plan to have a look at some point and fix them). Thus I decided to package this on a separate file so I can demonstrate the logic in a clear manner and receive feedback.
3. The tests are subject to the project executables to be installed on the local machine where the tests are run and being discoverable, they will otherwise fail.
3.1. I wanted users that run the test suite locally, not to fail these tests if the project exec (e.g. boot) was not locally installed, but skip it with a message. Thus, the user does not have to worry about an irrelevant test failing.
3.2. Likewise, I wanted the tests to properly fail in all circumstances when run in the CI environment and this can be done with the use of an env variable that is only set in the GH action.
3.3. And thus the use of macros to skip tests (there doesn't appear to be such a facility available just with `deftest) and test for an environment variable being set.

Thanks

ikappaki:
> The source change looks good, but I think we can change those complex tests to behave like integration ones

Sure, as per my review comment https://github.com/clojure-lsp/clojure-lsp/pull/1133#discussion_r927669034, I considered these to be unit test, but I suppose we can accommodate them as integration tests with the entry point being a higher level function.

Would it be possible to provide some guidance how the existing test in this PR calling `classpath/scan-classpath!` can be eventually called using a higher level integration fn? I've noticed integration tests are invoked by `release.yml` and live under `cli/integration` but not sure how to write a test to eventually invoke `classpath/scan-classpath!` from within the `cli` module. It is my first time looking into clojure-lsp source code and have only looked thus far in the code :)

btw, are integration tests current for all architectures? the windows one appears to be disabled:
```yaml
     # Windows return the json in the same line but for some reason clients seems to accept that.
      # - name: Run integration tests
      #   run: |
      #     bb integration-test/run-all.clj .\clojure-lsp.exe
```

thanks

lib/test/clojure_lsp/classpath_test.clj
ericdallo: This is where we run the [integration-tests](https://github.com/clojure-lsp/clojure-lsp/blob/master/.github/workflows/ci.yml#L75), but I just noticed we don't run windows ones on every PR, maybe we should 🤔 I think it makes sense.

ericdallo: I think we can keep it as unit tests too, but mock the existence of the commands on their machines so, otherwise those should be considered integration.
If so, we could remove the env var logic I think

ericdallo:
The integration tests are pretty high level, they don't call clojure but call the generated clojure-lsp binary manually from shell and test sending and receiving messages, like a real LSP client.
For this issue, we would need to create a new integration-test [here](https://github.com/clojure-lsp/clojure-lsp/tree/master/cli/integration-test/integration) and link it on [entrypoint](https://github.com/clojure-lsp/clojure-lsp/blob/master/cli/integration-test/entrypoint.clj#L7-L31).

But on a second though, I liked the idea to keep it as unit tests but mock external programs responses if we find a way, since what we want to test here is if we are calling the commands following a specific standard for each OS. Even so, we don't have any guarantee for cases like, Windows that don't have pwsh and cases like that

> btw, are integration tests current for all architectures? the windows one appears to be disabled:

Yep, this is something we need to fix, since you are a Windows user, your help would be very welcome :) 
BTW, there is that and this [issue](https://github.com/clojure-lsp/clojure-lsp/issues/767) I can't find the root cause in case you wanna take a look :)

ikappaki:
> The integration tests are pretty high level, they don't call clojure but call the generated clojure-lsp binary manually from shell and test sending and receiving messages, like a real LSP client. For this issue, we would need to create a new integration-test [here](https://github.com/clojure-lsp/clojure-lsp/tree/master/cli/integration-test/integration) and link it on [entrypoint](https://github.com/clojure-lsp/clojure-lsp/blob/master/cli/integration-test/entrypoint.clj#L7-L31).
> 
> But on a second though, I liked the idea to keep it as unit tests but mock external programs responses if we find a way, since what we want to test here is if we are calling the commands following a specific standard for each OS. Even so, we don't have any guarantee for cases like, Windows that don't have pwsh and cases like that

I need to think how to best mock that to get all invocation combinations possible. I'll probably need to refactor the class path code a little more.

In addition to that, I also like the idea of an integration tools with the actual tooling installed. If I were to write an integration test, and given an empty project file (say an empty temp-dir/deps.edn file), would it be possible to send a simple message to the lsp binary to return the project's class path? I assume this would eventually invoke by necessity the Clojure cli tool. I couldn't find an example in the integration tests to ask for project specific information, they all seem to centred around `.clj` files.

> > btw, are integration tests current for all architectures? the windows one appears to be disabled:
> 
> Yep, this is something we need to fix, since you are a Windows user, your help would be very welcome :) BTW, there is that and this [issue](https://github.com/clojure-lsp/clojure-lsp/issues/767) I can't find the root cause in case you wanna take a look :)

I can have a look at this, and most importantly I plan fixing the unit test suite to run on windows. It will have to wait until next month though :)

Thanks

ericdallo:
> In addition to that, I also like the idea of an integration tools with the actual tooling installed. If I were to write an integration test, and given an empty project file (say an empty temp-dir/deps.edn file), would it be possible to send a simple message to the lsp binary to return the project's class path? I assume this would eventually invoke by necessity the Clojure cli tool. I couldn't find an example in the integration tests to ask for project specific information, they all seem to centred around .clj files.

The current integration tests uses a `sample-test` folder which is a sample project, when LSP initializes, it scans the classpath, you can test if it scanned correctly calling the `serverInfo/raw` command which should return lots of information including the `:classpath`

ikappaki:
Hi @ericdallo,

I have finally found the time to refactor the code so that the tests can be mocked, sorry for the delay.

There are two new functions `locate-executable` and `shell` that are mapping to `fs.babashka/which` and `clojure.java.shell\sh` respectively, whose purpose is to be mocked during testing with static response so that the expected outcome is fixed in the tests.

The project executable (lein, clojure, bb etc) are now expanded to their absolute paths by clojure-lsp. This provides us with the option to try and locate the right command to invoke them based on run time availability.

on MS-Windows,
1. `lein` can be installed as an executable or a PowerShell module.
2. `clojure` can be installed as a PowerShell module only, that can only be invoked from PowerShell.
3. There are two PowerShell executables (`powershell.exe` and `pwsh.exe`), and each may have their own modules installed.

With this patch, native executable or native scripts are preferred over PowerShell and `powershell.exe` is preferred over `pwsh.exe` when project executable is installed on both.

I've tried to capture all the above cases in testing.

I've now tested this to work with `Calva`.

> The current integration tests uses a `sample-test` folder which is a sample project, when LSP initializes, it scans the classpath, you can test if it scanned correctly calling the `serverInfo/raw` command which should return lots of information including the `:classpath`

Could you please provide me with an example how to do this? I thought this would have been as simple as the following, but comes back as an unsupported command:

``` clojure
(lsp/start-process!)
(lsp/request! [:serverInfo/raw {}])
```

(Also, the linter seems to be failing on unrelated changes.  I might fix it on a follow up commit).

ericdallo:
Thanks for the update @ikappaki, it makes sense to me your explanation, I didn't review the new code yet though, but should do soon!

> Could you please provide me with an example how to do this? 

It should be pretty similar to [cursor-info-test](https://github.com/clojure-lsp/clojure-lsp/blob/master/cli/integration-test/integration/cursor_info_test.clj#L19), but it should be something like:
```clojure
["clojure/serverInfo/raw" {}}]
```

ikappaki:
Hi @ericdallo,

I have just provided a classpath integration test over all supported build tools (bb, boot, clojure cli, lein and shadow).

I did this by adding a minimalist sample test file with a simple dependency for each of the build tools, and check whether it is included in the `clojure/serverInfo/raw` output.

Unfortunately on windows, there are several issues run the existing integration tests that I plan to resolve next. Though I thought it is important to keep at least this test for windows, as a proof that the fix works.

For the jvm integration test to work on windows, I used the `launch4j` to build the clojuer-lsp executable, since `deps-bin` only appears to support *nix.  I think running at least a single instance of the integration tests for windows is very important, so I've added a new entry just for Java 8 on MS-Windows in `ci.yml`.

Running the integration tests on the native image revealed an unrelated issue with building the standalone jar on Linux while compiling the native image on windows: some variables (such as path separators which are different on Linux vs MS-Windows) were coming back with their Linux values instead of the MS-Windows ones (e.g. the `babashka.fs/path-separator` was coming back as `:` instead of `;`), so I've decided to build the standalone jar also on Windows. As such `release.yml` has an additional entry to build native cli on windows.

I had two occasions where the server timed out while running the integration tests. It once happened for the `boot` project on Linux and most recently for the `shadow-cljs` project on MS-Windows. I thus increased the server timeout from 30secs to 60secs and now 90secs. I am not sure if this is a genuine issue with the server or just it takes more time for the project build tools to bootstrap. I will keep an eye at it.

Other than that there a few integration tests that have been failing on MS-Windows. I fixed a couple but the rest are increasingly difficult to figure out at my level with my little knowledge of the codebase. I think I will take a step back and try to fix the unit tests first and then move onto the integration tests. I will open an issue for this.

Thanks

(Not sure what is the issue with the `Dev release (macos-aarch64)` failing. It seems to be running on `Cirrus` but not sure how to grab of the log file to investigate)

ikappaki:
(Opened #1211 to address the dev/test support issues on MS-Windows.)

.github/workflows/release.yml
```diff
@@ -322,13 +356,13 @@ jobs:
 
   windows:
     name: Build native Windows binary
-    needs: [build-lib-jar, build-cli-jar]
+    needs: [build-lib-jar, build-cli-jar-win]
```
ericdallo, line 359: Do we really need to build a jar with windows just to use for windows graal build? a jar compiled on any env should be compatible with any JVM in any OS right?

cli/build.clj
```diff
@@ -49,13 +52,74 @@
              :main 'clojure-lsp.main
              :basis basis})))
 
-(defn ^:private bin [opts]
+(defn ^:private l4j-xml
```
ericdallo, line 55: Not sure it's a good idea to add all this complexity about launch4j, can't we make deps-bin work for windows as well? I'm the maintainer of that lib so issue and PRs are very welcome, WDYT?

ericdallo (approved):
@ikappaki Looks good thanks for the effort on that, it should help a lot solve and avoid issues on windows :)
My biggest concern is the extra complexity of launch4j and the need to build a jar but on windows OS, LMK what you think

.github/workflows/release.yml
ikappaki: The final executable is behaving erratically when I create the uber jar on Linux and compile natively on windows. 

I've  traced it down to the following `babashka.fs/path-separator` constant, which ends up with the Linux value (`:`), instead of the windows one (`;`):

https://github.com/babashka/fs/blob/60aea913a99a39b430a3666b7f5d116463c640c4/src/babashka/fs.cljc#L239

``` clojure
(def ^:const path-separator File/pathSeparator)
```

I suppose this expected because the constant is set at when compiling to java and thus takes the value of the compiling platform rather than the runtime one?

https://github.com/clojure/clojure/blob/master/changes.md

```
^:const lets you name primitive values with speedier reference.

    (def constants
     {:pi 3.14
      :e 2.71})

    (def ^:const pi (:pi constants))
    (def ^:const e (:e constants))

The overhead of looking up :e and :pi in the map happens at compile time, as (:pi constants) and (:e constants) are evaluated when their parent def forms are evaluated.
```

This has the bad effect that when `babashka.fs` is trying to split the `PATH` env variable on windows, it uses the wrong path separator and as such can't locate executable programs in the path.

I thought there could be more cases like this in external libraries, and in the interest of time and effort, to proceed with the sledgehammer approach of compiling to jar on the same platform as native compilation.

Let me know if you want me to investigate further and reproduce a general test case outside of the above observation. 

(cc'ing @borkdude for awareness)

cli/build.clj
ikappaki: This will need some more time and effort to investigate whether it is possible `deps-bin` to support windows.

From a cursory look that I had, `deps-bin` appears to create a shell script that invokes the embedded jar code. Not sure how to translate this into a windows batch file.

The `lanuch4j` concept is not so complex btw, it just creates an xml file with the actual input, output and jre locations and invokes `launch4jc.exe` on it. 

Let me know how you want to proceed, trying to replicate what `deps-bin` on Windows looks like a task on its own which perhaps we can look separately?

borkdude:
Would it help if babashka fs removed `:const`?

cli/build.clj
borkdude: Would it help if babashka fs removed :const?

.github/workflows/release.yml
borkdude: Would it help if babashka fs removed :const?

borkdude: I'll fix it now.

borkdude: Try `0.1.11`

ikappaki: Thanks @borkdude 

I've updated `babashka.fs` to v0.1.11 and have tested it to work, the windows integration test now passes.

Just for the record, the "cross compilation" is still susceptible to this type of errors since we don't control what goes in the external libraries. I suppose the expectation here is that such errors will be either caught up by users or during testing (currently there is very limited set for windows with this PR though there is a plan with #1211 to align the windows test coverage with Linux).

cli/build.clj
ericdallo: We can proceed but i'd like to have this supported on deps-bin, LMK if you wanna help with that

ikappaki: > We can proceed but i'd like to have this supported on deps-bin, LMK if you wanna help with that

Sure, I will try to have a look as part of #1211 and see how far we can go with it on windows, thanks!

ikappaki:
> @ikappaki Looks good thanks for the effort on that, it should help a lot solve and avoid issues on windows :) My biggest concern is the extra complexity of launch4j and the need to build a jar but on windows OS, LMK what you think
> 
> As an alternative I'm ok removing that and leaving to be solved on #1211

Hi @ericdallo,

as per review comments:
1. The GH job to build the jar on MS-Windows has been removed. The error was fixed in `babashka.fs`.
2. A commitment has been made to investigate as part of #1211 whether `deps-bin` can be updated to accommodate building an executable, and thus replacing `launch4j`.

I have also added a changelog entry, so unless there are any further concerns, we are good to go with the merge.

For completion, I've noticed two glitches with the continuous integration that we should be aware off (hopefully they were one off):
1. The classpath integration test jobs timed out when executing the build tools. Hopefully this was fixed by increasing the test client timeout interval to 90secs.
2. The chocolately windows package manager failed to download `launch4j` because it couldn't connect to sourceforge.com. Hopefully this was a GH issue with their infrastructure and won't happen again.

Thanks

ericdallo (approved):
Amazing work @ikappaki thank you very much for your help!

ericdallo:
I'll keep an eye on windows builds on nexts PRs release and let you know

PR #1135, opened 2022-07-21
Do not check for blocking in async/go in prod CLI

mainej:
Though still check for it in the dev CLI.


mainej:
I'm not sure why [this integration test failed](https://github.com/clojure-lsp/clojure-lsp/runs/7454782839?check_suite_focus=true). It looks like all the tests in that file passed, but we blocked waiting for the client to shut down, which it never did. I'm not inclined to research more becuase that code is changing with #1117.

For this PR, I'm just going to re-run the test.

ericdallo (approved):

Issue #1136, opened 2022-07-21
Extract def


mainej:
**Is your feature request related to a problem? Please describe.**
It's repetitive to extract an expression into a top-level `def`.

**Describe the solution you'd like**
I'd like a code action and command that introduce a `def`. Like `extract-function`, `create-function` and `promote-function`, it should operate on the form under the cursor, converting it to a private top-level def.

The command should accept the name of the var as an argument. If the command is invoked from the code-actions menu, it should get a default name, perhaps `new-def` or `new-value`, to match the `new-function` you get by default with `extract-function` and `promote-fn`. At their discrettion editors should be able to call the command directly, providing a user-defined name.

clj-refactor calls this refactoring [extract def](https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-extract-def). I think we should use the `ed` mnemonic. 

clj-refactor does a find-and-replace on the whole file, substituting the new var name wherever it can. I'm not sure whether this should be part of the clojure-lsp version. It's a cool feature when it's what you wanted, but it's annoying to undo when it isn't.

Assuming we don't find-and-replace, I would prefer that the new def be placed directly above the current top-level form. If we find-and-replace, the new def could go at the top of the file (as clj-refactor always does), but I would prefer that it be placed directly above the first top-level form that had a replacement.

**Describe alternatives you've considered**
You can kind of hack this by using `extract-function`, which creates a no-arg function. Then you have to manually change the `defn` to a `def`, remove the empty arg vector, and remove the parens from the call site. Usually this doesn't save much time over doing the extraction by hand.

PR #1138, opened 2022-07-21
Refactoring to extract a definition

mainej:
This is the basic version of implementing #1136. It does not do find-and-replace.


lib/src/clojure_lsp/feature/code_actions.clj
```diff
@@ -155,6 +155,13 @@
              :command   "extract-function"
              :arguments [uri line character "new-function"]}})
 
+(defn ^:private extract-definition-action [uri line character]
+  {:title   "Extract definition"
```
ericdallo, line 159: WDYT about `extract do def`? seems more clear to me what this code action will do

lib/src/clojure_lsp/refactor/transform.clj
```diff
@@ -453,6 +453,32 @@
...
+  (boolean (extract-definition-params zloc)))
+
+(defn extract-definition [zloc def-name]
+  (when-let [{:keys [zloc form-loc]} (extract-definition-params zloc)]
+    (let [expr-node (z/node zloc)
+          expr-meta (meta expr-node)
+
+          def-name (or def-name "new-value")
+          expr-edit (z/of-node (symbol def-name))
+          def-loc (-> (format "(def ^:private %s\n  )" def-name)
```
ericdallo, line 475: IIRC extract function checks for the `:use-metadata-for-privacy?` setting which is actually disable by default, I think we should check of that and use `def-` if not enabled instead

lib/src/clojure_lsp/feature/code_actions.clj
mainej: Sounds good—that is much clearer.

lib/src/clojure_lsp/refactor/transform.clj
mainej: I thought that too at first, but then I remembered—there's no such thing as `def-`!

lib/src/clojure_lsp/feature/code_actions.clj
mainej: Done.

lib/src/clojure_lsp/refactor/transform.clj
ericdallo: Oh, great point 😂

ericdallo (approved):
@mainej Feel free to solve conflicts and merge, thanks!

Issue #905, opened 2022-04-03
Destructure keys


snoe:
**Is your feature request related to a problem? Please describe.**
Destructuring keys after a number of `get`s have already been used can be a chore.

**Describe the solution you'd like**
Following clj-refactor would be a decent start.
https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-destructure-keys

Ideally:
Could destructure namespaced keywords to `::keys` / `::alias/keys` style. 
Handle `:syms` and `:strs` as well.

**Describe alternatives you've considered**
This could cycle destructuring/non-destructuring.

**Additional context**
More than top level destructuring should probably not occur by default but be behind a setting if supported.
`[params (get req :params) param1 (get params :param1)]` `{{:keys [param1]} :params}`

ericdallo:
Sounds like a good refactoring! Liked the name too.

> More than top level destructuring should probably not occur by default but be behind a setting if supported.
[params (get req :params) param1 (get params :param1)] {{:keys [param1]} :params}

Could you elaborate a little bit this? with a more concise example and why a setting would be good?

snoe:
You could imagine someone wanting to do this, I believe it is generally frowned upon nest destructurings.

```clojure
(defn handler [req]
  (let [params (get req :params) 
         param1 (get params :param1)]
    (inc param1)))
=>
(defn handler [{{:keys [param1]} :params}]
  (inc param1))
```

ericdallo:
I see, a setting to allow nested destructuring, yeah, I'm not sure which one would good to be the default, maybe the not nested one indeed, makes sense!

ericdallo:
Isn't that done @mainej ?

mainej:
@ericdallo, yes, I think this could be closed. A refactoring named `destructure-keys` was implemented in #1139.

A few features were discussed here that weren't implemented in #1139. We could make separate tickets if those things are still needed:

* `destructure-keys` doesn't destructure `(get foo :bar)`. However, because of #1175, this can be converted to `(:bar foo)`, which can then be destructured.
* `destructure-keys` doesn't encourage nested destructuring. That said, it can be coerced into doing it. It won't help you write `(let [{|bar :bar} foo] (:baz bar))`, but if you do, then calling `destructure-keys` will lead to `(let [{{:keys [baz]} :bar} foo] baz)`

Issue #904, opened 2022-04-03
Transforms for get, assoc, update & *-in forms


snoe:
**Is your feature request related to a problem? Please describe.**
Often things change and what used to be deeply nested is at the root, and vice versa. Manually changing code from one form to the other can be cumbersome. Maybe `transform-path-fn [tp]` for a name? It would cycke back and forth between forms.

Simplest
`(get m k)` <=> `(get-in m [k])`
`(assoc m k v)` <=> `(assoc-in m [k] v)`
`(update m k f f-args)` <=> `(update-in m [k] f f-args)`

If `k` is a keyword
`(:k m)` <=> `(get-in m [:k])`

Potentially support multiple gets
`(:k (:j m))` <=> `(get-in m [:j :k])`

Potentially work in thread-first
`(-> m :j :k inc)` <=> `(-> m (get-in [:j :k) inc)`
`(-> m :j :k)` <=> `(get-in m [:j :k])`

**Potential other solutions**
Instead of a cycling transform, it could be more like `thread` (one/all) and `unwind` (one/all). But I don't think I've ever wanted intermediate steps like this.

`(:k (:j m))` =in-one> `(:k (get-in m [:j])` =in-one> `(get-in m [:j :k])`
`(:k (:j m))` =in-all> `(get-in m [:j :k])`
`(get-in m [:j :k])` =?-all>`(:k (:j m))`

ericdallo:
Sounds good. the refactor name is a little bit weird to me, maybe we are trying to fit mnemonics and making the name more confusing that it could be?
Since this is a `cycle` operation, it could be cycle-path or something? the path name sounds odd still to me

I like the intermediate since we can leave to users decide if they want to stop or keep applying, while the "apply all" version doesn't support that.

snoe:
> Sounds good. the refactor name is a little bit weird to me, maybe we are trying to fit mnemonics and making the name more confusing that it could be? Since this is a `cycle` operation, it could be cycle-path or something? the path name sounds odd still to me

Yes, unfortunately `cp` is taken, and that's the tradeoff, gotta come up with unique mnemonics (the code action can say whatever).

> I like the intermediate since we can leave to users decide if they want to stop or keep applying, while the "apply all" version doesn't support that.

When would you use the intermediate? I couldn't think of a use case. If there's only an intermediate, I also worry that it would become cumbersome like `expand-let` and applying that over and over.

ericdallo:
[...]

PR #1139, opened 2022-07-22
Destructure keys

mainej:
Adds a refactoring to destructure keys. Similar to https://github.com/clojure-emacs/clj-refactor.el/wiki/cljr-destructure-keys.

@snoe this doesn't address everything in https://github.com/clojure-lsp/clojure-lsp/issues/905, but it's a start. Is this feature set enough for a first draft?
- Also destructures syms, although they're rare in most Clojure code.
- Available either from a local usage or definition. 
- Destructures `(:type map)` but not `(get map :type)`. (I think it would be good to do #904 as an orthogonal feature—useful in its own right.)
- Does not handle strs.
- Does not try to detect [clashing bindings](https://clojure.org/guides/destructuring#_namespaced_keywords), where both `:person/name` and `:company/name` are destructured. If this happens, users will get a lint warning.
- Does not complain if there is nothing to destructure.

Destructuring permits many styles of writing the local definitions. This PR generally follows the styles used in the [Clojure docs](https://clojure.org/reference/special_forms#associative-destructuring) on this subject:

- local definitions become symbols, when possible
  - `(:type map)` -> `{:keys [type]}`
  - `(:ns/type map)` -> `{:keys [ns/type]}`
  - `('type map)` -> `{:syms [type]}`
  - `('ns/type map)` -> `{:syms [ns/type]}`
- otherwise, auto-resolved qualified keywords maintain their form
  - `(::alias/type map)` -> `{:keys [::alias/type]}`
  - `(::type map)` -> `{:keys [::type]}`

This PR doesn't implement the special destructuring forms introduced in Clojure 1.9, which allow `{:ns/keys [type]}` and similar. This is largely personal preference—I find it hard to grep for keys when they've been destructured that way. But, if the community has strong preferences, as I suspect they might, we can introduce settings for other styles.


ericdallo (approved):
Looks pretty good to me!

We could consider extending it to allow calling this on `let`s, probably not a common usage though

snoe (approved):
This is really great! Thanks @mainej

mainej:
> consider extending it to allow calling this on lets

@ericdallo will you clarify? I'm confused because it already works on `let`—or anywhere that defines a local, like `loop`, `for`, `(for :let)`, etc.

ericdallo:
Oh, true, I missed that!

PR #1142, opened 2022-07-22
Ensure that :keep no-ops the initial form indent.

### LINKED ISSUE #1141: Clean-ns with :keep indent introduces an extra new line.
**Describe the bug**
When a namespace is on the same line as `:require`, and :keep indentation is set, the namespace will be placed on the line below the `:require` after clean-ns is called. This also happens with other actions that trigger the clean-ns code like add-missing.

**To Reproduce**
1.  `:clean {:ns-inner-blocks-indentation :keep}` setting
2. have name space like
```
(ns foo
  (:require [foo.a :as a]
            [foo.b :as b])
```
3. clean-ns
4. See that foo.a has moved
```
(ns foo
  (:require 
            [foo.a :as a]
            [foo.b :as b])
```

**Expected behavior**
The ns shouldn't change

on clojure-lsp master

I'll work on this PR

snoe:
Fixes #1141 

I made `:import` also use `ns-inner-blocks-indentation-parent-col` for followup cols, as I see no reason why that should be different than `:require`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
   - Fix to navigate to var defined by declare, when there aren't any later defs. #1107
   - Fix to always go to the definition of the correct var imported by potemkin. #1020
   - Fix to correctly rename namespaces. #1121
+  - Fix to correctly :keep require and import indent spacing. #1141
 
 ## 2022.06.29-19.32.13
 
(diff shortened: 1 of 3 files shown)


ericdallo (approved):
Looks good, thank you!

PR #1166, opened 2022-08-02
Extend prior map destructuring

### LINKED ISSUE #1156: Destructuring previously destructured objects
The new destructure-keys refactoring doesn't interact well with things that have already been destructured.

This is true if the thing being destructured is an `:as` clause:
```clojure
(defn foo [{:keys [a b] :as x}]
  (:y |x))

;; expected
(defn foo [{:keys [a b y]}]
  y)

;; or maybe
(defn foo [{:keys [a b y] :as x}]
  y)

;; actual
(defn foo [{:keys [a b] :as {:keys [y]}}]
  y)
```
It also happens when the thing being destructured is within a `:keys` clause
```clojure
(defn foo [{:keys [x y]}]
  (:a |x))

;; expected
;; no change or _maybe_
(defn foo [{:keys [y], {:keys [a]} :x}]
  a)

;; actual
(defn foo [{:keys [{:keys [a]} y]}]
  a)
```

These both lead to invalid code, so the only real alternative is to recognize that something bad is happening and then to avoid the refactoring.

mainej:
When a local is the `:as` part of a prior map destructuring, this patch extends the destructuring.

```clojure
(let [{:keys [a b] :as x} loc]
  (+ a b (:c |x)))
;; =>
(let [{:keys [a b c]} loc]
  (+ a b c))
```

This patch also refuses to destructure when a prior map destructuring means we would introduce invalid syntax:

```clojure
(let [{:keys [a]} loc]
  (:x |a))
;; we DON'T refactor to this, because it's invalid
(let [{:keys [{:keys [x]}]} loc]
  x)
```

At also does not refactor the above form into a nested destructuring, even though that would be valid syntax:

```clojure
;; we DON'T refactor to this either, even though it's valid
(let [{{:keys [x]} :a} loc]
  x)
```

We avoid this both because it would add complexity to the refactoring implementation and because nested destructuring is sometimes considered poor style.

Finally, this patch refuses to destructure when no changes are necessary:

```clojure
(let [loc' loc] (tangent |loc'))
```

Fixes #1156.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
   - Improve completion sorting, showing locals before functions and other completion items. #1158
   - Fix hover to show current var definition docs instead of `def`/`defn`. #1157
   - Improve hover for requires to show ns docstring. #1171
+  - Fix destructuring of things that have been destructured before. #1156
 
 ## 2022.07.24-18.25.43
 
--- a/lib/src/clojure_lsp/feature/destructure_keys.clj
+++ b/lib/src/clojure_lsp/feature/destructure_keys.clj
@@ -7,70 +7,113 @@
    [rewrite-clj.node :as n]
    [rewrite-clj.zip :as z]))
 
-(defn local-def-elem-to-destructure [zloc uri db]
+(defn ^:private loc-destructuring-key
+  "When `loc` is a val inside a map and its key is a keyword, returns its key.
+  `{:as |x} ;; => :as`"
+  [loc]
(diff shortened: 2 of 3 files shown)


mainej:
@ericdallo I tried to re-run the graalvm-build a few times, but it keeps timing out during the setup-graalvm task. Is that a known problem?

ericdallo:
It seems it started to fail for static linux after #1160, I intend to investigate it this week If I find the time

ericdallo (approved):

PR #1170, opened 2022-08-04
Add refactoring to restructure keys

mainej:
This introduces a new refactoring `restructure-keys`—a companion or inverse to `destructure-keys`. It converts the locals defined by a destructured map back into keyword- or symbol-based gets.

The simplest case is `(defn foo [|{:keys [a]}] a)` -> `(defn foo [element] (:a element))`. A summary of other cases is below, but also see the tests for the most complete list of what's covered.

- Restructures regular `:keys` and `:syms`
- Restructures namespaced `:prefix/keys` and `:prefix/syms`, observing local overrides. Also works for auto-resolved namespaces.
- Restructures namespaced maps `#:prefix{:keys ...}`, observing local overrides. Also works for auto-resolved namespaces.
- Restructures named keys `{a :a}`
- If provided, preserves `:as`
- If provided, restructures `{:or {k default}}` into `(get element :k default)`
- Preserves things that cannot be restructured, like `{{:keys [person/name], :as person} :company/ceo, :as company}`
  - In this case, a user could call restructure-keys twice, first on the inner destructuring, to get `{person :company/ceo, :as company}` and then again to get `company`. But there are other scenarios (such as positional destructuring) which cannot be restructured at all. The refactoring preserves the destructured map in these cases, though it will erase comments within it.

Tasks:


lib/src/clojure_lsp/feature/restructure_keys.clj
```diff
@@ -0,0 +1,190 @@
...
+                                        {:keys [auto-resolved? prefix]} (z/node qualifier-loc)]
+                                    {:filename filename
+                                     :replace-loc loc
+                                     :map-loc (z/right qualifier-loc)
+                                     :map-auto-resolved? auto-resolved?
+                                     :map-prefix prefix})
+          :else nil)))))
+
+(defn can-restructure-keys? [zloc uri db]
+  (boolean (map-to-restructure zloc uri db)))
```
ericdallo, line 38: I wonder how much time this adds to code actions processing, thanks for adding that too BTW, my point is we could have a simple/faster check for code actions not handling complex cases and only fail during code action apply if necessary, only if this code proves affect performance of code actions. WDYT?

lib/src/clojure_lsp/feature/restructure_keys.clj
```diff
@@ -0,0 +1,190 @@
...
+                                 ;; {:keys [my-ns/a]}        -> (:my-ns/a element)
+                                 ;; {:keys [:my-ns/a]}       -> (:my-ns/a element)
+                                 (qualified-ident? local-sexpr) (n/keyword-node (keyword local-sexpr))
+                                 ;; {:my-ns/keys [a]}        -> (:my-ns/a element)
+                                 ;; {::my-alias/keys [a]}    -> (::my-alias/a element)
+                                 default-ns                     (n/keyword-node (keyword default-ns (name local-sexpr)) default-ns-auto-resolved?)
+                                 ;; {::keys [a]}             -> (::a element)
+                                 default-ns-auto-resolved?      (n/keyword-node (keyword local-sexpr) true)
+                                 ;; {:keys [a]}              -> (:a element)
+                                 :else                          (n/keyword-node (keyword local-sexpr)))))
```
ericdallo, line 103: thanks for the comments, it helped reviewing this section

lib/test/clojure_lsp/feature/restructure_keys_test.clj
```diff
@@ -0,0 +1,170 @@
...
+  (assert-cannot-restructure "{|:a 1}")
+  (assert-cannot-restructure "{:a |1}")
+  (assert-cannot-restructure "(|def foo)")
+  (assert-cannot-restructure "(def |foo)")
+  (assert-cannot-restructure "|")
+  (assert-cannot-restructure "|;; comment")
+  (assert-cannot-restructure "(let [loc' loc] (tangent |loc'))")
+  (assert-cannot-restructure "|{:a 1}"))
+
+(deftest should-restructure-local
```
ericdallo, line 47: aweseome tests

ericdallo (approved):
Looks amazing! sorry for the delay
A little concerned about code actions performance, but not sure if it should be a concern right now, feel free to merge if agreed

lib/src/clojure_lsp/feature/restructure_keys.clj
mainej: So you're saying [this](https://github.com/clojure-lsp/clojure-lsp/blob/d611822eec77079a805168afb0cbb2c92ff626c4/lib/src/clojure_lsp/feature/code_actions.clj#L277-L299) worries you? Lol, me too!

IIRC, this feature didn't change the performance of code-actions noticeably. But I can plan to benchmark code-actions anyway. It's a bit tricky because it happens across so many threads, and the work that gets done varies a lot depending on your location. I'll see if I can identify restructure-keys as a performance problem.

If it is, you're right... the answer is to find a simpler code action check that then fails during command execution. I'd feel better about that if we showed the user an error about why we couldn't perform the refactoring. If you return a JSON-RPC error like [here](https://github.com/clojure-lsp/clojure-lsp/blob/d611822eec77079a805168afb0cbb2c92ff626c4/lib/src/clojure_lsp/feature/rename.clj#L231-L233) does the client show that? Or do you have to use `window/showMessage`?

lib/src/clojure_lsp/feature/restructure_keys.clj
mainej: I couldn't keep track of what was going on 10 minutes after I wrote each of these lines, so I had to add these comments 😂

lib/src/clojure_lsp/feature/restructure_keys.clj
mainej: @ericdallo I didn't set up a flamegraph, but by watching the log, the code-actions performance doesn't seem to be a problem. It tends to range between 10ms to 40ms, or even sometimes 60ms. Both ends of that range feel fast enough.

The fact that `restructure-keys` can be included on the low end of the range suggests it's not contributing to the performance significantly. Anecdotally, it seems like the overall time is proportional to the size of the file. I'd guess that means rewrite-clj dominates the time, as it often does. So, that's something to keep in mind... if code-actions ever does get slow, we could try caching the rewrite-clj parse, as we've talked about before.

I'm going to go ahead and merge this.

ericdallo: > If you return a JSON-RPC error like [here](https://github.com/clojure-lsp/clojure-lsp/blob/d611822eec77079a805168afb0cbb2c92ff626c4/lib/src/clojure_lsp/feature/rename.clj#L231-L233) does the client show that?

Not sure about other clients, but emacs would print the error IIRC

> The fact that restructure-keys can be included on the low end of the range suggests it's not contributing to the performance significantly. Anecdotally, it seems like the overall time is proportional to the size of the file. I'd guess that means rewrite-clj dominates the time, as it often does. So, that's something to keep in mind... if code-actions ever does get slow, we could try caching the rewrite-clj parse, as we've talked about before.

Exactly, I already spent some time tunning performance of those actions so it's expected to not be a issue right now, but we don't have control if we add a new code action that will bring all code action time because of that refactor, that was my concern, I don't believe this is the case for this action but we need to be aware of this

mainej: Yeah, understood. This one was particularly good to be wary of because there are a few queries, not just traversals of the parse tree. Something else to add to the performance tests!

PR #1174, opened 2022-08-08
Use dep-graph-queries by default

mainej:
This turns on dep-graph-queries by default. The flag still exists so that we have an easy way to let users disable the feature if it's causing problems. But if we don't get any problem reports, the flag will be removed in the subsequent release.


lib/src/clojure_lsp/feature/file_management.clj
```diff
@@ -32,7 +32,7 @@
         (shared/client-changes changes db)))))
 
 (defn load-document! [uri text db*]
-  (swap! db* #(assoc-in % [:documents uri] {:v 0 :text text :saved-on-disk false})))
```
mainej, line 35: The pod tests for `format` started failing after defaulting to dep-graph queries. This is the fix—`:documents` includes dep-graph data now, so we can't clobber it.

PR #1175, opened 2022-08-09
Thread get

### LINKED ISSUE #1172: Threading get
# Problem

I frequently want to convert between `(:x m)` and `(get m :x)`, and vice-versa. In a related case, when there is more nesting, I want to convert between `(:y (:x m))` and `(get-in m [:x :y])`.

# Proposal
I'd like refactorings that facilitate these conversions. By analogy with the threading refactorings, I propose the following commands.

## Get all

`ga`, `get-all`: convert all nested forms to `get` or `get-in`, as appropriate

```
      |(:x m)
-ga-> |(get m :x)

      |(:z (:y (:x m)))
-ga-> |(get-in m [:x :y :z])
```

## Unwind get

`ug`, `unwind-get`: fully unwind a `get` expression

```
      |(get m :x)
-ug-> |(:x m)

      |(get-in m [:x :y :z])
-ug-> |(:z (:y (:x m)))
```

## Get more

`gm`, `get-more`: convert one nested form into a `get`, or extend an existing `get` or `get-in`

```
      |(:z (:y (:x m)))
-gm-> |(get (:y (:x m)) :z)
-gm-> |(get-in (:x m) [:y :z])
-gm-> |(get-in m [:x :y :z])
```

## Get less

`gl`, `get-less`: unwind a `get` expression, one step at a time

```
      |(get-in m [:x :y :z])
-gl-> |(get-in (:x m) [:y :z])
-gl-> |(get (:y (:x m)) :z)
-gl-> |(:z (:y (:x m)))
```

## Getting outside

When you want to "get more", there are actually two possibilities. The proposed
`get-more` lets you extend the left side of the get path, by getting more from
_inside_ the get expression. But you may want to extend the right side of the
get path, by getting more from _outside_ the get expression.

To do this, unwind fully, then `get-all`.

Starting with a `get`:  
```
      (:z (:y |(get m :x)))          start: :y is outside the get
-ug-> (:z (:y |(:x m)))
->    (:z |(:y (:x m)))
-ga-> (:z |(get-in m [:x :y]))       goal: :y is inside, and the get has become a get-in
```

Starting with a `get-in`:
```
      (:z |(get-in m [:x :y]))       start: :z is outside the get-in
-ug-> (:z |(:y (:x m)))
->    |(:z (:y (:x m)))
-ga-> |(get-in m [:x :y :z])         goal: :z is inside the get-in
```

This approach is by analogy to the (regular) threading refactorings. There's a dedicated
command (in clj-refactor, but not clojure-lsp, afaik) for threading more from
the inside:

```
      (-> (:x m) :y :z)
-th-> (-> m :x :y :z)
```

But threading more from the outside requires multiple steps:
```
      (:z |(-> m :x :y))
-ua-> (:z |(:y (:x m)))
->    |(:z (:y (:x m)))
-tf-> |(-> m :x :y :z)
```

Similarly, when you want to "get less" there are two possibilities. The proposed
`get-less` lets you unwind from the left of the get path, adding to the inside of
the get. But you may want to unwind from the right of the get path, adding to
the outside of the get.

To do this, first unwind fully:

```
      |(get-in m [:x :y :z])        start: everything is inside the get expression
-ug-> |(:z (:y (:x m)))
```

Then pick the desired depth, then `get-all`:
```

->    (:z |(:y (:x m)))
-ga-> (:z |(get-in m [:x :y]))      goal: :z is outside the get expression

-- or --

->    (:z (:y |(:x m)))
-ga-> (:z (:y |(get m :x)))         goal: both :z and :y are outside, and the get-in has become a get
```

Again, this approach is by analogy to the regular threading refactorings. There's a dedicated
command for unwinding the first element:

```
      (-> m :x :y :z)
-uw-> (-> (:x m) :y :z)
```

But unwinding the last element requires multiple steps:

```
      (-> m :x :y :z)
-ua-> |(:z (:y (:x m)))
->    (:z |(:y (:x m)))
-ta-> (:z |(-> m :x :y))
```

# Alternatives

I've come up with a few other ways to "get more" or "get less" which provide more direct routes to operating on the last element in the path.

## Option 1, Use two commands

The first option is to introduce another command that "gets less" by unwinding to the outside. This command would be in addition to `get-less`, although to be consistent, `get-less` would probably be renamed `unwind-inside-get`.

`uo`, `unwind-outside-get`: unwind a get expression, adding to the outside

```
      |(get-in m [:x :y :z])
-uo-> (:z |(get-in m [:x :y]))
-uo-> (:z (:y |(get m :x)))
-uo-> (:z (:y (:x |m)))
```

Correspondingly, to "get more" from outside, `get-more` would be renamed to `get-more-from-inside` and a new command would be added:

`go`, `get-more-from-outside`: get more from outside

```
      (:z (:y (:x |m)))
-go-> (:z (:y |(get m :x)))
-go-> (:z |(get-in m [:x :y]))
-go-> |(get-in m [:x :y :z])
```

I like that this provides very focused refactorings that do exactly what I want. But I don't like that it requires adding two more commands. There are already 4 commands proposed and this would bring the total to 6. The regular threading refactorings have 5 commands, so maybe it would be OK, but still, it feels like a lot to learn and remember.

## Option 2, Location aware commands

Another option is to use the cursor position to infer the user's intention. Instead of invoking `get-less` from outside of the get expression, we invoke it from inside the path.

When at beginning of path:

```
      (get-in m [|:x :y :z])
-gl-> (get-in (:x m) [|:y :z])
-gl-> (get (:y (:x m)) |:z)
-gl-> (:z (:y (:x |m)))
```

When at end of path:

```
      (get-in m [:x :y |:z])
-gl-> (:z (get-in m [:x |:y]))
-gl-> (:z (:y (get m |:x)))
-gl-> (:z (:y (:x |m)))
```

At first blush, this is really nice. It gives you the same quick refactorings without introducing another command. Unfortunately, it breaks down when you consider `get-more`.

The analogous idea works OK when getting more from a `get-in` (with two or more elements in its path). But the intention is ambiguous when getting more from `get` (or `get-in` with one element in its path). Do you want to add to the right or left?

```
      (:z (get (:x m) |:y))
```

What should `gm` mean from here?

```
-gm-? (:z (get-in m [|:x :y]))

-- or --

-gm-? (get (:x m) [:y |:z])
```

I've decided against this option because of this ambiguity, though I could be convinced to reconsider.

# Other concerns

## Default values

I haven't thought very carefully about default values, like`(get m :x 1)` or `(:x m 1)`. I think it would be nice if they were handled, but we'd have to be careful. Default values work with keyword keys (and symbols keys like `'x`?)  but it wouldn't be correct to convert `(get m x 1)` to `(x m 1)`.

## Threading locals, variables, and other forms

I also haven't though much about which keys can be safely transformed. Keywords and symbols are probably always safe. But what about locals?

These have the same meanings:
```clojure
(let [m {:x 1, :y -1}
      k :x]
  (get m k)) ;; => 1

(let [m {:x 1, :y -1}
      k :x]
  (k m)) ;; => 1
```

But these do not:

```clojure
(let [m {inc 1, dec -1}
      k inc]
  (get m k)) ;; => 1

(let [m {inc 1, dec -1}
      k inc]
  (k m)) ;; => <throws ClassCastException>
```

We don't (and can't) have static analysis of what every local is bound to, so I think we have to pick. Either (A) we refuse to refactor when we reach a local (or halt a recursive refactoring), or (B) we blindly continue, leaving the problem to the user. If we _really_ wanted to get fancy, we could ask the user whether to continue.

The same concerns apply not just to locals, but to variables  `(get m my-alias/x)` and forms `(get m (name :k))` as well.

## Arity

We'll have to be observant of the arity of the forms we're refactoring, refusing to refactor (or halting recursive refactorings) when reaching a form with 2 or more arguments. (Although see above about default values, which are 2-arg forms we would refactor.)

```
      |(z (y (x m) 5))
-ga-> |(get (y (x m) 5) z)     second argument to y maintained

-- not --

-ga-> |(get-in m [x y z])      second argument to y lost

### PR DESCRIPTION
**Pull request by mainej**
Introduces four refactorings for manipulating get and get-in.

* get-in-all: convert all nested forms to get or get-in, as appropriate
```
      |(:x m)
-ga-> |(get m :x)

      |(:z (:y (:x m)))
-ga-> |(get-in m [:x :y :z])
```

* get-in-none: fully unwind a get expression
```
      |(get m :x)
-gn-> |(:x m)

      |(get-in m [:x :y :z])
-gn-> |(:z (:y (:x m)))
```

* get-in-more: convert one nested form into a get, or extend an existing get or get-in
```
      |(:z (:y (:x m)))
-gm-> |(get (:y (:x m)) :z)
-gm-> |(get-in (:x m) [:y :z])
-gm-> |(get-in m [:x :y :z])
```

* get-in-less: unwind a get expression, one step at a time
```
      |(get-in m [:x :y :z])
-gl-> |(get-in (:x m) [:y :z])
-gl-> |(get (:y (:x m)) :z)
-gl-> |(:z (:y (:x m)))
```

The refactorings convert between different forms of default values: `(:x m 1)` and `(get m :x 1)`.

They're also careful to halt when they reach a form that can't be safely refactored.

```
      |(:z (:y (:x m) 5))
-ga-> |(get (:y (:x m) 5) :z)     second argument to :y maintained

-- not --

-ga-> |(get-in m [:x :y :z])      second argument to :y lost
```

This fixes #1172


ericdallo (approved):
Sorry for the delay, Looks good and works perfectly! thank you!

PR #1182, opened 2022-08-11
More carefully slurp

mainej:
Sometimes editors use URIs that don't actually exist on disk. In these
cases, slurping will fail. This change introduces a tool to slurp more
carefully; it returns nil instead of throwing. Other code is adapted to
handle the potential absence of a file's text.

@ericdallo this is a fix for the bug you [noticed](https://github.com/clojure-lsp/clojure-lsp/pull/1117#issuecomment-1207264273) about renaming files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
   - Some completions require that a new alias be added to the ns form. Fix this feature for Calva users, and improve performance for all users. #1068
   - Fix resolve-macro-as code action corner case. #1084
   - Ensure line numbers provided by document-symbol correspond to the latest version of the file. #1178
+  - Avoid exceptions when clients use URIs that don't exist on disk.
 
 ## 2022.07.24-18.25.43
 
--- a/lib/src/clojure_lsp/feature/file_management.clj
+++ b/lib/src/clojure_lsp/feature/file_management.clj
@@ -273,10 +273,11 @@
       :changed (when (settings/get @db* [:compute-external-file-changes] true)
                  (shared/logging-task
                    :changed-watched-file
-                   (did-change uri
-                               [{:text (slurp (shared/uri->filename uri))}]
-                               (get-in @db* [:documents uri :v] 0)
-                               db*)))
+                   (when-let [text (shared/slurp-uri uri)]
+                     (did-change uri
+                                 [{:text text}]
+                                 (get-in @db* [:documents uri :v] 0)
+                                 db*))))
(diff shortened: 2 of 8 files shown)


mainej:
> I wonder why we need to slurp

We need to slurp in a few different scenarios.

First, when we receive a `didChangeWatchedFiles` (for a `:changed` file) the editor gives us a URI, but not the text of the document. The editor is saying that it knows a file has changed, but not _how_ it has changed. In particular, it doesn't know what the file's current text is. So, to keep in sync with the file system we slurp.

Second, note that the DB doesn't contain the text of every file. We add a file's text only after we receive a `didOpen` (or a `didChangeWatchedFiles` as noted above). But some features need the text of files that haven't been opened yet. For example, `move-form` might be moving a function into a file that hasn't been opened. It needs to parse that file's text with rewrite-clj, so that it can append the function, adjust the `ns` form, etc. In these cases, we use `f.file-management/force-get-document-text`, which slurps.

There are five uses of `f.file-management/force-get-document-text`. Two of these I understand.

1. move-form: Makes sense; explained above.
2. call-hierarchy: As you navigate through the call hierarchy you often end up in files that haven't been opened. We use the rewrite-clj tree to figure out which var is using/calling the var we're on, so we need to slurp unopened files.

The other three don't make sense to me.

1. `format`: Format is used both in the API and by clients. When called from the API, we slurp all the files before we call format. And I think clients only format files they have open already. So I don't understand when we'd need to slurp.
2. `hover`: Again, I think clients call this only on files that they have open.
3. `signature-help`: Again, I think clients call this only on files that they have open.

But even in the cases I don't understand, it doesn't really hurt to use `f.file-management/force-get-document-text`. It prefers text that's in the db already and only slurps if we don't have that text yet.

Finally there are some places that slurp, but which could use `f.file-management/force-get-document-text`, potentially avoiding an unnecessary slurp.

1. `internal-api/edit->summary`: The slurp here looks like a re-implementation of `force-get-document-text`. (The slurp is necesssary, I think.)
2. `transform/create-test-for-source-path`: This slurps the test file even if we already have its text in the db. Might be better to use `force-get-document-text`.

I'm going to merge this PR as is. If you'd like, I could look at these other cases more to see if I can clean things up.

ericdallo:
I completely agree with all those points :)
I think it's ok to use force-get-document when we are not sure we will have, but I'd avoid using for all cases and stop using for cases we know we should have the text like format, hover and signatureHelp as you mentioned

PR #1188, opened 2022-08-11
Harmonize meaning of row, col, line and character

mainej:
clojure-lsp has three actively used ways of referring to a position in a
document.

* line, character: These are the terms used in the LSP spec, and are
  0-based.
* row, col: These are the terms used by rewrite-clj and clj-kondo and
  are 1-based.
* line, column: I'm not sure where these terms come from, perhaps vim or
  perhaps the Clojure parser. They are also 1-based.

Note that the first and third both use the term "line", but one is
0-based and the other is 1-based. While reading the code, the only way
to understand what "line" means is by looking at whether it's paired
with "character" or "column". And the second uses "col" while the third
uses "column", though they have different meanings.

To add to the cognitive burden, some code uses row/col when it
means line/character, or uses one set of terminology in var definitions,
but another at call sites.

This commit brings the naming into harmony. In 0-based scenarios, we use
line/character. In 1-based scenarios, we use row/col. We no longer use
line/column.

### DIFF
--- a/cli/integration-test/integration/fixture.clj
+++ b/cli/integration-test/integration/fixture.clj
@@ -18,25 +18,25 @@
   []
   [:shutdown {}])
 
-(defn completion-request [path row col]
+(defn completion-request [path line character]
   [:textDocument/completion
    {:textDocument {:uri (h/source-path->uri path)}
-    :position {:line row :character col}}])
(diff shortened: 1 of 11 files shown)


ericdallo (approved):
Thanks! that was something that was always a pain while debugging

PR #1190, opened 2022-08-12
Fixes #1189

### LINKED ISSUE #1189: Some clojure-lsp integration tests fail if project checkout path contains "src"
**Describe the bug**
If clojure-lsp source code is checked out in a directory that contains "src" (e.g. "/Users/jgracin/src/clojure-lsp), running `make test` fails on 2 tests in api_test.clj because the tests use `(string/replace %1 "src" "fixtures")` which replaces all occurrences of `src` in the path instead of just the last one.

**To Reproduce**

1. Clone the clojure-lsp project to a directory that contains "src" somewhere in its path, e.g. /tmp/src/clojure-lsp.
2. Run `make test`.
3. Observe that 2 tests in api_test.clj failed (specifically "_format!/when project-root is a valid file/when a single namespace is specified_" and "_clean-ns!/when project-root is a valid file/when a single namespace is specified_").

**Expected behavior**

All tests should pass (independently of the checkout directory name).

**User details**
 - OS: MacOS
 - Version: Latest clojure-lsp master branch, commit f385e1c1b

jgracin:
Here's a simple fix for 1189.

### DIFF
--- a/lib/test/clojure_lsp/api_test.clj
+++ b/lib/test/clojure_lsp/api_test.clj
@@ -8,6 +8,12 @@
 
 (h/reset-db-after-test :api-test)
 
+(defn- replace-last
+  "Replaces the last occurrence of `old-value` in `s` with `new-value`."
+  [s old-value new-value]
+  (string/reverse
+   (string/replace-first (string/reverse s) (string/reverse old-value) (string/reverse new-value))))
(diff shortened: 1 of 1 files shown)


ericdallo (approved):
Thank you!

PR #1196, opened 2022-08-16
Fix: publishing diagnostics at startup fills up the channel

### LINKED ISSUE #1153: Publishing diagnostics at startup fills up the channel
**Describe the bug**

I noticed an exception in the logs when testing the latest `clojure-lsp` binary. Not sure how this affects the editor, perhaps some of the diagnostics get discarded. I don't think I've seen this before.

**To Reproduce**
1. Open a large project (thousands of namespaces) in an editor
2. Look at the logs
3. Notice hundreds of "Publishing N diagnostics for file..." messages
4. See stacktrace

**Expected behavior**

No errors should occur. All diagnostics should be reported.

<details>
<pre>
2022-07-26T04:21:37.250Z  ERROR [taoensso.timbre:828] - Uncaught exception on thread: async-dispatch-2
com.oracle.svm.core.posix.thread.PosixJavaThreads.pthreadStartRoutine    PosixJavaThreads.java:  194
            com.oracle.svm.core.thread.JavaThreads.threadStartRoutine         JavaThreads.java:  597
                                                 java.lang.Thread.run              Thread.java:  829
                                                                  ...                               
   clojure.core.async.impl.concurrent/counted-thread-factory/reify/fn           concurrent.clj:   29
                   java.util.concurrent.ThreadPoolExecutor$Worker.run  ThreadPoolExecutor.java:  628
                    java.util.concurrent.ThreadPoolExecutor.runWorker  ThreadPoolExecutor.java: 1128
                                                                  ...                               
                                   clojure-lsp.handlers/initialize/fn             handlers.clj:  105
         clojure.core.async.impl.ioc-macros/run-state-machine-wrapped           ioc_macros.clj:  982
                 clojure.core.async.impl.ioc-macros/run-state-machine           ioc_macros.clj:  978
             clojure-lsp.handlers/initialize/fn/state-machine--auto--             handlers.clj:  105
          clojure-lsp.handlers/initialize/fn/state-machine--auto--/fn             handlers.clj:  106
             clojure-lsp.feature.diagnostics/publish-all-diagnostics!          diagnostics.clj:  165
            clojure-lsp.feature.diagnostics/sync-publish-diagnostics!          diagnostics.clj:  145
                                              clojure.core.async/put!                async.clj:  199
              clojure.core.async.impl.channels.ManyToManyChannel/put!             channels.clj:  156
java.lang.AssertionError: Assert failed: No more than 1024 pending puts are allowed on a single channel. Consider using a windowed buffer.
                          (< (.size puts) impl/MAX-QUEUE-SIZE)
</pre>
</details>

 - OS: Arch Linux
 - Editor emacs
 - Version: `clojure-lsp 2022.07.24-18.25.43` + `clj-kondo 2022.06.23-SNAPSHOT`

mainej:
At startup clojure-lsp analyzes project files with clj-kondo. clj-kondo finds diagnostics (a.k.a. lint) and clojure-lsp sends those diagnostics to the editor by spooling them onto a core.async channel called `db/diagnostics-chan`. As described in #1153, in very large projects (thousands of namespaces) this can lead to an exception "Assert failed: No more than 1024 pending puts are allowed on a single channel." When this happens the editor doesn't receive all the diagnostics.

To explain, let me tell a story.

Sometimes I find myself standing in line at the grocery store alone, waiting while my wife looks for "one more thing." This is a useful model for reasoning about core.async.

At the core.async grocery store, shoppers wander around picking out items and putting them in their cart. They wheel their carts to the checkout area, and here's where things get weird. At this store, you aren't allowed to put an item on a conveyor (`chan`) yourself. Instead each conveyor has an attendant. You hand an item to the attendant, who puts it on the conveyor for you. There isn't even a line. Many people can hand items to the attendant all at once.

Usually the cashier is quick and keeps the conveyor empty, so the attendant can place each new item immediately. But sometimes the cashier gets behind. If things back up, the attendant has a basket of their own. They'll keep taking the shopper's items, storing them in their own basket until the cashier catches up.

The attendant has space for 1024 items in their basket. Any more than that, and as you hand an item over, poof! it disappears (`No more than 1024 pending puts are allowed on a single channel`).

As a shopper you have three options as you hand an item to the attendant.

If you don't care whether your item makes it onto the conveyor, you can hand it over and walk away (`put!`), getting back to shopping.

Even if you're the only person in the store, this can cause problems. If you revisit checkout many times in a row, walking away each time, you can accidentally fill up the basket.

To try to avoid this problem, you can wait until you see the attendant put the item on the conveyor (`>!!`) before returning to your shopping. This still doesn't guarantee the attendant's basket won't fill up. If there are other shoppers, they could all come to checkout at the same time. But at least you know you won't cause a problem by yourself.

Some shoppers feel like waiting is a waste of their time so the store has arranged a third option.

With a little ceremony (`go`) you can _hire someone else_ to wait with your cart. The hiring process takes a moment to set up (HR benefit forms) but is usually quite fast, so shoppers love this option. The new worker will hand the items in the cart to the attendant one at a time, waiting to see each one reach the conveyor `>!`. If the cart has only one item (`go` and a single `>!`) this [isn't much different](https://github.com/clojure/core.async/wiki/Go-Block-Best-Practices#general-advice) than walking away (`put!`). But if the cart has several items, the new worker will wait for each one to reach the conveyor before giving the attendant another (`go-loop` and `>!`, or more simply `onto-chan!`). This lets you continue your shopping, with more confidence that your items won't overwhelm the attendant.

Coming back to this bug...

Suppose your project has 1200 files. At startup, clojure-lsp calls `sync-publish-diagnostics!` in a loop, once for each file. Each one of these invokes `(put! diagnostics-chan diagnostic)`. This is like making 1200 trips to checkout, walking away after handing over a single item each time. The attendant ends up with too many items and poof! some of them disappear. You might think you could fix this by calling `(go (>! diagnostics-chan diagnostic))` in a loop instead, and indeed, there are places where we were doing that. But this won't work either. It's like hiring 1200 new workers, each with one item in their cart. The new workers wouldn't coordinate with each other. They'd each hand their single item over, the attendant would get 1200 items all at once, and again, poof!

When looked at this way, there a few possible fixes. You could make the conveyor longer (i.e., `(chan 1000)`), effectively giving the attendant some extra space in addition to their basket. But I think it's better to hire only one worker, with 1200 items in their cart. They can hand items to the attendant one at a time, waiting for each one to reach the conveyor before handing over another. The attendant's basket is much less likely to fill up this way. (It's still possible of course—there are other shoppers in the store.)

This commit changes the code so that whenever we have several file's worth of diagnostics, we add them to the diagnostics channel with `onto-chan!`. That is, we hire one worker with a very full cart, instead of walking away after each item or hiring thousands of workers each with single-item carts.

That's the main fix.

This commit also removes the test-specific pathways through the diagnostics code. These pathways were put into place to avoid test flakiness. That was working—the tests haven't been flaky lately—but not for the reason we thought. To understand what I mean and why I don't think the test-specific pathways are necessary, let's start with some very old code, before any flakiness fixes were in place.

```clojure
;; flaky
(defn maybe-publish [{:keys [publish?]} diagnostic]
  (when publish?
    (go (>! db/diagnostics-chan diagnostic))))

(deftest diagnostics-should-not-always-be-published
  (maybe-publish {:publish? true} diagnostic)
  (let [mock-chan (chan 1)]
    (alter-var-root #'db/diagnostics-chan mock-chan)
    (maybe-publish {:publish? false} diagnostic)
    (is (= :timeout (h/take-or-timeout mock-chan 200 :timeout)))))
```

These tests were flaky. The final line would fail occasionally, because a message _was_ put on the mock-chan. That meant that sometimes `>!` from the first line of the test didn't actually run until _after_ the mock-chan was installed with `alter-var-root`.

The original fix for the flakiness looked like this:

```clojure
;; not flaky
(defn maybe-publish [{:keys [publish? env]} diagnostic]
  (when publish?
    (if (= :test env)
      (put! db/diagnostics-chan diagnostic)
      (go (>! db/diagnostics-chan diagnostic)))))

(deftest diagnostics-should-not-always-be-published
  (maybe-publish {:publish? true, :env :test} diagnostic)
  (let [mock-chan (chan 1)]
    (alter-var-root #'db/diagnostics-chan mock-chan)
    (maybe-publish {:publish? false, :env :test} diagnostic)
    (is (= :timeout (h/take-or-timeout mock-chan 200 :timeout)))))
```

This worked—these tests weren't flaky. That seemed to prove that `put!` was somehow "less asynchronous".

But wait! Earlier I pointed out that the Clojure docs say `go` + `>!` isn't much different than `put!`. And I'll go further and link to the [docs](https://clojuredocs.org/clojure.core.async/put!) that say `put!` "Asynchronously puts a val into port." So how did using a different but equivalent thing which is also asynchronous fix flakiness caused by asynchronicity?

Let's go back to the grocery store. Remember what I said about hiring a new worker and how it takes a moment? What I meant is that the body of a `go` block is executed asynchronously. And this is the key.

When you write `(go (>! db/diagnostics-chan diagnostic))` then you are _picking which channel_ to put the message on asynchronously. Since the choice is asynchronous, you could pick before or after `alter-var-root`. If it's before, you pick the original channel, and the tests will pass. But if it's after, you pick the mock-chan, and the tests will fail.

While the first new worker is filling out their HR forms (`go`), the attendant, conveyor and cashier are being replaced with temporary versions (`alter-var-root`). If the worker fills out their forms fast enough, then they put (`>!`) the item on the original conveyor. But if they're too slow, they put the item, incorrectly, on the temporary conveyor.

The **real** fix is to pick the channel before you start any asynchronous task. That's why using `(put! db/diagnostics-chan diagnostic)` fixed the flakiness. That code picks the channel, _then_ runs the async code in `put!`. If you're going to use `go`, you have to tell each worker which conveyor to use before they start filling out their HR forms. Indeed, it's possible to avoid the flakiness, while still using `go` + `>!`, by making a very small change to the original flaky code:

```clojure
;; not flaky
(defn maybe-publish [{:keys [publish?]} diagnostic]
  (when publish?
    (let [ch db/diagnostics-chan] ;; pick channel before running async code
      (go (>! ch diagnostic)))))
```

Either style—picking the channel before running `go`, or using `put!`—fix the flaky code. And despite what we originally thought, they are both equally asynchronous. And `put!` is shorter. So as far as I can see, there's no reason not to use `put!` in production and in tests, so that's what I've switched it to.

The fact that we thought `put!` was "less asynchronous" is why pathways that used `put!` were named "sync", and pathways that used `go` + `>!` were named "async". That terminology is misleading, so I've removed it.

As a final note, there's an even deeper problem still, which is that `db/diagnostics-chan` is a global variable. If instead of holding the channels in global vars, we threaded them through the app (as components perhaps, as we do with `db*` as of recently), it would be much easier and safer to provide a mock-chan in tests. That's another refactoring, for a later time.

Fixes #1153, I hope.

### DIFF
--- a/lib/src/clojure_lsp/feature/diagnostics.clj
+++ b/lib/src/clojure_lsp/feature/diagnostics.clj
@@ -141,38 +141,31 @@
         (not= :off depend-level)
         (concat (clj-depend-violations->diagnostics filename depend-level db))))))
 
-(defn sync-publish-diagnostics! [uri db]
-  (async/put! db/diagnostics-chan
-              {:uri uri
-               :diagnostics (find-diagnostics uri db)}))
+(defn publish-diagnostic!* [diagnostic]
+  (async/put! db/diagnostics-chan diagnostic))
 
-(defn async-publish-diagnostics! [uri db]
-  (if (#{:unit-test :api-test} (:env db)) ;; Avoid async on test which cause flakeness
-    (async/put! db/diagnostics-chan
-                {:uri uri
-                 :diagnostics (find-diagnostics uri db)})
-    (async/go
-      (async/>! db/diagnostics-chan
-                {:uri uri
-                 :diagnostics (find-diagnostics uri db)}))))
+(defn publish-all-diagnostics!* [diagnostics]
+  (async/onto-chan! db/diagnostics-chan diagnostics false))
+
+(defn publish-diagnostics! [uri db]
+  (publish-diagnostic!* {:uri uri
+                         :diagnostics (find-diagnostics uri db)}))
 
 (defn publish-all-diagnostics! [paths db]
-  (doseq [path paths]
-    (doseq [file (file-seq (io/file path))]
-      (let [filename (.getAbsolutePath ^java.io.File file)
-            uri (shared/filename->uri filename db)]
-        (when (not= :unknown (shared/uri->file-type uri))
-          (sync-publish-diagnostics! uri db))))))
+  (publish-all-diagnostics!*
+    (eduction (map io/file)
+              (mapcat file-seq)
+              (map #(.getAbsolutePath ^java.io.File %))
+              (map #(shared/filename->uri % db))
+              (remove #(= :unknown (shared/uri->file-type %)))
+              (map (fn [uri]
+                     {:uri uri
+                      :diagnostics (find-diagnostics uri db)}))
+              paths)))
 
-(defn publish-empty-diagnostics! [uri db]
-  (if (#{:unit-test :api-test} (:env db))
-    (async/put! db/diagnostics-chan
-                {:uri uri
-                 :diagnostics []})
-    (async/go
-      (async/>! db/diagnostics-chan
-                {:uri uri
-                 :diagnostics []}))))
+(defn publish-empty-diagnostics! [uri]
+  (publish-diagnostic!* {:uri uri
+                         :diagnostics []}))
 
 (defn ^:private unused-public-vars [var-defs kw-defs project-db kondo-config]
   (let [var-definitions (remove (partial exclude-public-diagnostic-definition? kondo-config) var-defs)
--- a/lib/src/clojure_lsp/feature/diagnostics.clj
+++ b/lib/src/clojure_lsp/feature/diagnostics.clj
@@ -141,10 +141,10 @@
         (not= :off depend-level)
         (concat (clj-depend-violations->diagnostics filename depend-level db))))))
 
-(defn publish-diagnostic!* [diagnostic]
+(defn ^:private publish-diagnostic!* [diagnostic]
   (async/put! db/diagnostics-chan diagnostic))
 
-(defn publish-all-diagnostics!* [diagnostics]
+(defn ^:private publish-all-diagnostics!* [diagnostics]
   (async/onto-chan! db/diagnostics-chan diagnostics false))
 
 (defn publish-diagnostics! [uri db]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
   - Avoid exceptions when clients use URIs that don't exist on disk.
   - Fix documentation on completion items. #1181
   - Fix rename of defrecords. #1165
+  - Fix to send all diagnostics to client at startup, even in very large projects. #1153
 
 ## 2022.07.24-18.25.43
 
--- a/common-test/src/clojure_lsp/test_helper.clj
+++ b/common-test/src/clojure_lsp/test_helper.clj
@@ -85,17 +85,17 @@
        (f)))))
 
 (defmacro let-mock-chans [bindings & body]
-  (assert (even? (count bindings)))
+  {:pre [(even? (count bindings))]}
   (let [bs (partition 2 bindings)
         let-bindings (mapcat (fn [[binding _]]
-                               `[~binding (async/chan 1)])
+                               [binding `(async/chan 1)])
                              bs)
-        alter-vars (map (fn [[binding chan-var]]
-                          `(alter-var-root ~chan-var (constantly ~binding)))
-                        bs)]
+        redef-bindings (mapcat (fn [[binding chan-var]]
+                                 [chan-var binding])
+                               bs)]
     `(let [~@let-bindings]
-       ~@alter-vars
-       ~@body)))
+       (with-redefs [~@redef-bindings]
+         ~@body))))
 
 (defn take-or-timeout [c timeout-ms]
   (let [timeout (async/timeout timeout-ms)
--- a/lib/src/clojure_lsp/feature/file_management.clj
+++ b/lib/src/clojure_lsp/feature/file_management.clj
@@ -43,7 +43,7 @@
                  (-> state-db
                      (lsp.kondo/db-with-results kondo-result)
                      (lsp.depend/db-with-results depend-result))))
-    (f.diagnostic/async-publish-diagnostics! uri @db*))
+    (f.diagnostic/publish-diagnostics! uri @db*))
   (when allow-create-ns
     (when-let [create-ns-edits (create-ns-changes uri text @db*)]
       (async/>!! db/edits-chan create-ns-edits))))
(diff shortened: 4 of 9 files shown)


mainej:
@ericdallo two follow-ups to my very long explanation... We thought that the [similar fix](https://github.com/clojure-lsp/lsp4clj/commit/b2c3399355989adb9e20614e6ba15e61002c2be6) in lsp4clj might avoid this problem. But I hope it's clear that we essentially just moved it, from the lsp4clj server's `:output-ch` to `db/diagnostics-chan`.

Second, there's only one place that uses `env` now. In `settings/all` we skip memoization in tests, I assume so we can always see the latest values in settings.edn. There are a few other ways we could do that. First, I think it's possible to clear a memoization cache, which we could do as part of `t/clean-db!`. Or second, we could create a settings component, and thread it through the system. (I think we talked about that a long time ago.) The production version of the component would need to refresh itself, but the test version could be fixed. Anyway, that, along with creating components for the chans, can be done some other time.

lib/src/clojure_lsp/feature/diagnostics.clj
```diff
@@ -141,38 +141,31 @@
...
-                 :diagnostics (find-diagnostics uri db)})
-    (async/go
-      (async/>! db/diagnostics-chan
-                {:uri uri
-                 :diagnostics (find-diagnostics uri db)}))))
+(defn publish-diagnostic!* [diagnostic]
+  (async/put! db/diagnostics-chan diagnostic))
+
+(defn publish-all-diagnostics!* [diagnostics]
+  (async/onto-chan! db/diagnostics-chan diagnostics false))
```
ericdallo, line 148: Can we make those private functions?

ericdallo:
> Second, there's only one place that uses env now. In settings/all we skip memoization in tests, I assume so we can always see the latest values in settings.edn. There are a few other ways we could do that. First, I think it's possible to clear a memoization cache, which we could do as part of t/clean-db!. Or second, we could create a settings component, and thread it through the system. (I think we talked about that a long time ago.) The production version of the component would need to refresh itself, but the test version could be fixed. Anyway, that, along with creating components for the chans, can be done some other time.

Good catch! I really like the settings component idea, using memorize hard coded on that code really make things hard to test, extracting settings to a component should help with that, sounds great.

lib/src/clojure_lsp/feature/diagnostics.clj
mainej: Yep, done.

lassemaatta:
@mainej fyi: a while back when testing performance stuff wrt. `clojure-lsp`, I created a quick-and-dirty tool for generating large mock leiningen projects [[link](https://github.com/lassemaatta/dummy-project-generator)]. I'd imagine it might be somewhat useful when you need to try out various "what if we have 10 000 namespaces" type stuff.

mainej:
> generating large mock leiningen projects

Absurd but useful. Love it.

PR #1199, opened 2022-08-17
Separate dev, test and api databases; componentize chans

mainej:
This PR improves the development experience when working on clojure-lsp itself.

With the lsp4clj refactoring, components—including db*—are threaded through handlers instead of referenced as global variables.

In development, the handler's db* is also put in the global variable db/db*. This is for convenience when using the nREPL, so you can inspect the contents of the db outside of a handler function.

Unfortunately, the tests also use the db/db* global variable. What this means is that if you start the nREPL and then run a test, the test-helper/clean-db! fixture clears the contents of the handler's db*. With an empty db, all the navigation and refactorings stop working. So, you could do one or the other—have a working lsp, or run tests—but not both.

Similarly, four channels were stored as global variables in the db namespace (diagnostics-chan, et. al.) The tests clobber these channels. So, after running a test, the client would stop receiving messages, leading to freezes and timeouts.

This PR gives the tests their own db*. It also turns the channels into components. Together these allow you to run tests and have a working lsp. db/db* still exists, but is only ever populated in dev. It's the same object as the handler uses.

This PR also gives the API its own db*. This db* is a little different than the others, because it needs to be preserved between calls into the API.

### DIFF
--- a/cli/src/clojure_lsp/server.clj
+++ b/cli/src/clojure_lsp/server.clj
@@ -496,11 +496,11 @@
   ;; We don't have an ENV=development flag, so the next best indication that
   ;; we're in a development environment is whether we're able to start an nREPL.
   (when-let [nrepl-port (nrepl/setup-nrepl)]
-    ;; We're in the development environment, so make the db* atom available
-    ;; globally as db/db*. In other environments it's empty (except for unit
-    ;; tests, which set it via a different mechanism).
-    (alter-var-root #'db/db* (constantly db*))
-    (swap! db/db* assoc :port nrepl-port)))
+    ;; Save the port in the db, so it can be reported in server-info.
+    (swap! db* assoc :port nrepl-port)
+    ;; In the development environment, make the db* atom available globally as
+    ;; db/db*, so it can be inspected in the nREPL.
+    (alter-var-root #'db/db* (constantly db*))))
 
 (defn run-server! []
   (lsp.server/discarding-stdout
--- a/cli/src/clojure_lsp/server.clj
+++ b/cli/src/clojure_lsp/server.clj
@@ -457,8 +457,8 @@
        (recur))))
 
 (defn ^:private spawn-async-tasks!
-  [{:keys [producer current-changes-chan created-watched-files-chan edits-chan] :as components}]
-  (let [debounced-diags (shared/debounce-by db/diagnostics-chan diagnostics-debounce-ms :uri)
+  [{:keys [producer current-changes-chan diagnostics-chan created-watched-files-chan edits-chan] :as components}]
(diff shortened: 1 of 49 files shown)


mainej:
@ericdallo I have a follow-up PR that makes the db chans components. It touches a lot of the same code. Reviewing it will be equally as dull. Would you prefer to do two reviews, or have me merge that branch into this PR so you can review both at once?

ericdallo:
> With an empty db, all the navigation and refactorings would stop
working.

Yeah, I noticed that, emacs completely freezes and I need to force quit clojure-lsp 😅 

> or have me merge that branch into this PR so you can review both at once?

I don't mind reviewing both, but in this case I'm fine with both ideas, feel free to choose

ericdallo:
On another thought, maybe merge into one PR would be better @mainej as you will undo a lot of things you would on this PR right?

mainej:
@ericdallo after playing with the build from the first PR I realized that it wasn't working by itself. Since the tests clobber both the db* _and_ the channels, you need both fixes together. I've merged the componentize-channels branch into this one, and updated the PR title and description. If you'd like to review separately, you can look over the commits.

Nothing requires any of the test namespaces, so they aren't loaded when you connect to the nREPL. For me `cider-ns-refresh` and `cider-ns-reload-all` don't load them either. I have to `cider-eval-buffer` on them before running their tests. But it's still a better experience than before.

cli/test/clojure_lsp/main_test.clj
```diff
@@ -5,7 +5,7 @@
    [clojure.java.io :as io]
    [clojure.test :refer [deftest is testing]]))
 
-(h/reset-db-after-test)
+(h/reset-components-before-test)
```
mainej, line 8: This has always been incorrectly named—the reset happened before each test, not after.

mainej:
In the longer term I'd like to turn settings into a component too, as we've discussed elsewhere. Then it may be possible to write all the tests in a different style... At the beginning of the test, you'd construct the components as a local variable. You'd pass it into the test subject functions. And then you'd assert that its db* had been changed or that its channels had messages. Someday.

mainej:
@ericdallo did you want to do a final review of this, or is it ready to go?

cli/src/clojure_lsp/server.clj
```diff
@@ -516,7 +516,11 @@
           components {:db* db*
                       :logger timbre-logger
                       :producer producer
-                      :server server}]
+                      :server server
+                      :current-changes-chan (async/chan 1)
+                      :diagnostics-chan (async/chan 1)
+                      :created-watched-files-chan (async/chan 1)
+                      :edits-chan (async/chan 1)}]
```
ericdallo, line 523: nice

ericdallo (approved):

PR #1224, opened 2022-09-01
End dep-graph-queries experiment

mainej:
The experiment is considered successful; now queries use the dep-graph whenever possible.

This PR also moves some helpers from the `queries` namespace to the `dep-graph` namespace. These helpers used to need the analysis, which is why they were in `queries`, but now only need the dep-graph. 

In the process it fixes a bug in `clojure-lsp.api`. Before if a namespace was given to `format`, `clean-ns` or `diagnostics`, only one file that defined that namespace was updated. Now all files that define the namespace are. Since most namespaces are defined in only one file, this bug was probably rare, although there's a chance it was affecting people who have pairs of clj/cljs files that define the same ns. See `internal-api/nses->ns+uri` and the deleted `queries/nses-some-internal-uri`.


ericdallo (approved):
🚀

PR #1230, opened 2022-09-06
Drag params

### LINKED ISSUE #1131: Drag forward/backward in defn arg list should change calling sites as well
When editing a function definition, it's cumbersome to rearrange the order of the arguments, as this requires finding all places where it's called and editing them similarly as well.

Expand the code actions "Drag forward" and "Drag backward" to apply their changes to all calling sites as well.

Instead of using Drag forward/backward, you could create whole new code actions.

**Additional context**
Things to think about: What if the call site is in a threading macro? What if the function is used in an `apply` or `partial`?

mainej:
Adds a refactoring when in the params vector of a defn or defmacro that offers to **Drag params**. 

Invoking the command will move the param in the direction indicated, and also move the args at the var's call sites. The refactoring limits its actions to situations where it believes it can perform the refactoring correctly. If any of the call sites cannot be refactored, the whole refactoring is aborted, and the user is shown a warning.

* Does not offer refactoring in multi-arity functions
* Does not offer refactoring when on vararg
* Does not perform refactoring when drag would take an arg into a vararg position.
* Does not perform refactoring when the var is referred to outside of function-call position, e.g. in `{:f f}`.
* Does not perform refactoring when the var is used in `partial` unless this can be done safely. To be safe the params that are being swapped must not be "partialled". E.g. for the function `(defn f [a b c])`, the form `(partial f 1 2)` can be refactored if `b` is being swapped with `a`, to make `(partial f 2 1)`, but not if it is being swapped with `c`.
* Does not perform refactoring when the var is used in apply `(apply f 1 2 [3])`. It may be possible someday to augment this refactoring to treat `apply` like `partial`, performing the refactoring in situations that can be deemed "safe". Again, in this example, `a` and `b` could be safely swapped, but not `b` and `c`.
* Does not perform refactoring when the function is used in a threading form, e.g. `(-> 1 (f 2 3))` or `(->> 3 (f 1 2))`. It may be possible someday to augment this refactoring to handle some threading scenarios. In the thread-first example, `b` and `c` could be safely swapped, but not `a` and `b`, whereas it's the reverse in the thread-last example.

Fixes #1131


ericdallo (approved):
Looks good and works great, I felt a little bit confused when we present both the existing `Drag forward` and `Drag param forward`, maybe we could hide the old one when the param one is available? not sure if it'd not make sense to do that.
Anyway we could improve in the future, thanks

PR #1231, opened 2022-09-07
Indexes analysis and findings by URI

### LINKED ISSUE #1207: URIs vs Filenames
### History

The LSP spec uses URIs to identify files. clj-kondo uses filenames. clojure-lsp uses both—URIs most of the time, but filenames when running queries on the analysis and findings. At the moment there are over 100 places in the code where we convert a URI to a filename, or vice versa, as we switch between paradigms.

### Proposal

I propose that we use URIs everywhere, except on the boundary with clj-kondo.

1. Convert URIs to filenames as we're sending data off to clj-kondo.
2. Convert filenames to URIs in the kondo normalization. Index analysis and findings by URI.
3. Convert the rest of the code to rely on URIs exclusively.

### Context

I recall seeing a discussion about teaching clj-kondo to take and return URIs, which I can't find now. I believe the rational at that point was the possibility of using clojure-lsp across a network, for example for remote pairing. Even if those changes are done in clj-kondo someday, we'd still have to update clojure-lsp as described here to take advantage of it.

### Concerns

I'm writing this Issue primarily to collect concerns. @ericdallo, @snoe and perhaps @borkdude, I assume you've thought about this at some point. Do you know of any roadblocks to doing this work? Or has it just been a low priority?

I haven't done an exhaustive analysis of the ways we use filenames. There may be other places besides our interaction with clj-kondo where we use filenames. Are you aware of any such places?

mainej:
This does URI/filename conversion at the boundary with clj-kondo. Analysis and findings are indexed by URI. That lets most of the rest of clojure-lsp use uris, which it receives from LSP clients, without needing to convert to filenames.

I'd suggest reviewing the PR with [difftastic](https://github.com/Wilfred/difftastic) or a similar structural diffing tool—there are many renames of `filename` to `uri`, that don't have any other significant changes, but which GitHub over-emphasizes.

Fixes #1207, or at least does the bulk of the work. There are still several places where we do conversion or need filenames for some reason or another. I'll annotate those spots in the PR. There are a few things we could consider to further reduce the amount of conversion that happens:
* We could calculate `project-root-path` during init and store it in the db. There are several places that need this directory name (as opposed to `project-root-uri`, which is already stored in the db).
* We could reason through the algorithms that detect which "source-path" a file is in to decide if they could be switched from using filenames to using URIs. As an example of such an algorithm, when creating a new test file we figure out a subdirectory for it, and then generate a namespace based on that directory/filename. I think this could be done using URIs, but I'm hesitant to include further refactorings in this PR.

### DIFF
--- a/lib/src/clojure_lsp/db.clj
+++ b/lib/src/clojure_lsp/db.clj
@@ -12,11 +12,10 @@
 
 (def initial-db {:documents {}
                  :processing-changes #{}
-                 :dep-graph {}
-                 :file-meta {}})
+                 :dep-graph {}})
 (defonce db* (atom initial-db))
 
-(def version 5)
+(def version 6)
 
 (defn ^:private sqlite-db-file [project-root]
   (io/file (str project-root) ".lsp" ".cache" "sqlite.db"))
--- a/lib/src/clojure_lsp/feature/call_hierarchy.clj
+++ b/lib/src/clojure_lsp/feature/call_hierarchy.clj
@@ -23,6 +23,8 @@
              (name el-name))
      :kind (f.document-symbol/element->symbol-kind parent-element)
      :tags (cond-> [] deprecated (conj 1))
+     ;; TODO Consider using URI for display purposes, especially if we support
+     ;; remote LSP connections
      :detail (or (some-> ns str) (shared/uri->filename el-uri))
      :uri uri
      :range (shared/->range parent-element)
--- a/lib/src/clojure_lsp/feature/call_hierarchy.clj
+++ b/lib/src/clojure_lsp/feature/call_hierarchy.clj
@@ -17,20 +17,20 @@
 
 (defn ^:private element-by-uri->call-hierarchy-item
   [{:keys [uri parent-element usage-element]}]
-  (let [{:keys [ns filename arglist-strs deprecated] el-name :name} parent-element]
+  (let [{:keys [ns arglist-strs deprecated] el-uri :uri el-name :name} parent-element]
     {:name (if arglist-strs
              (str (name el-name) " " (some->> arglist-strs (remove nil?) (string/join " ")))
              (name el-name))
      :kind (f.document-symbol/element->symbol-kind parent-element)
      :tags (cond-> [] deprecated (conj 1))
-     :detail (or (some-> ns str) filename)
+     :detail (or (some-> ns str) (shared/uri->filename el-uri))
      :uri uri
      :range (shared/->range parent-element)
      :selection-range (shared/->range usage-element)}))
 
 (defn prepare [uri row col db*]
   (let [db @db*
-        cursor-element (q/find-element-under-cursor db (shared/uri->filename uri) row col)]
+        cursor-element (q/find-element-under-cursor db uri row col)]
     [(element-by-uri->call-hierarchy-item
        {:uri uri
         :usage-element cursor-element
@@ -39,21 +39,21 @@
 (defn ^:private parent-var-def
   "Returns var def analysis element that surrounds the element at row/col (or
   that is the element at that position)."
-  [root-zloc db filename row col]
+  [root-zloc db uri row col]
   (when-let [{:keys [row col]} (some-> root-zloc
                                        (parser/to-pos row col)
                                        (edit/find-var-definition-name-loc)
                                        z/node
                                        meta)]
-    (when-let [definition (q/find-element-under-cursor db filename row col)]
+    (when-let [definition (q/find-element-under-cursor db uri row col)]
       (when (or (identical? :var-definitions (:bucket definition))
                 (and (identical? :var-usages (:bucket definition))
                      (:defmethod definition)))
         definition))))
 
 (defn ^:private element->incoming-usage-by-uri
-  [db {:keys [uri root-zloc]} {:keys [name-row name-col filename] :as element}]
-  (when-let [parent-element (parent-var-def root-zloc db filename name-row name-col)]
+  [db root-zloc {:keys [name-row name-col uri] :as element}]
+  (when-let [parent-element (parent-var-def root-zloc db uri name-row name-col)]
     {:uri uri
      :usage-element element
      :parent-element parent-element}))
@@ -61,39 +61,30 @@
 (defn ^:private element->outgoing-usage-by-uri
   [db element]
   (when-let [definition (q/find-definition db element)]
-    (let [def-filename (:filename definition)
-          definition-uri (if (shared/plain-uri? def-filename)
-                           def-filename
-                           (shared/filename->uri def-filename db))]
-      {:uri definition-uri
-       :usage-element element
-       :parent-element definition})))
+    {:uri (:uri definition)
+     :usage-element element
+     :parent-element definition}))
 
 (defn incoming [uri row col {:keys [db*] :as components}]
   (let [db @db*
-        references (q/find-references-from-cursor db (shared/uri->filename uri) row col false)
-        filenames (->> references (map :filename) distinct)
-        file-meta (zipmap filenames
-                          (map (fn [filename]
-                                 (let [uri (shared/filename->uri filename db)]
-                                   {:uri uri
-                                    :root-zloc (safe-zloc-of-forced-uri components uri)}))
-                               filenames))
+        references (q/find-references-from-cursor db uri row col false)
+        uris (->> references (map :uri) distinct)
+        root-zlocs (zipmap uris
+                           (map #(safe-zloc-of-forced-uri components %) uris))
         db @db*]
     (->> references
          (keep (fn [element]
-                 (element->incoming-usage-by-uri db (get file-meta (:filename element)) element)))
+                 (element->incoming-usage-by-uri db (get root-zlocs (:uri element)) element)))
          (mapv (fn [element-by-uri]
                  {:from-ranges []
                   :from (element-by-uri->call-hierarchy-item element-by-uri)})))))
 
 (defn outgoing [uri row col {:keys [db*] :as components}]
-  (let [filename (shared/uri->filename uri)
-        root-zloc (safe-zloc-of-forced-uri components uri)
+  (let [root-zloc (safe-zloc-of-forced-uri components uri)
         db @db*]
-    (when-let [definition (parent-var-def root-zloc db filename row col)]
+    (when-let [definition (parent-var-def root-zloc db uri row col)]
       (->> (q/find-var-usages-under-form db
-                                         filename
+                                         uri
                                          (:name-row definition)
                                          (:name-col definition)
                                          (:end-row definition)
--- a/lib/src/clojure_lsp/feature/code_lens.clj
+++ b/lib/src/clojure_lsp/feature/code_lens.clj
@@ -20,40 +20,31 @@
 (defn ^:private test-references->string [references]
   (references->string references " test"))
 
-(defn ^:private var-definitions-lens [db filename]
-  (->> (q/find-var-definitions db filename true)
-       (remove (partial q/exclude-public-definition? (:kondo-config db)))))
-
-(defn ^:private keyword-definitions-lens
-  [db filename]
-  (->> (q/find-keyword-definitions db filename)
-       (remove (partial q/exclude-public-definition? (:kondo-config db)))))
-
 (defn reference-code-lens [uri db]
-  (let [filename (shared/uri->filename uri)]
-    (into []
-          (map (fn [element]
-                 {:range (shared/->range element)
-                  :data  [uri (:name-row element) (:name-col element)]}))
-          (concat (q/find-namespace-definitions db filename)
-                  (var-definitions-lens db filename)
-                  (keyword-definitions-lens db filename)))))
+  (into []
+        (map (fn [element]
+               {:range (shared/->range element)
+                :data  [uri (:name-row element) (:name-col element)]}))
+        (concat (q/find-namespace-definitions db uri)
+                (->> (concat (q/find-var-definitions db uri true)
+                             (q/find-keyword-definitions db uri))
+                     (remove (partial q/exclude-public-definition? (:kondo-config db)))))))
 
-(defn test-reference? [source-path {:keys [filename]}]
-  (and source-path
-       (not (string/starts-with? filename source-path))
-       (string/includes? filename "_test.")))
+(defn ^:private test-reference? [source-uri reference-uri]
+  (and source-uri
+       ;; when in test file, don't count usages of helpers as test references
+       (not (string/starts-with? reference-uri source-uri))
+       (string/includes? reference-uri "_test.")))
 
 (defn resolve-code-lens [uri row col range db]
-  (let [filename (shared/uri->filename uri)
-        segregate-lens? (settings/get db [:code-lens :segregate-test-references] true)
-        references (q/find-references-from-cursor db filename row col false)]
+  (let [segregate-lens? (settings/get db [:code-lens :segregate-test-references] true)
+        references (q/find-references-from-cursor db uri row col false)]
     (if segregate-lens?
-      (let [source-path (->> (settings/get db [:source-paths])
-                             (filter #(string/starts-with? filename %))
-                             first)
-            main-references (filter (complement (partial test-reference? source-path)) references)
-            test-references (filter (partial test-reference? source-path) references)]
+      (let [source-uri (some-> uri
+                               (shared/uri->source-path (settings/get db [:source-paths]))
+                               (shared/filename->uri db))
+            main-references (remove (comp (partial test-reference? source-uri) :uri) references)
+            test-references (filter (comp (partial test-reference? source-uri) :uri) references)]
         (if (seq test-references)
           {:range range
            :command {:title (str (main-references->string main-references)
--- a/lib/src/clojure_lsp/feature/diagnostics.clj
+++ b/lib/src/clojure_lsp/feature/diagnostics.clj
@@ -1,12 +1,12 @@
 (ns clojure-lsp.feature.diagnostics
   (:require
    [clj-kondo.impl.config :as kondo.config]
+   [clojure-lsp.dep-graph :as dep-graph]
    [clojure-lsp.logger :as logger]
    [clojure-lsp.queries :as q]
    [clojure-lsp.settings :as settings]
    [clojure-lsp.shared :as shared]
-   [clojure.core.async :as async]
-   [clojure.java.io :as io]))
+   [clojure.core.async :as async]))
 
 (set! *warn-on-reflection* true)
 
@@ -39,7 +39,7 @@
         kondo-config (if (:ns element)
                        (kondo-config-for-ns kondo-config (:ns element) (:filename element))
                        kondo-config)]
-    {:filename (:filename element)
+    {:uri (:uri element)
      :row (:name-row element)
      :col (:name-col element)
      :end-row (:name-end-row element)
@@ -88,15 +88,16 @@
     (or (and row col)
         (logger/warn "Invalid clj-kondo finding. Cannot find position data for" finding))))
 
-(defn ^:private exclude-ns? [filename linter db]
-  (when-let [namespace (shared/filename->namespace filename db)]
+(defn ^:private exclude-ns? [uri linter db]
+  ;; TODO: it's possible, though unusual, to have several namespaces in a file.
+  ;; What should we do in that case?
+  (when-let [namespace (first (dep-graph/ns-names-for-uri db uri))]
     (when-let [ns-exclude-regex-str (settings/get db [:linters linter :ns-exclude-regex])]
       (re-matches (re-pattern ns-exclude-regex-str) (str namespace)))))
 
-(defn ^:private kondo-findings->diagnostics [filename linter db]
-  (when-not (exclude-ns? filename linter db)
-    (->> (get (:findings db) filename)
-         (filter #(= filename (:filename %)))
+(defn ^:private kondo-findings->diagnostics [uri linter db]
+  (when-not (exclude-ns? uri linter db)
+    (->> (get-in db [:findings uri])
          (filter valid-finding?)
          (mapv kondo-finding->diagnostic))))
 
@@ -112,10 +113,12 @@
     2 :yellow
     3 :cyan))
 
-(defn ^:private clj-depend-violations->diagnostics [filename level db]
-  (when-let [namespace (shared/filename->namespace filename db)]
+(defn ^:private clj-depend-violations->diagnostics [uri level db]
+  ;; TODO: it's possible, though unusual, to have several namespaces in a file.
+  ;; What should we do in that case?
+  (when-let [namespace (first (dep-graph/ns-names-for-uri db uri))]
     (mapv (fn [{:keys [message]}]
-            (let [ns-definition (q/find-namespace-definition-by-filename db filename)]
+            (let [ns-definition (q/find-namespace-definition-by-uri db uri)]
               {:range (shared/->range ns-definition)
                :tags []
                :message message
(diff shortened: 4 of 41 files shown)


lib/src/clojure_lsp/feature/diagnostics.clj
```diff
@@ -128,48 +131,49 @@
...
-(defn publish-all-diagnostics! [paths {:keys [db*], :as components}]
+(defn publish-all-diagnostics! [uris {:keys [db*], :as components}]
   (let [db @db*]
     (publish-all-diagnostics!*
       components
-      (eduction (map io/file)
-                (mapcat file-seq)
-                (map #(.getAbsolutePath ^java.io.File %))
-                (map #(shared/filename->uri % db))
-                (remove #(= :unknown (shared/uri->file-type %)))
```
mainej, line 161: I'd like a more careful review of this. The `crawler` used to pass `publish-all-diagnostics!` the `:source-paths` directly, which is just a list of directories like `["src", "test"]`. We used `file-seq` to figure out what files are in those directories, removed the non-Clojure files, and published their diagnostics. But this is overkill I think. By the time `publish-all-diagnostics!` is called, we've already done the initial project analysis. During that process clj-kondo has figured out which files are in those directories, and analyzed whatever it can. Those files are all represented in the `:findings` map in the db. So we might as well just publish the diagnostics we have for those files—i.e., all the findings we have. I've switched the code in `crawler` to do that and confirmed that the other uses of `publish-all-diagnostics!` still work as expected. Here I've removed the `file-seq` related code and assumed we'll be passed URIs of files, not paths. I've left in the `(remove #(= :unknown (shared/uri->file-type %)))`, but I think that could be removed too. How would we have findings for an `:unknown` file? So, 1) does it seem safe to remove `file-seq`? And 2) do we need the `(remove :unknown ...)` code?

lib/src/clojure_lsp/feature/document_symbol.clj
```diff
@@ -31,17 +31,17 @@
 (defn ^:private symbol-order [{:keys [selection-range]}]
   [(:line (:start selection-range)) (:character (:start selection-range))])
 
-(defn document-symbols [db filename]
-  (let [namespace-definition (q/find-namespace-definition-by-filename db filename)]
+(defn document-symbols [db uri]
+  (let [namespace-definition (q/find-namespace-definition-by-uri db uri)]
     [{:name (or (some-> namespace-definition :name name)
-                filename)
+                (shared/uri->filename uri))
```
mainej, line 37: Filename is being used for display purposes here.

lib/src/clojure_lsp/feature/file_management.clj
```diff
@@ -249,42 +247,42 @@
...
-                 :delete-watched-file
-                 (file-deleted components uri (shared/uri->filename uri))))))
+  (let [{:keys [created changed deleted]} (group-by :type changes)
+        observe-changed? (settings/get @db* [:compute-external-file-changes] true)]
+    (doseq [created-or-changed (concat created (when observe-changed? changed))]
+      (async/>!! watched-files-chan (:uri created-or-changed)))
+    (when (seq deleted)
+      (shared/logging-task
+        :delete-watched-files
+        (files-deleted components (map :uri deleted))))))
```
mainej, line 278: Processing all the deleted URIs at once lets us use `onto-chan!`, which helps avoid "too many pending puts" exceptions.

lib/src/clojure_lsp/feature/rename.clj
```diff
@@ -192,62 +191,60 @@
...
-             :message "Can't rename, only namespaced keywords can be renamed."}}
-
-    :else
-    {:result :success}))
+(defn ^:private rename-status [db element]
+  (let [references (q/find-references db element true)
+        definition (q/find-definition db element)
+        client-capabilities (:client-capabilities db)
+        source-paths (settings/get db [:source-paths])
+        source-path (shared/uri->source-path (:uri definition) source-paths)]
```
mainej, line 199: Calculating the `definition`, `references` and `source-path` in `rename-status` lets us remove some duplication between `prepare-rename` and `rename-element`

lib/src/clojure_lsp/feature/workspace_symbols.clj
```diff
@@ -27,33 +27,25 @@
...
   (->> (q/find-internal-definitions db)
        (fuzzy-filter query)
        (map element->workspace-symbol)
-       ;; Group elements by file, but otherwise preserve ordering by search score.
-       ;; Also replace filename with uri.
-       (reduce (fn [{:keys [next-idx filenames] :as result} symb]
-                 (let [filename (:filename (:location symb))]
-                   (if-let [{:keys [idx uri]} (get filenames filename)]
-                     (update-in result [:members idx] conj (symbol-with-uri symb uri))
-                     (let [uri (shared/filename->uri filename db)]
```
mainej, line 49: Removes some code that let us limit the number of times we converted a particular filename to a URI.

lib/src/clojure_lsp/handlers.clj
```diff
@@ -93,19 +93,18 @@
           {}
           work-done-token
           components)
-        (let [db @db*]
+        (let [db @db*
+              internal-uris (dep-graph/internal-uris db)]
           (when (settings/get db [:lint-project-files-after-startup?] true)
-            (f.diagnostic/publish-all-diagnostics! (-> db :settings :source-paths) components))
+            (f.diagnostic/publish-all-diagnostics! internal-uris components))
```
mainej, line 99: This is the counter-part to the change to `publish-all-diagnostics!`. We publish all project findings, instead of re-calculating which files we might have findings for.

lib/src/clojure_lsp/internal_api.clj
```diff
@@ -251,27 +226,35 @@
...
-                         (map #(shared/filename->namespace (.getCanonicalPath ^File %) db)))
-                    (shared/filename->namespace (.getCanonicalPath filename-or-dir) db))))
-           flatten
-           (remove nil?)
-           (map symbol)
-           seq)
-      (into #{}
-            (remove (partial exclude-ns? options))
-            (dep-graph/internal-ns-names db))))
+(defn ^:private options->uris [{:keys [namespace filenames project-root] :as options} db]
```
mainej, line 229: The api accepts filenames or namespaces as a specification for what to clean or report on. As a fallback, it operates on the whole project. The code for figuring out what to do used to convert everything to namespaces, then find matching filenames. Now we convert everything to URIs first, and report on what namespaces were operated on afterward. This removes a lot of duplication and tricky code.

lib/src/clojure_lsp/kondo.clj
```diff
@@ -162,30 +157,38 @@
...
 (defn ^:private normalize
   "Put kondo result in a standard format, with `analysis` normalized and
-  `analysis` and `findings` indexed by filename."
+  `analysis` and `findings` indexed by uri."
   [{:keys [analysis findings] :as kondo-results}
-   {:keys [external? ensure-filenames filter-analysis] :or {filter-analysis identity}}]
-  (let [analysis (->> analysis
+   {:keys [external? ensure-uris filter-analysis] :or {filter-analysis identity}}
+   db]
+  (let [filename->uri (memoize #(shared/filename->uri % db))
```
mainej, line 164: This conversion is a new, and potentially expensive, step in normalization. I benchmarked startup and ongoing analysis in some large projects and it didn't seem to make an overall difference, partially because we used to incur this same expense in the dep-graph code, and no longer do. Still, memoization seems prudent, to avoid re-calculating the same URI over and over.

lib/src/clojure_lsp/refactor/transform.clj
```diff
@@ -735,49 +729,47 @@
...
+  (let [{:keys [uri bucket name-row name-col] :as definition} (q/find-definition-from-cursor db uri row col)]
+    (when (or (identical? :locals bucket)
+              (identical? :var-definitions bucket))
+      (when-let [zloc (some-> (parser/safe-zloc-of-file db uri)
+                              (parser/to-pos name-row name-col))]
+        (when-let [op (some-> zloc edit/find-op z/sexpr #{'let 'def})]
+          {:def-elem definition
+           :def-zloc zloc
+           :def-uri uri
+           :def-op op})))))
```
mainej, line 742: Again, remove some duplication between `inline-symbol?` (and code-actions) and `inline-symbol`.

lib/src/clojure_lsp/refactor/transform.clj
```diff
@@ -821,20 +813,20 @@
...
-  (let [filename (shared/uri->filename uri)
-        ;; TODO: shouldn't this also look for unaliased ns-usages?
+  (let [;; TODO: shouldn't this also look for unaliased ns-usages?
         ;; See https://github.com/clojure-lsp/clojure-lsp/issues/1023
-        ns-usage (q/find-namespace-usage-by-alias db filename (symbol ns-or-alias))
+        ns-usage (q/find-namespace-usage-by-alias db uri (symbol ns-or-alias))
         ns-definition (when ns-usage
                         (q/find-definition db ns-usage))
-        source-paths (settings/get db [:source-paths])
+        source-path (shared/uri->source-path uri (settings/get db [:source-paths]))
```
mainej, line 821: Another place we still use filenames—to figure out what URI a namespace should correspond to.

lib/test/clojure_lsp/features/code_lens_test.clj
```diff
@@ -114,4 +115,44 @@
                (h/file-uri "file:///a.clj") 1 13
                {:start {:line 1 :character 5}
                 :end {:line 1 :character 12}}
-               (h/db)))))))
+               (h/db))))))
+  (testing "test references"
```
mainej, line 119: I couldn't find unit tests for the splitting of test and main code-lens references, so this adds them.

lib/src/clojure_lsp/feature/call_hierarchy.clj
```diff
@@ -17,20 +17,20 @@
...
   [{:keys [uri parent-element usage-element]}]
-  (let [{:keys [ns filename arglist-strs deprecated] el-name :name} parent-element]
+  (let [{:keys [ns arglist-strs deprecated] el-uri :uri el-name :name} parent-element]
     {:name (if arglist-strs
              (str (name el-name) " " (some->> arglist-strs (remove nil?) (string/join " ")))
              (name el-name))
      :kind (f.document-symbol/element->symbol-kind parent-element)
      :tags (cond-> [] deprecated (conj 1))
-     :detail (or (some-> ns str) filename)
+     :detail (or (some-> ns str) (shared/uri->filename el-uri))
```
mainej, line 26: Filename is used for display purposes here.

lib/src/clojure_lsp/feature/diagnostics.clj
```diff
@@ -193,49 +198,28 @@
...
-                        project-db kondo-config)))
+        db-of-uris (update project-db :analysis select-keys uris)]
+    (unused-public-vars db-of-uris project-db kondo-config)))
 
 (defn ^:private finalize-findings! [findings reg-finding!]
-  (run! reg-finding! findings)
-  (group-by :filename findings))
+  (let [uri->filename (memoize shared/uri->filename)]
+    (run! #(reg-finding! (assoc % :filename (uri->filename (:uri %)))) findings))
+  nil)
```
mainej, line 215: This return value is unused.

lib/src/clojure_lsp/feature/file_management.clj
```diff
@@ -249,42 +247,42 @@
...
 
 (defn did-close [uri {:keys [db*], :as components}]
   (let [filename (shared/uri->filename uri)
         source-paths (settings/get @db* [:source-paths])]
     (when (and (not (shared/external-filename? filename source-paths))
                (not (shared/file-exists? (io/file filename))))
```
mainej, line 284: We still use filename here to decide whether a file exists on the local filesystem.

lib/src/clojure_lsp/feature/hover.clj
```diff
@@ -88,7 +88,8 @@
                    (if markdown?
                      (docstring->formatted-markdown doc)
                      doc))
-        clojuredocs (f.clojuredocs/find-hover-docs-for sym-name sym-ns db*)]
+        clojuredocs (f.clojuredocs/find-hover-docs-for sym-name sym-ns db*)
+        filename (shared/uri->filename uri)]
```
mainej, line 92: Filename is used for display purposes here.

lib/src/clojure_lsp/shared.clj
```diff
@@ -208,12 +208,21 @@
     (catch Exception _
       false)))
 
+(defn ^:private filename->source-paths [filename source-paths]
+  (filter #(string/starts-with? filename %) source-paths))
+
+(defn uri->source-paths [uri source-paths]
+  (filename->source-paths (uri->filename uri) source-paths))
```
mainej, line 215: If we had source-uris, this code wouldn't need to convert uri to filename.

ericdallo:
> We could calculate project-root-path during init and store it in the db. There are several places that need this directory name (as opposed to project-root-uri, which is already stored in the db).

Sounds like a cheap and easy change, so makes sense

> We could reason through the algorithms that detect which "source-path" a file is in to decide if they could be switched from using filenames to using URIs. As an example of such an algorithm, when creating a new test file we figure out a subdirectory for it, and then generate a namespace based on that directory/filename. I think this could be done using URIs, but I'm hesitant to include further refactorings in this PR.

Agreed, we could leave this kind of refactor to a next one

lib/src/clojure_lsp/db.clj
```diff
@@ -12,11 +12,10 @@
 
 (def initial-db {:documents {}
                  :processing-changes #{}
-                 :dep-graph {}
-                 :file-meta {}})
+                 :dep-graph {}})
 (defonce db* (atom initial-db))
 
-(def version 5)
+(def version 6)
```
ericdallo, line 18: Do we really need to bump the version if we are removing `file-meta`? we don't need to care if existing loaded db would have that unnecessary entry, right?

lib/src/clojure_lsp/feature/diagnostics.clj
```diff
@@ -88,15 +88,16 @@
     (or (and row col)
         (logger/warn "Invalid clj-kondo finding. Cannot find position data for" finding))))
 
-(defn ^:private exclude-ns? [filename linter db]
-  (when-let [namespace (shared/filename->namespace filename db)]
+(defn ^:private exclude-ns? [uri linter db]
+  ;; TODO: it's possible, though unusual, to have several namespaces in a file.
+  ;; What should we do in that case?
+  (when-let [namespace (first (dep-graph/ns-names-for-uri db uri))]
```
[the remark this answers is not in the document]
ericdallo: Yeah, it's an issue indeed, we could discuss in an issue but supporting diagnostics for all ns of that file would be nice

lib/src/clojure_lsp/feature/diagnostics.clj
```diff
@@ -193,49 +198,28 @@
...
-                        (file-kw-definitions project-db filename)
-                        project-db kondo-config)))
+        db-of-uris (update project-db :analysis select-keys uris)]
+    (unused-public-vars db-of-uris project-db kondo-config)))
 
 (defn ^:private finalize-findings! [findings reg-finding!]
-  (run! reg-finding! findings)
-  (group-by :filename findings))
+  (let [uri->filename (memoize shared/uri->filename)]
+    (run! #(reg-finding! (assoc % :filename (uri->filename (:uri %)))) findings))
```
ericdallo, line 214: If kondo support uri internally in the future we could remove this in the future

lib/src/clojure_lsp/feature/file_management.clj
ericdallo: maybe we should have a `uri-exists?` as we will end up using more and more uris and there are other places that need to do similar checks if file exists too

lib/src/clojure_lsp/feature/hover.clj
ericdallo: Maybe this one we could show the URI? we would need to check how clients behave with that, I can see relevant if I'm using a remote LSP connection (we could support it in the future), showing the URI would be better than showing a filename I guess

ericdallo: We could leave to a next PR too

lib/src/clojure_lsp/feature/rename.clj
ericdallo: good catch

lib/src/clojure_lsp/handlers.clj
ericdallo: Looks better indeed

lib/src/clojure_lsp/internal_api.clj
```diff
@@ -304,8 +284,10 @@
          :message (edits->diff-string edits options db)
          :edits edits}
         (do
-          (mapv (comp #(cli-println options "Cleaned" (uri->ns (:uri %) ns+uris))
-                      #(apply-workspace-edit-summary! % components)) edits)
+          (run! #(apply-workspace-edit-summary! % components) edits)
+          (->> edits
+               (mapcat #(dep-graph/ns-names-for-uri db (:uri %)))
+               (run! #(cli-println options "Cleaned" %)))
```
ericdallo, line 290: Wouldn't this change make clojure-lsp first apply all edits and only then print all applied changes? if so, maybe the user xp would be affected since we would lose the visual feedback that clojure-lsp cleaned 3/10 files and will only know that after cleaning the last 10 file, am I reading that correctly?

lib/src/clojure_lsp/kondo.clj
```diff
@@ -123,17 +124,11 @@
 
     :java-class-definitions
     [(-> element
-         (dissoc :uri)
```
ericdallo, line 126: I can't recall why I decided to dissoc uri, not sure if was only related to not store unused info on the db 🤔

lib/src/clojure_lsp/kondo.clj
ericdallo: Looks ok, but when we memoize we memoize by all the args of the function right? so we would memorize given that db which could be a huge map, and memoizing lots of URIs for the same big db could affect application memory, right?

ericdallo: But I suppose you tested performance because of that?

ericdallo: Hum, this function is called once per call, so memoizing once may not be that bad I guess

lib/src/clojure_lsp/kondo.clj
```diff
@@ -162,30 +157,38 @@
...
+  `analysis` and `findings` indexed by uri."
   [{:keys [analysis findings] :as kondo-results}
-   {:keys [external? ensure-filenames filter-analysis] :or {filter-analysis identity}}]
-  (let [analysis (->> analysis
+   {:keys [external? ensure-uris filter-analysis] :or {filter-analysis identity}}
+   db]
+  (let [filename->uri (memoize #(shared/filename->uri % db))
+        trade-filename-for-uri (fn [obj]
+                                 (-> obj
+                                     (assoc :uri (filename->uri (:filename obj)))
```
ericdallo, line 167: Maybe we should assoc only if kondo already didn't return a URI for that obj?

ericdallo: Maybe overhead for now, so it's ok

lib/test/clojure_lsp/features/code_lens_test.clj
ericdallo: thanks!

lib/src/clojure_lsp/feature/diagnostics.clj
mainej: Ah, I didn't realize didChangeWatchedFiles could include directories, but the spec does mention that. Hmm... I'll look into restoring some of this.

lib/src/clojure_lsp/kondo.clj
mainej: Yeah, this is a little tricky to read, but it seems to keep performance good and doesn't affect memory.
1. You're correct, we create a memoized function every time kondo is called. After we're done normalizing the analysis data from that one run we discard the memoized function, releasing the memory consumed by its arg/return-value cache. On the next run, we create a new memoized function. Using `memoize` seemed easier to read than using `loop` or `reduce` to collect a hashmap of filenames to uris. We've used that trick elsewhere to reduce the number of conversions, and it would be possible here too, just harder to read. The other benefit of `memoize` is that its cache is shared between the analysis normalization and the findings normalization. Again, the `loop` or `reduce` trick could do the same thing, it'd just be trickier.
2. The thing that is getting memoized is a function of one argument—the filename. The `db` is closed over, but isn't an argument, so the db isn't part of the cache key. But even if it were, as noted above, that cache is discarded at the end of normalization.

lib/src/clojure_lsp/feature/file_management.clj
mainej: Yep, I'd love to add this someday. There's another review of the code to be done to clean up this kind of thing. The dep-graph includes `:internal?` so we might be able to reduce the number of uses of `external-filename?` too.

lib/src/clojure_lsp/feature/hover.clj
mainej: I'll add some TODO comments so we remember to look at this again in a future PR.

lib/src/clojure_lsp/db.clj
mainej: We need to re-index the external `analysis` by URI.

lib/src/clojure_lsp/feature/diagnostics.clj
mainej: Yep, that's true. I think `clojure-lsp.feature.diagnostics` and `clojure-lsp.kondo` will need some changes when that happens, but hopefully not nearly as many other places.

lib/src/clojure_lsp/internal_api.clj
mainej: I had trouble reading this too, which is why I re-wrote it. Actually, it's doing the same thing it's always done. That is, it always changed the files before printing. The thing to remember is that `comp` runs the _rightmost_ function first. So the old code first ran `apply-workspaced-edit-summary!` (which returned the edits it was passed) and then second ran `cli-println` based on those edits. But the `comp` is a weird way to do that. I also switched from `mapv` to `run!` because I think it's important to call out that these are being called for their side-effects. It's true that `mapv` forces side-effects because it has to realize a vector, but that's more obscure than `run!` or `doall`.

lib/src/clojure_lsp/kondo.clj
mainej: I think I remember seeing nil `:uri` being returned by clj-kondo? Maybe. I can't remember. Anyway, saving space or avoiding weird data could both be reasons we were dissocing uri. But now we add uri before calling this code, so we definitely don't want to dissoc it.

lib/src/clojure_lsp/kondo.clj
mainej: I hope that kondo will start returning URI and stop returning filename, in which case this code can be removed entirely. But we can figure out those details when the time comes.

lib/src/clojure_lsp/feature/file_management.clj
mainej: We're already changing the order somewhat by grouping together several changes into one analysis batch. But I think you're right—order doesn't matter much for didChangeWatchedFiles, and it's best to treat everything as one big batch if possible.

lib/src/clojure_lsp/kondo.clj
ericdallo: Oh, that makes sense, so it's completely fine!

lib/src/clojure_lsp/feature/diagnostics.clj
mainej: I added `file-seq` back in, but only for `didChangeWatchedFiles`. I did that a higher level, at the beginning of `clojure-lsp.feature.file-management/did-change-watched-files`, so that `analyze-watched-files!` and `files-deleted` both receive file URIs, not directory URIs. It looks like the code in those functions (including, but not limited to `publish-all-diagnostics!`) expects file URIs.

lib/src/clojure_lsp/feature/code_lens.clj
```diff
@@ -20,40 +20,31 @@
...
     (if segregate-lens?
-      (let [source-path (->> (settings/get db [:source-paths])
-                             (filter #(string/starts-with? filename %))
-                             first)
-            main-references (filter (complement (partial test-reference? source-path)) references)
-            test-references (filter (partial test-reference? source-path) references)]
+      (let [source-uri (some-> uri
+                               (shared/uri->source-path (settings/get db [:source-paths]))
+                               (shared/filename->uri db))
+            main-references (remove (comp (partial test-reference? source-uri) :uri) references)
```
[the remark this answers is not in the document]
ericdallo: Sounds good

lib/src/clojure_lsp/internal_api.clj
ericdallo: You are right, I missed that `comp`. Yeah, replacing all mapvs with run! seems way better, totally agreed with that

ericdallo (approved):
Looks great @mainej

Issue #1211, opened 2022-08-27
First class dev/test support for MS-Windows


ikappaki:
**Is your feature request related to a problem? Please describe.**

At the moment, the lib/cli test and cli integration-test command fail with a lot of errors on MS-Windows.

In addition it is impossible to create debug/prod cli builds because the jar to binary conversion is not compatible with MS-Windows.

This means that the MS-Windows builds are likely to be dysfunctional and makes the release verification and the development cycle very difficult to manage.

**Describe the solution you'd like**

Fix all the test errors and make the development and CI integration on MS-Windows the same or very close to what is provided for Linux.

**Describe alternatives you've considered**

Can't think of any.

**Additional context**

The following is a non-exhaustive list that is likely to need addressing on MS-Windows
- Fix tests
- Fix integration tests
- Support for jar to executable conversion.
- Migrate makefile to bb.edn
- Update Github actions to run the same tests on MS-Windows.

Happy to have a look at all of the above.

Thanks

ericdallo:
Totally agree with that! I'm no windows user but LMK if need any support or context, your help is appreciated

ericdallo:
@ikappaki is this done after all your recent improvements?

ikappaki:
> @ikappaki is this done after all your recent improvements?

Hi @ericdallo, the only thing left is to correct the formatting of the writing tests section of the documentation here, which some how comes back garbled: https://clojure-lsp.io/development sorry haven't figured out yet how this page is generated and how to "debug" it ... :(

ericdallo:
No problem, I can take a look, usually running `make local-webpage` should build the docs locally

ikappaki:
> No problem, I can take a look, usually running `make local-webpage` should build the docs locally

yeah, it asks me for a password and I don't have a docker account ...

(btw it is `bb local-webpage` now :) )

ericdallo:
Oh, I think that's because the docker image comes from [here](https://github.com/clojure-lsp/docs-image), I should make it public somehow

yeah haha, just fixed on master

ericdallo:
The image is public, but you need to make your docker login on github to download images from github :/

ikappaki:
> The image is public, but you need to make your docker login on github to download images from github :/
> https://github.com/clojure-lsp/docs-image/pkgs/container/docs-image%2Fdocs-image

... :)

ikappaki:
[...]

PR #1232, opened 2022-09-08
Replace Makefile and MS-Windows scripts with bb tasks

ikappaki:
Hi,

could you please consider patch to replace the Makefile and windows scripts with bb tasks. This is first step of the attempt discussed in #1211 to align the MS-Windows builds and dev cycle with that of Linux.

The `make` tool is not readily available nor fully compatible with the MS-Windows, thus windows users need to do extra work to develop clojure-lsp, thus it is much easier if we have a common build tool.

Changes include
- Replicating the `Makefile` targets as close as possible in `scripts/make.clj`.
- Changing the GH actions and other script to replace `make xxx` invocations with `bb xxx` invocations.
- Changing documentation to refer to `bb xxx` instead of `make xxx`
- Removing `Makefile` and the `native-windows-compile.bat` script.

The user experience should remain the same as before, but instead of running `make` users should now run `bb`.  

The only thing I could not replicate is the plain `make` command which invokes `make debug-cli`. Running `bb` will just lists the available tasks instead. Should we add a `bb cli` alias instead for `bb debug-cli` to easy the transition?

The following workflows have been tested to work:
- `ci.yml`.
- `release.yml`, except the release jobs because they require tokens.
- `nightly.hml`, except those parts that require GH tokens.

The following have not been tested:
- `cirrus.yml`  - not sure how to test this.
- `bump_downstreams.yml`, because it requires authentication.

All the untested elements mentioned above shall be tested I believe on commit by the build infrastructure. Any unexpected issues should be easy to fix, since the changes are very simple in nature (installing the bb script where it does not exist, changing `make` to `bb`).

Thanks


CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
   - Add support to decompile jar as a java project when finding the definition of a java class, allowing to have LSP features on that java project. #1187
   - Add refactorings similar to `Sort map`: `Sort vector`, `Sort list`, `Sort set`, and `Sort clauses` for functions like `assoc`. #1155
 
+- Build
+  - Replace Makefile and MS-Windows scripts with bb tasks. #1211
```
ericdallo, line 14: Usually, we add an entry to refer to changes that affect users, this is not one that will affect the final user so not sure it's worth polluting the changelog with these changes, WDYT?

ericdallo (approved):
I think we could remove the changelog, besides that looks really good, thank you for the help!

CHANGELOG.md
ikappaki: Done, hopefully existing developers won't have much trouble figuring out the move to bb :)

ikappaki:
> I think we could remove the changelog, besides that looks really good, thank you for the help!

Done! thank you for the review!

PR #1236, opened 2022-09-11
Fix unit tests on MS-Windows

ikappaki:
Hi,

could you please consider patch to fix the unit tests on MS-Windows and any codebase fixes resulting from it. This is part of providing windows support as per #1211.

Most of the unit test updates are about using the
`test-helper/file-path` and `test-helper/file-uri` function around test paths where it is missing.

Other important changes
1. Introduction of `test-helper/string=` function to use in text comparisons that treats all Line Endings (LE) as equal. Line endings can be different on unix (`\n`) vs windows (`\n\r`) which might mess up with the testing.
2. Enhance test-helper/assert-submaps so that it accepts a `set` as input to disregard order of items. It turns out some lsp vector results are coming out of maps, and thus their order is unpredictable. See `handler-test/document-symbol` for an example.
~~3. Use the Apache Commons IO as a new dependency to deal with path separators in a uniformly way across the different architectures where needed. Most notable use is fixing `shared/namespace+source-path->filename`. Note that this brings a new library in but uses just a small portion of it.~~
4. `cljfmt/reformat-string` is made aware of line endings.
5. The default case for drive letter across tests and in shared/conform-uri is now uppercase. This is to align with how java always treat the drive letter in the `.toURI` calls. Thus we uniformly handle this in the codebase and test as such. Not sure what external impact would changing the default case have to external clients.
6. Fixes issue with the custom `file:*.jar:subpath` syntax in `shared/uri->filename`, since that pattern is not valid on windows.
7. Update `ci.yml` to unit test on windows but only on JDK-8, and add macos as an added bonus.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - General
+  - Fix unit tests on MS-Windows. #1211
   - End dep-graph-queries experiment; clojure-lsp now uses the dep-graph to optimize queries whenever possible
   - Bump clj-kondo to `2022.09.08`. 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 - General
-  - Fix unit tests on MS-Windows. #1211
+  - Improve MS-Windows support by fixing various path, URI and line ending issues coming out of repairing the unit tests suite on windows. #1211
   - End dep-graph-queries experiment; clojure-lsp now uses the dep-graph to optimize queries whenever possible
   - Bump clj-kondo to `2022.09.08`. 
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,11 +44,17 @@ jobs:
         run: bb lint-diagnostics
 
   unit-test:
-    runs-on: ubuntu-latest
+    runs-on: ${{matrix.os}}
     strategy:
       fail-fast: false
       matrix:
+        os: [ubuntu-latest]
         jdk: [8, 11, 15, 17]
+        include:
+          - os: windows-latest
+            jdk: 8
+          - os: macos-latest
+            jdk: 8
     steps:
       - uses: actions/checkout@v2.2.0
         with:
--- a/common-test/src/clojure_lsp/test_helper.clj
+++ b/common-test/src/clojure_lsp/test_helper.clj
@@ -23,7 +23,13 @@
 
 (defn file-uri [uri]
   (cond-> uri windows?
-          (string/replace #"^(file|zipfile|jar:file):///(?!\w:/)" "$1:///c:/")))
+          (string/replace #"^(file|zipfile|jar:file):///(?!\w:/)" "$1:///C:/")))
+
+(defn string=
+  "Like `clojure.core/=` applied on STRING1 and STRING2, but treats
+  all line endings as equal."
+  [string1 string2]
+  (= (string/split-lines string1) (string/split-lines string2)))
 
 (defn code [& strings] (string/join "\n" strings))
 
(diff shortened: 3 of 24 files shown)


ericdallo:
> Note that this brings a new library in but uses just a small portion of it.

This concerns me a little bit, is that really necessary? clojure-lsp is used on other projects as a dependency so adding a new dependency may affect all downstream projects as well

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - General
+  - Fix unit tests on MS-Windows. #1211
```
ericdallo, line 6: I think we could improve this mentioning what affects the final user, like "Improve MS-Windows support fixing path issues and etc"

ericdallo:
Also, I think this will have a lot of conflict with #1231 😔 , so any suggestion of what should go first @mainej ?

CHANGELOG.md
ikappaki: Done

ikappaki:
> > Note that this brings a new library in but uses just a small portion of it.
> 
> This concerns me a little bit, is that really necessary? clojure-lsp is used on other projects as a dependency so adding a new dependency may affect all downstream projects as well

I have removed the `commons-io` dependency in the latest commit.

ikappaki:
(The macos unit test job failed with `too many open files` exception. I have rerun this in private and it has passed. The open files limit is set to 10240 as per `ulimit -n` command, which is high enough number. I will keep an eye on this, it will otherwise be indicative of an issue either in the test framework or in clojure-lsp itself. Hopefully it was just a glitch.
```
#error {
│  :cause /Users/runner/work/clojure-lsp/clojure-lsp/cli/integration-test/sample-test/deps.edn (Too many open files)
│  :via
│  [{:type java.io.FileNotFoundException
│    :message /Users/runner/work/clojure-lsp/clojure-lsp/cli/integration-test/sample-test/deps.edn (Too many open files)
│    :at [java.io.FileInputStream open0 FileInputStream.java -2]}]
│  :trace
│  [[java.io.FileInputStream open0 FileInputStream.java -2]
│   [java.io.FileInputStream open FileInputStream.java 195]
│   [java.io.FileInputStream <init> FileInputStream.java 138]
│   [clojure.java.io$fn__11617 invokeStatic io.clj 229]
│   [clojure.java.io$fn__11617 invoke io.clj 229]
│   [clojure.java.io$fn__11569$G__11523__11576 invoke io.clj 69]
│   [clojure.java.io$input_stream invokeStatic io.clj 136]
│   [clojure.java.io$input_stream doInvoke io.clj 121]
│   [clojure.lang.RestFn invoke RestFn.java 410]
│   [clojure_lsp.classpath$md5 invokeStatic classpath.clj 21]
│   [clojure_lsp.classpath$md5 invoke classpath.clj 19]
```
)

mainej:
> Also, I think this will have a lot of conflict with #1231 😔 , so any suggestion of what should go first @mainej ?

This MS-Windows fix should go first. Go ahead and merge it whenever it's ready. You're right @ericdallo, it'll cause conflicts, but #1231 is actually moving in the same direction as this PR—using `h/file-path` and `h/file-uri` in more places—so the conflicts shouldn't be too hard to resolve. And having Windows in the CI build will make me more confident about #1231.

@ikappaki it's cool to see unit tests passing for Windows. Is it safe to say that wherever the unit tests have filenames or uris, we should be wrapping those in `h/file-path` or `h/file-uri`?

ikappaki:
Hi @mainej 

> @ikappaki it's cool to see unit tests passing for Windows. Is it safe to say that wherever the unit tests have filenames or uris, we should be wrapping those in `h/file-path` or `h/file-uri`?

yes please. The helper fns to be aware of are
- `h/file-path` and `h/file-uri` for wrapping all paths and URIs.
- [NEW fn] `h/string=` to compare strings in tests with embedded `\n` line endings (windows uses `\r\n`).
- [UPDATED fn] `h/assert-submaps` with the first argument being a `#{...}` instead of a ordered collection, to compare results without taking their order into account. This is helpful when results are vector created from hash maps whose keys are filenames; the resulting order is not guaranteed to be the same across archs.

Happy to help in case you encountered window's CI failing without any apparent reason.

Thanks!

ericdallo (approved):
Thank you very much! I'm really glad there is someone with Windows knowledge helping this project :)

ericdallo:
FYI: I just made required the new windows and mac unit-tests

PR #1238, opened 2022-09-11
Fix troubleshooting.md display issue

ikappaki:
Fixes display issue, hopefully it is such a simple change in the docs and obvious fix that doesn't need an open issue to commit?

screenshot:
image


ericdallo:
thanks!

PR #1244, opened 2022-09-13
Fix lint

mainej:
Fixes lint that was introduced with some of the latest fixes for Windows.

@ericdallo it seems that CI doesn't lint all the files. Should it?

@ikappaki was your editor showing you lint on these files as you were working on them? If not, that's probably a bug in how clojure-lsp or clj-kondo works on Windows.

### DIFF
--- a/scripts/make.clj
+++ b/scripts/make.clj
@@ -1,4 +1,5 @@
 (ns make
+  (:refer-clojure :exclude [test])
   (:require [babashka.fs :as fs]
             [babashka.deps :as deps]
             [babashka.process :as p]))
@@ -7,6 +8,7 @@
                "clojure-lsp.exe"
                "clojure-lsp"))
 
+#_{:clj-kondo/ignore [:clojure-lsp/unused-public-var]}
 (defn clean []
   (let [files (into ["cli/target"
                      (fs/path "cli" lsp-bin)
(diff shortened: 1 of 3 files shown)


ericdallo (approved):
I guess lint is not liniting test folders 🤔 pretty weird, I can take a look later

ikappaki:
Hi @mainej 

Good catches!

> Fixes lint that was introduced with some of the latest fixes for Windows.
> 
> @ericdallo it seems that CI doesn't lint all the files. Should it?

@ericdallo the issue appears to be with the CWD where the diagnostics are executed from; I get different results when running the diagnostics from the `cli/` with a project root of `../` (what the ci action does) vs when running the same command from the top of the clojure-lsp project with a `--project-root` of `.`. It appears as if somehow the CWD interferes with the project root directory passed in as input.

The actual command ran as in `scripts/make.clj`
``` clojure
(defn lint-diagnostics []
  (-> (deps/clojure ["-M:run" "diagnostics" "--dry" "--ns-exclude-regex" "sample-test.*" "--project-root" "../"]
                    {:dir "cli" :inherit true})
      (p/check)))
```



> @ikappaki was your editor showing you lint on these files as you were working on them? If not, that's probably a bug in how clojure-lsp or clj-kondo works on Windows.

Yes, I remember that the editor did show me the warnings about the unused vars/namespaces, but i left it for the CI linter to remind me of them and then I forgot about them.

I think the unused functions warnings in `scripts/make.clj` are false positives; they are referenced by the top `bb.edn` project file and should not show up as unused fns. I believe this is something for `clojure-lsp` @ericdallo to consider? I've tried to lint a simple test project as such with the `clj-kondo` cli tool, and it didn't give me any warnings about them (but then again I've never used `clj-kondo` directly before, thus my conclusion might be wrong).

Changes look good. Though I am bit sceptical about silencing the warnings in `scripts/make.clj`, since they are false posiitives? I approve the changes anyway, we can always remove them later.

Thanks

scripts/make.clj
```diff
@@ -7,6 +8,7 @@
                "clojure-lsp.exe"
                "clojure-lsp"))
 
+#_{:clj-kondo/ignore [:clojure-lsp/unused-public-var]}
```
ikappaki, line 11: Note: The unused public fn warnings are all false positives since they are referenced from the top level `bb.edn` file.

ikappaki (approved):

ikappaki:
Raised #1245 (missed diagnostics when running from non root-project root dir) and #1246 (misreporting of unused var in `bb.edn` tasks)

ericdallo:
> I think the unused functions warnings in scripts/make.clj are false positives (https://github.com/clojure-lsp/clojure-lsp/issues/1246); they are referenced by the top bb.edn project file and should not show up as unused fns. I believe this is something for clojure-lsp @ericdallo to consider? I've tried to lint a simple test project as such with the clj-kondo cli tool, and it didn't give me any warnings about them (but then again I've never used clj-kondo directly before, thus my conclusion might be wrong).

This is a known issue, the unused-public-var linter is from clojure-lsp, not clj-kondo, and we don't have kondo analysis for edn files, so we don't know that those functions are really called there, something to improve in the future, but it'd need kondo changes as well

Issue #1245, opened 2022-09-14
clojure-lsp misses diagnostics when not run from within the --project-root directory


ikappaki:
**Describe the bug**
`clojure-lsp` misses diagnostics when running with `diagnostics --project-root` from a non project root directory

**To Reproduce**

create a test directory with the following files (say in `~/clsp-diags-root`) :

`bb.edn`:
``` clojure
{:paths ["scripts"]
 :tasks
 {hi make/hi}}
```
`scripts/make.clj`
``` clojure
(ns make)
(defn hi []
  (println :hi))
```

1. cd to project root and run `clojure-lsp diagnostics --project-root ~/clsp-diags-root --verbose`:
```
~/clsp-diags-root $ clojure-lsp diagnostics --project-root ~/clsp-diags-root --verbose
[ 10%] Finding cache               [INFO] [DB] Reading transit analysis cache from /home/ikappaki/clsp-diags-root/.lsp/.cache/db.transit.json db took 130ms
[INFO] :maintain-dep-graph 75ms
[INFO] [Startup] Using cached db for project root /home/ikappaki/clsp-diags-root
[INFO] [Startup] Using source-paths from classpath: ["/home/ikappaki/clsp-diags-root/scripts"]
[ 15%] Copying kondo configs       [INFO] Copying kondo configs from classpath to project if any...
Configs copied:
- .clj-kondo/babashka/fs
- .clj-kondo/rewrite-clj/rewrite-clj
[INFO] Copied kondo configs, took 43ms secs.
[ 20%] Analyzing project files     [INFO] [Startup] Analyzing source paths for project root /home/ikappaki/clsp-diags-root
[INFO] [Startup] Project only paths analyzed by clj-depend, took 0ms
[ 99%] Analyzing project files     No configs copied.
[INFO] :maintain-dep-graph 0ms
[INFO] Linting whole project for unused-public-var took 1ms
[INFO] [Startup] Project only paths analyzed by clj-kondo, took 9ms
[INFO] :maintain-dep-graph 0ms
[100%] Project analyzed            
Finding diagnostics...
scripts/make.clj:1:6: info: [clojure-lsp/unused-public-var] Unused public var 'make/hi'
```
2. cd to project root's `script/` and run the same command   `clojure-lsp diagnostics --project-root ~/clsp-diags-root --verbose`:
``` 
~/clsp-diags-root/scripts $ clojure-lsp diagnostics --project-root ~/clsp-diags-root --verbose
[...]

PR #1249, opened 2022-09-15
Wrap files in h/file-path/uri, follow up fix to #1245 breaks

ikappaki:


ericdallo (approved):
Thanks, I was about to take a look at those tests

PR #1251, opened 2022-09-16
Fixes NPE when checking code actions of file that begins with comment

mainej:
This fixes an NPE that is thrown when a client requests code actions but a zloc can't be found.

One scenario where a zloc can't be found is when a source file begins with a comment or whitespace. (Calva automatically invokes code actions as you navigate, so this can easily be reproduced by adding a comment at the top of a file and clicking on it.)

### DIFF
--- a/lib/src/clojure_lsp/feature/sort_clauses.clj
+++ b/lib/src/clojure_lsp/feature/sort_clauses.clj
@@ -6,13 +6,14 @@
    [rewrite-clj.zip :as z]))
 
 (defn ^:private clause-spec [zloc uri db]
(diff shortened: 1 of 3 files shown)


mainej:
Also see #1252 for a longer-term fix for this kind of problem.

ericdallo:
@mainej I didn't see this PR and [did a commit ](https://github.com/clojure-lsp/clojure-lsp/commit/7ac822d3efe3dd51db4fef8959b8708ad7fb2e16) to fix a issue reported by a Calva user on a thread, even so it seems a separated problem you are solving here, will take a look soon, just letting you know

ericdallo (approved):

PR #1255, opened 2022-09-19
Consider original text's line separator type when reporting changes

ikappaki:
Hi,

could you please consider patch to fix linter issues with line endings when reporting changes back to the client. This is part of #1211.

`clojure-lsp` normalizes all line endings to those of *unix, i.e. `\n` (aka LF), either directly or some times indirectly via third party libraries (see https://github.com/clj-commons/rewrite-clj/issues/187). On MS-Windows the native line endings are `\r\n` (aka CRLF), and thus if we internally treat all line endings as `\n`, we might end up reporting text as changing, although the only thing that changed was the internal change of `\r\n` to `\n`, i.e. false positive changes. This is what happened when running the `bb lint-clean` linter on windows, although there was no namespace to clean, differences were reported.

The patch tries to determine the line ending type of the original text and forces that to the replacement text, thus maintaining the original line endings.

It also:
- fixes java interop test failure on win introduced with earlier unrelated commit.
- introduces a convenient `bb lint` task.

thanks


ericdallo (approved):
Looks good, thank you!

PR #1259, opened 2022-09-20
Process requests in parallel

### LINKED ISSUE #1240: Small freezes regression during editing on big buffers
**Describe the bug**

@borkdude noticed small freezes on emacs while coding on squint-cljs/squint, I can confirm the issue, it seems to be a regression [since 2022.06.22-14.09.50 tag ](https://github.com/clojure-lsp/clojure-lsp/compare/2022.06.22-14.09.50...2022.09.01-15.27.31).

After some debugging, I noticed it's related to [process-after-all-changes](https://github.com/clojure-lsp/clojure-lsp/blob/master/lib/src/clojure_lsp/handlers.clj#L45), the issue is gone only after commenting all usages of that macro.

@mainej mentioned that it may be related with lsp4clj migration, but I suspect it could be related to the change on that macro where it supports waiting for multiple URIs, not only one.

**To Reproduce**
- Clone [squint-cljs/quint](https://github.com/squint-cljs/squint)
- Open compiler.cljc
- add a new line after line 491, and start writing `x`s slowly to repro the slowness/freezing

**Expected behavior**
Do not block editor or avoid freezing

mainej:
Uses https://github.com/clojure-lsp/lsp4clj/pull/25 to process request in parallel, fixing typing lag and other performance problems documented in #1240.

This PR is currently incomplete. At the moment the lsp4clj dependency points to a local repo. It needs https://github.com/clojure-lsp/lsp4clj/pull/25 to be merged and released, and then needs to use that version in the deps. I'm submitting it now though @ericdallo, so that https://github.com/clojure-lsp/lsp4clj/pull/25 can be tested more easily.

Fixes #1240.

### DIFF
--- a/.clj-kondo/funcool/promesa/config.edn
+++ b/.clj-kondo/funcool/promesa/config.edn
@@ -1,0 +1,8 @@
+{:lint-as {promesa.core/->          clojure.core/->
+           promesa.core/->>         clojure.core/->>
+           promesa.core/as->        clojure.core/as->
+           promesa.core/let         clojure.core/let
+           promesa.core/plet        clojure.core/let
+           promesa.core/loop        clojure.core/loop
+           promesa.core/recur       clojure.core/recur
+           promesa.core/with-redefs clojure.core/with-redefs}}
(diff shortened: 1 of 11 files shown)


ericdallo:
@mainej I made the version changes, and made the merge as it seemed ready to me, I also tested manually and saw no issues until now, LMK if it needs any follow-up PR

PR #1261, opened 2022-09-20
Simplify finding of class defined by defrecord

mainej:
This re-uses the code to find a var-definition defined by defrecord, when the usage was part of an `:import`.


ericdallo (approved):
Good catch!

PR #1276, opened 2022-09-25
Skip cancelled futures

mainej:
The body of a promise created by `p/future` is always run, even if the promise is cancelled (by lsp4clj) with `p/cancel!`. For this reason, clojure-lsp was doing more work on cancelled futures than it really needed to.

This switches back to `CompletableFuture/supplyAsync` + `Supplier`, which can actually be cancelled. This is how the code worked before the migration away from lsp4j.

To be more concrete, the following will always return `[1, 2, ... max-n]`.

```clojure
(defn check [make-future]
  (let [results (atom [])
        common-parallelism (java.util.concurrent.ForkJoinPool/getCommonPoolParallelism)
        max-n (* 2 common-parallelism)
        process-result (fn [n]
                         (Thread/sleep (* 100 n))
                         (swap! results conj n))
        futs (doall
               (map (fn [n]
                      (make-future #(process-result n)))
                    (range 1 (inc max-n))))]
    ;; give executor time to start first `common-parallelism` tasks
    (Thread/sleep 20)
    ;; cancel all futures, before some have started, and before any are finished
    (run! p/cancel! futs)
    ;; wait until all futures _would_ have finished, if they had been run
    (Thread/sleep (* 200 max-n))
    ;; check which futures actually ran
    @results))

(check (fn [f] (p/future (f))))
```

This is in contrast to the following, which returns a vector containing fewer results, usually `[1, 2, ... common-parallelism]`, but never more than `common-parallelism` of the `n`s.

```clojure
(check (fn [f]
         (java.util.concurrent.CompletableFuture/supplyAsync
           (reify java.util.function.Supplier
             (get [_] (f))))))
```

That is, if a `supplyAsync` future is cancelled before the executor starts it, then the body of `Supplier/get` is never run, whereas the body of a `p/future` is always run, even if it is cancelled.

This is an amendment to #1259 and may improve lingering problems from #1240. 

### DIFF
--- a/cli/src/clojure_lsp/server.clj
+++ b/cli/src/clojure_lsp/server.clj
@@ -18,8 +18,13 @@
    [lsp4clj.liveness-probe :as lsp.liveness-probe]
    [lsp4clj.lsp.requests :as lsp.requests]
    [lsp4clj.server :as lsp.server]
-   [promesa.core :as p]
-   [taoensso.timbre :as timbre]))
+   [taoensso.timbre :as timbre])
+  (:import
+   (java.util.concurrent CompletableFuture)
+   (java.util.function Supplier)))
+
+(defmacro eventually [& body]
+  `(CompletableFuture/supplyAsync (reify Supplier (get [this] ~@body))))
 
 (set! *warn-on-reflection* true)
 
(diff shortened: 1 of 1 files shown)


ericdallo (approved):
Weird behavior of p/future, maybe upstream is interested on supporting that?

mainej:
> maybe upstream is interested on supporting that?

Yeah, I plan to submit this as a question to `promesa`. It may be a bug, or perhaps there's another tool to use besides `p/future`.

mainej:
https://github.com/funcool/promesa/issues/111

PR #1275, opened 2022-09-25
Use `deps-bin` on MS-Windows, add dev testing and bb tasks documentation

ikappaki:
Hi @ericdallo,

as requested some time ago, could you please review patch to remove the `launch4j` widnows dependency for generating cli script wrappers in favor of https://github.com/ericdallo/deps-bin/pull/3. This is part of #1211.

I've also 
- added a dev testing doc section explaining the diffs of *nix vs windows, and updated the doc of the bb tasks mentioned in there. Please update/correct as you see fit.
- fixed test path issues from previous unrelated commit while enforcing `babashka.fs/canonicalize` uses in helper functions.

- [x] I created an issue to discuss the problem I am trying to solve or an open issue already exists.
- [x] I updated documentation if applicable (`docs` folder)

Thanks,

PR #1277, opened 2022-09-25
Handle clojure-lsp executable according to its usage

ikappaki:
Fixes #1275 PR issue with running integration test in `release.yml` with native executable on MS-Windows

There can be now two executables with different extensions created for windows (`.bat` and `.exe`) and the logic should be able to differentiate between the two based on their usage.

btw the following is called as if `babashka.fs/windows?` was private (it is not) to get around a false clojure-lsp warning:

```clojure
(def windows? (#'fs/windows?))
```

There must be a bug somewhere in the diagnostics linter? as if the kondo config from `babashka.fs` has not been copied over ...
``` bat
d:/src/clojure-lsp $ bb lint-diagnostics
[100%] Project analyzed            
Finding diagnostics...
scripts\make.clj:8:14: error: [private-call] #'babashka.fs/windows? is private
----- Error --------------------------------------------------------------------
```

### DIFF
--- a/scripts/make.clj
+++ b/scripts/make.clj
@@ -5,19 +5,41 @@
    [babashka.fs :as fs]
    [babashka.process :as p]))
 
-(def lsp-bin (if (#'fs/windows?)
-               "clojure-lsp.bat"
-               "clojure-lsp"))
+(def windows? (#'fs/windows?))
+
+(defn lsp-bin-filename
+  "Returns the `clojure-lsp` executable filename for this system based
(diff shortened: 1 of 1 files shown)


ericdallo:
> There must be a bug somewhere in the diagnostics linter? as if the kondo config from babashka.fs has not been copied over ...

I noticed that recently, and AFAICS it only happens with that lib 🤔

ericdallo (approved):

PR #1282, opened 2022-09-29
Bump graalvm to 22.2.0 for the MS-Windows release binary

ikappaki:
Hi,

could you please review patch to upgrade Graalvm for the MS-Windows executable to 22.2.0, in sync with the other archs. This is part of #1211

This Graalvm version does not work with well with the UPX compression program, as per https://github.com/oracle/graal/issues/4340

A workaround suggested in that thread is to remove a class file from GraalVMs `svm.jar` file, which appears to work -- all integration tests pass.

Thanks

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -326,7 +326,7 @@ jobs:
   windows:
     name: Build native Windows binary
     needs: [build-lib-jar, build-cli-jar]
-    runs-on: windows-2022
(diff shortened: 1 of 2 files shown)


ericdallo (approved):
I'm so glad about that! That was something I knew and was unsure how to fix, thank you very much!

ericdallo:
FIxes #767

PR #1294, opened 2022-10-04
Rename test files and crawler

mainej:
This is two different filenames.

First, this renames `lib/test/clojure_lsp/{features => feature}/*`. This allows easier navigation and test execution in editors that understand the relationship between implementation file names and their test file names.

Second, it renames crawler.clj. We talked about this at some point @ericdallo, but I can't find the discussion. I renamed it to `clojure-lsp.startup`, since that's what the functions in the namespace support, but let me know if you'd prefer another name.

### DIFF
--- a/cli/build.clj
+++ b/cli/build.clj
@@ -1,11 +1,9 @@
 (ns build
   (:require
    [babashka.fs :as fs]
-   [babashka.process :as p]
    [clojure.java.io :as io]
    [clojure.string :as string]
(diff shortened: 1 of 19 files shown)


ericdallo (approved):
I was aware of the  test source path divergence but was avoiding doing a huge git change, it turns out it was pretty simple the gif diff :)
Nice, I thought recently about that again, startup sounds great!

PR #1314, opened 2022-10-06
Tidy make scripts

mainej:
Tidies the make scripts, and removes unused dependencies.

### DIFF
--- a/scripts/make.clj
+++ b/scripts/make.clj
@@ -22,10 +22,10 @@
 
   :script The filename of the executable script wrapper."
   [usage]
-  (let [lsp-bin "clojure-lsp"]
-    (case usage
-      :native (cond-> lsp-bin windows? (str ".exe"))
-      :script (cond-> lsp-bin windows? (str ".bat")))))
+  (cond-> "clojure-lsp"
+    windows? (str (case usage
+                    :native ".exe"
+                    :script ".bat"))))
 
 #_{:clj-kondo/ignore [:clojure-lsp/unused-public-var]}
 (defn clean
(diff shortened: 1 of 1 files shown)


bb.edn
```diff
@@ -1,13 +1,6 @@
-{:paths ["scripts"
-         "cli/integration-test"]
```
ericdallo, line 2: So, this is kind of a ugly hack to make clojure-lsp consider cli/integration-test as source-paths and lint passes, without that clojure-lsp won't consider that folder :( 
On deps.edn we have the feature of local/root to cli/deps.edn, but bb doesn't support that, WDYT, any suggestions?

mainej: How does the linter know to lint cli/src and lib/src?

ericdallo: because clojure-lsp run `clj -A:dev:test -Spath`, this checks dev and test aliases which points to local/roots of cli and lib

ericdallo: Yeah, I think we would need to rollback the deps as well :(

mainej: I see... so when we run lint we need two things. clojure-lsp/lib/src needs to be in the source paths so that it is linted, and the :deps in clojure-lsp/lib/deps.edn need to be on the classpath so that the analysis works. To accomplish this we have:

1. `clojure-lsp/deps.edn -> :dev alias -> :extra-paths ["lib/src" ...]` which puts clojure-lsp/lib/src in the source paths.
2. `clojure-lsp/deps.edn -> :dev alias -> :extra-deps clojure-lsp/lib -> :local/root "lib" -> clojure-lsp/lib/deps.edn -> :paths ["src"] + :deps ...`, which puts the lib deps and its src directory on the classpath.

We have something similar for the cli directory and its deps.

For the integration tests we could achieve the first goal, getting the integration test directory in the source paths, by adding `clojure-lsp/deps.edn -> :integration-test alias -> :extra-paths ["cli/integration-test"]`. But we'd still have a problem with the second goal, getting the integration test deps on the classpath. To do that we'd need those deps in a deps.edn file which can act as a local/root. But those deps are specified in clojure-lsp/cli/bb.edn, which can't be a local/root. And we can't turn things around either. If we moved those deps to a deps.edn file then the babashka integration test runner couldn't see them.

I don't think I every really appreciated how clojure-lsp/deps.edn isn't really doing anything except for instructing clojure-lsp (the tool) on how to interpret clojure-lsp (the project and its file structure).

(This brings up several other questions... Why does the :dev alias in clojure-lsp/deps.edn include tools.build? We never build from the root directory. Why does the :test alias include kaocha, jvm-opts and main-opts? We never run tests from the root directory.)

So, I think I see now why we need to duplicate the `:paths` and `:deps` from cloure-lsp/cli/bb.edn to a deps.edn where they'll get picked up for the root source paths and classpath.

Personally, I'd rather have all those instructions to clojure-lsp the tool in the same place, so I'd suggest we move them to an `:integration-test` alias in clojure-lsp/deps.edn (and add that to :source-aliases in .lsp/config.edn) rather than have them in clojure-lsp/bb.edn. In my mind, clojure-lsp/bb.edn exists to run make tasks, not to run integration tests, so it's confusing to have that information there, even though clojure-lsp the tool doesn't mind. I'd also like to document that the aliases in clojure-lsp/deps.edn are there for the benefit of clojure-lsp the tool (assuming that's their sole purpose). WDYT?

mainej: Oh wait... I think I can answer a few of my own questions.

> Why does the :dev alias in clojure-lsp/deps.edn include tools.build? ... Why does the :test alias include kaocha?

The reason is that the local/root trick has a limitation. You can use it to include the `:paths` and `:deps` from a subtree, but not the `:extra-paths` or `:extra-deps` from an _alias_ in that subtree. If you want clojure-lsp to analyze those paths and deps too, you have to duplicate them into the parent tree's deps.edn. 👎 

Our root deps.edn duplicates tools.build and kaocha because they're deps of nested aliases.

I think I'm starting to appreciate why we get requests to support multiple roots. I know we sometimes say "clojure-lsp has several source directories, and we get by with `:source-aliases` and local/roots", but I'm not very happy with that answer right now.

ericdallo: yeah, I'm not 100% happy with that as well, but it's something I started to understand after studying [polylith](https://polylith.gitbook.io/polylith/), they use the same approach of having a dev/test alias that link all submodules together since when developing in monorepos you usually want to touch functions and see how they affect other modules.
I liked your idea of having a integration-test alias, that could help, but it´d be a little bit weird to have it there since integration-tests run on babashka, not on `clojure`.

ericdallo: I feel like we should start a issue or discussion about that and it'd be nice to see a outcome of a proper clojure-lsp docs explaining how to support mono-repos

bb.edn
```diff
@@ -1,4 +1,5 @@
 {:paths ["scripts"
+         ;; ensure lint checks integration tests:
```
ericdallo, line 2: Not only liniting, but clojure-lsp as a whole, like when checking unused-public-var which will only work properly if those paths are on classpath for clojure-lsp during initialize

mainej: I've reverted all the changes to `bb.edn`, so this is ready for another review.

bb.edn
mainej: > I feel like we should start a issue or discussion about that and it'd be nice to see a outcome of a proper clojure-lsp docs explaining how to support mono-reposSounds good. I'll do that when I have a chance.

ericdallo (approved):
Nice

PR #1322, opened 2022-10-12
Avoid blocked requests, and abort cancelled requests

mainej:
This makes two changes related to more efficiently handling process-after-changes requests.

First, as we noted while debugging borkdude's clojure-lsp setup, on machines with fewer cores the process-after-changes requests can fill up the thread pool, blocking requests that would find room in the pool on machines with more cores. This is despite the fact that the process-after-changes requests aren't doing anything except sleeping. This PR makes clojure-lsp use two pools, one for the process-after-changes requests, and one for the other requests. That way they don't interfere with each other. If we notice any problems with memory usage from having additional threads, this is easy to turn off... we can simply revert to using the same pool.

Second, this PR changes the process-after-changes requests to abort if they get a signal from lsp4clj that the client has aborted them. This keeps the process-after-changes pool working on requests that the client actually cares about. It also reduces CPU usage by avoiding unnecessary work.

Note that these changes aren't about mitigating performance problems introduced by the migration away from lsp4j. Both of them (or variations on them) would have benefited clojure-lsp even before the migration. 

Finally this PR removes a global lock from `range-formatting`. See the associated commit message for details.

- [x] I created an issue to discuss the problem I am trying to solve or an open issue already exists. Part of this is the clojure-lsp side of https://github.com/clojure-lsp/lsp4clj/pull/30

### DIFF
--- a/lib/src/clojure_lsp/handlers.clj
+++ b/lib/src/clojure_lsp/handlers.clj
@@ -47,15 +47,21 @@
        (take 20))
   #_{})
 
-(defn wait-for-all-changes [uris db*]
+(defn ^:private wait-for-changes [{:keys [db* :lsp4clj.server/req-cancelled?]} uris]
   (let [delay-start (System/nanoTime)]
     (loop [immediate? true
            backoff backoff-start
            uris uris]
       (let [now (System/nanoTime)]
-        (if (> (quot (- now delay-start) 1000000) 60000) ; one minute timeout
+        (cond
+          (some-> req-cancelled? deref)
+          {:delay/outcome :cancelled
+           :delay/start delay-start
+           :delay/end now}
+          (> (quot (- now delay-start) 1000000) 60000) ; one minute timeout
           {:delay/outcome :timed-out
            :delay/timeout-uris uris}
+          :else
           (if-let [processing-uris (seq (filter (:processing-changes @db*) uris))]
             (do
               (Thread/sleep backoff)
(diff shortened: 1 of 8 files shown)


lib/src/clojure_lsp/handlers.clj
```diff
@@ -67,31 +68,48 @@
...
            result#)))))
 
+(defmacro process-after-all-changes
+  [components uris task & body]
+  (with-meta
+    `(logging-delayed-task
+       (wait-for-changes ~components ~uris)
+       ~task
+       ~@body)
+    (meta &form)))
```
ericdallo, line 105: I was about to suggest moving both calls to a single function in the previous PR, glad you noticed that as well

ericdallo (approved):
Changes look good, could you take a look at the CI @mainej ?

lib/src/clojure_lsp/handlers.clj
mainej: 👍 That PR also removed the name `process-after-changes`, and this restores it. I'm glad to have it back—it's a useful shorthand for this part of the code.

mainej:
CI is fixed so this is ready @ericdallo. Apologies—I got into a weird merge situation and ended up force-pushing a rebase that removed your merge of master. If you have a local copy of this branch, it'll be broken.

ericdallo:
I tested manually, briefly, but looks everything ok

ferdinand-beyer:
We hit a similar issue at Scarlet, based on code from this PR.

My guess is that the original problem this PR aims to fix is that we used `p/future` for waiting tasks. This will submit them to the `ForkJoinPool/commonPool`, which uses a (somewhat) fixed number of threads and is actually not designed for blocking tasks. This pool can fill up and stop accepting new tasks.

Delegating some tasks to a custom `forkjoin-executor` shifts this problem, but this again is not designed for blocking tasks.

Instead, we should use a cached thread pool, as we do now with the `p/thread` macro. I think we could remove the `changes-executor` again and submit all task to promesa's `:thread` executor. If we want to keep it, we should change it from `forkjoin-executor` to `cached-executor`.

PR #1327, opened 2022-10-15
Make encode work for zipfile deps

### LINKED ISSUE #1287: Regression: analysis within zipfile
**Describe the bug**
analysis within jar files seems to have stopped. jump to definition and hover provide no information once inside a jar.

**To Reproduce**
in neovim, 
jump to a definition in a jar dependency, 
then try to jump to a second definition

**Expected behavior**
Expect to be able to continue navigating within jar files.

[Discussion in clojurians](https://clojurians.slack.com/archives/CPABC1H61/p1664472587981779) has some investigation of current commits. Seems to have broken somewhere between a4a44f55 and master

server-logs starting at a jump into schema dependency.
```
2022-09-30T17:42:53.391Z  INFO [clojure-lsp.handlers:183] - :definition 0ms
2022-09-30T17:42:54.427Z  INFO [clojure-lsp.dep-graph:271] - :maintain-dep-graph 0ms
2022-09-30T17:42:54.438Z  WARN [clojure-lsp.kondo:299] - Non-fatal error from clj-kondo: No configs copied.

2022-09-30T17:42:54.444Z  INFO [clojure-lsp.dep-graph:271] - :maintain-dep-graph 0ms
2022-09-30T17:42:54.445Z  INFO [clojure-lsp.handlers:120] - :did-open 936ms
2022-09-30T17:42:54.445Z  INFO [clojure-lsp.handlers:213] - :document-highlight 0ms
2022-09-30T17:42:54.550Z  DEBUG [clojure-lsp.server:92] - Publishing 0 diagnostics for zipfile:/Users/snoe/.m2/repository/prismatic/schema/1.2.1/schema-1.2.1.jar%3A%3Aschema/core.cljc
2022-09-30T17:42:54.550Z  INFO [clojure-lsp.server:93] - :publish-diagnostics 0ms
2022-09-30T17:43:04.407Z  INFO [clojure-lsp.handlers:213] - :document-highlight 0ms
2022-09-30T17:43:08.742Z  INFO [clojure-lsp.handlers:183] - :definition 0ms
2022-09-30T17:43:09.466Z  INFO [clojure-lsp.handlers:213] - :document-highlight 0ms
2022-09-30T17:43:30.917Z  INFO [clojure-lsp.handlers:213] - :document-highlight 0ms
2022-09-30T17:43:31.354Z  INFO [clojure-lsp.handlers:333] - :hover 17ms
2022-09-30T17:43:31.694Z  INFO [clojure-lsp.handlers:213] - :document-highlight 0ms
2022-09-30T17:43:38.143Z  INFO [clojure-lsp.handlers:213] - :document-highlight 0ms
2022-09-30T17:43:41.934Z  INFO [clojure-lsp.handlers:333] - :hover 17ms
2022-09-30T17:43:43.804Z  INFO [clojure-lsp.handlers:213] - :document-highlight 0ms
2022-09-30T17:43:43.919Z  INFO [clojure-lsp.handlers:333] - :hover 15ms
2022-09-30T17:43:44.258Z  INFO [clojure-lsp.handlers:213] - :document-highlight 0ms
2022-09-30T17:43:44.535Z  INFO [clojure-lsp.handlers:183] - :definition 0ms
2022-09-30T17:43:45.195Z  INFO [clojure-lsp.handlers:213] - :document-highlight 0ms
```

client logs
```
2022-09-30T11:42:53.370 INFO (pid:77631) [attach] - receive notification: jumpDefinition []
2022-09-30T11:42:53.993 INFO (pid:77631) [attach] - receive notification: highlight []
2022-09-30T11:43:04.386 INFO (pid:77631) [attach] - receive notification: highlight []
2022-09-30T11:43:06.690 INFO (pid:77631) [attach] - receive notification: highlight []
2022-09-30T11:43:08.717 INFO (pid:77631) [attach] - receive notification: jumpDefinition []
2022-09-30T11:43:09.381 INFO (pid:77631) [attach] - receive notification: highlight []
2022-09-30T11:43:30.891 INFO (pid:77631) [attach] - receive notification: highlight []
2022-09-30T11:43:31.314 INFO (pid:77631) [attach] - Request action: doHover []
2022-09-30T11:43:31.670 INFO (pid:77631) [attach] - receive notification: highlight []
2022-09-30T11:43:38.128 INFO (pid:77631) [attach] - receive notification: highlight []
2022-09-30T11:43:41.039 INFO (pid:77631) [attach] - receive notification: highlight []
2022-09-30T11:43:41.894 INFO (pid:77631) [attach] - Request action: doHover []
2022-09-30T11:43:42.250 INFO (pid:77631) [attach] - receive notification: highlight []
2022-09-30T11:43:43.773 INFO (pid:77631) [attach] - receive notification: highlight []
2022-09-30T11:43:43.840 INFO (pid:77631) [attach] - Request action: doHover []
2022-09-30T11:43:44.231 INFO (pid:77631) [attach] - receive notification: highlight []
2022-09-30T11:43:44.513 INFO (pid:77631) [attach] - receive notification: jumpDefinition []
2022-09-30T11:43:45.171 INFO (pid:77631) [attach] - receive notification: highlight []
2022-09-30T11:44:56.117 INFO (pid:77631) [attach] - receive notification: highlight []
2022-09-30T11:44:57.111 INFO (pid:77631) [attach] - receive notification: highlight []
2022-09-30T11:45:01.128 INFO (pid:77631) [attach] - receive notification: openLog []

```

ericdallo:
Fixes #1287

### DIFF
--- a/cli/integration-test/integration/definition_test.clj
+++ b/cli/integration-test/integration/definition_test.clj
@@ -108,23 +108,23 @@
 
 (deftest definition-external-dependency-zipfile-scheme-escaped-uris
   (h/delete-project-file "../../.lsp/.cache/")
-  (h/set-escape-uris! true)
-  (lsp/start-process!)
-  (lsp/request! (fixture/initialize-request))
-  (lsp/notify! (fixture/initialized-notification))
-  (lsp/notify! (fixture/did-open-source-path-notification "definition/a.clj"))
+  (with-redefs [h/*escape-uris?* true]
+    (lsp/start-process!)
+    (lsp/request! (fixture/initialize-request))
+    (lsp/notify! (fixture/initialized-notification))
+    (lsp/notify! (fixture/did-open-source-path-notification "definition/a.clj"))
 
-  (let [{:keys [uri]} (lsp/request! (fixture/definition-request "definition/a.clj" 15 2))]
-    (lsp/notify! (fixture/did-open-external-path-notification uri (-> (h/unescape-uri uri)
-                                                                      (string/replace "zipfile:" "jar:file:")
-                                                                      (string/replace "::" "!/")
-                                                                      slurp)))
+    (let [{:keys [uri]} (lsp/request! (fixture/definition-request "definition/a.clj" 15 2))]
+      (lsp/notify! (fixture/did-open-external-path-notification (h/escape-uri uri)
+                                                                (-> (h/unescape-uri uri)
+                                                                    (string/replace "zipfile:" "jar:file:")
+                                                                    (string/replace "::" "!/")
+                                                                    slurp)))
 
-    (testing "LSP features work on external clojure opened files"
-      (h/assert-submap
-        {:language "clojure"
-         :value "[x]\n[x message]"}
-        (-> (lsp/request! (fixture/hover-external-uri-request uri 7612 5))
-            :contents
-            (get 1)))))
-  (h/set-escape-uris! false))
+      (testing "LSP features work on external clojure opened files"
+        (h/assert-submap
+          {:language "clojure"
+           :value "[x]\n[x message]"}
+          (-> (lsp/request! (fixture/hover-external-uri-request (h/escape-uri uri) 7612 5))
+              :contents
+              (get 1)))))))
--- a/cli/integration-test/integration/definition_test.clj
+++ b/cli/integration-test/integration/definition_test.clj
@@ -105,3 +105,26 @@
         (-> (lsp/request! (fixture/hover-external-uri-request uri 7612 5))
             :contents
             (get 1))))))
+
+(deftest definition-external-dependency-zipfile-scheme-escaped-uris
+  (h/delete-project-file "../../.lsp/.cache/")
+  (h/set-escape-uris! true)
+  (lsp/start-process!)
+  (lsp/request! (fixture/initialize-request))
+  (lsp/notify! (fixture/initialized-notification))
+  (lsp/notify! (fixture/did-open-source-path-notification "definition/a.clj"))
+
+  (let [{:keys [uri]} (lsp/request! (fixture/definition-request "definition/a.clj" 15 2))]
+    (lsp/notify! (fixture/did-open-external-path-notification uri (-> (h/unescape-uri uri)
+                                                                      (string/replace "zipfile:" "jar:file:")
+                                                                      (string/replace "::" "!/")
+                                                                      slurp)))
+
+    (testing "LSP features work on external clojure opened files"
+      (h/assert-submap
+        {:language "clojure"
+         :value "[x]\n[x message]"}
+        (-> (lsp/request! (fixture/hover-external-uri-request uri 7612 5))
+            :contents
+            (get 1)))))
+  (h/set-escape-uris! false))
--- a/cli/integration-test/integration/definition_test.clj
+++ b/cli/integration-test/integration/definition_test.clj
@@ -121,10 +121,16 @@
                                                                     (string/replace "::" "!/")
                                                                     slurp)))
 
-      (testing "LSP features work on external clojure opened files"
+      (testing "LSP hover feature works on external clojure opened files"
         (h/assert-submap
           {:language "clojure"
            :value "[x]\n[x message]"}
           (-> (lsp/request! (fixture/hover-external-uri-request (h/escape-uri uri) 7612 5))
               :contents
-              (get 1)))))))
+              (get 1))))
+
+      (testing "LSP definition from external clojure opened file works"
+        (h/assert-submap
+          {:language "clojure"
+           :value "[x]\n[x message]"}
+          (lsp/request! (fixture/definition-request (h/escape-uri uri) 7612 5)))))))
--- a/cli/integration-test/integration/definition_test.clj
+++ b/cli/integration-test/integration/definition_test.clj
@@ -21,21 +21,21 @@
         {:uri (h/source-path->uri "definition/a.clj")
          :range {:start {:line 8 :character 16}
                  :end {:line 8 :character 33}}}
-        (lsp/request! (fixture/definition-request "definition/a.clj" 11 2))))
+        (lsp/request! (fixture/definition-request (h/source-path->uri "definition/a.clj") 11 2))))
 
     (testing "find definition of local var"
       (h/assert-submap
         {:uri (h/source-path->uri "definition/a.clj")
          :range {:start {:line 3 :character 5}
                  :end {:line 3 :character 13}}}
-        (lsp/request! (fixture/definition-request "definition/a.clj" 9 9))))
+        (lsp/request! (fixture/definition-request (h/source-path->uri "definition/a.clj") 9 9))))
 
     (testing "find definition of public var"
       (h/assert-submap
         {:uri (h/source-path->uri "definition/a.clj")
          :range {:start {:line 5 :character 6}
                  :end {:line 5 :character 22}}}
-        (lsp/request! (fixture/definition-request "definition/b.clj" 3 4)))))
+        (lsp/request! (fixture/definition-request (h/source-path->uri "definition/b.clj") 3 4)))))
 
   (testing "keywords"
     (testing "definition of local simple keyword is the keyword itself"
@@ -43,28 +43,28 @@
         {:uri (h/source-path->uri "definition/b.clj")
          :range {:start {:line 5 :character 0}
                  :end {:line 5 :character 17}}}
-        (lsp/request! (fixture/definition-request "definition/b.clj" 5 2))))
+        (lsp/request! (fixture/definition-request (h/source-path->uri "definition/b.clj") 5 2))))
 
     (testing "definition of invalid local namespaced keyword is the keyword itself"
       (h/assert-submap
         {:uri (h/source-path->uri "definition/b.clj")
          :range {:start {:line 7 :character 0}
                  :end {:line 7 :character 8}}}
-        (lsp/request! (fixture/definition-request "definition/b.clj" 7 2))))
+        (lsp/request! (fixture/definition-request (h/source-path->uri "definition/b.clj") 7 2))))
 
     (testing "find definition of valid other ns namespaced keyword"
       (h/assert-submap
         {:uri (h/source-path->uri "definition/a.clj")
          :range {:start {:line 13 :character 7}
                  :end {:line 13 :character 15}}}
-        (lsp/request! (fixture/definition-request "definition/b.clj" 9 2))))
+        (lsp/request! (fixture/definition-request (h/source-path->uri "definition/b.clj") 9 2))))
 
     (testing "find definition of valid other aliased keyword"
       (h/assert-submap
         {:uri (h/source-path->uri "definition/a.clj")
          :range {:start {:line 13 :character 7}
                  :end {:line 13 :character 15}}}
-        (lsp/request! (fixture/definition-request "definition/b.clj" 11 2))))))
+        (lsp/request! (fixture/definition-request (h/source-path->uri "definition/b.clj") 11 2))))))
 
 (deftest definition-external-dependency-jar-scheme
   (h/delete-project-file "../../.lsp/.cache/")
@@ -74,7 +74,7 @@
   (lsp/notify! (fixture/initialized-notification))
   (lsp/notify! (fixture/did-open-source-path-notification "definition/a.clj"))
 
-  (let [{:keys [uri]} (lsp/request! (fixture/definition-request "definition/a.clj" 15 2))]
+  (let [{:keys [uri]} (lsp/request! (fixture/definition-request (h/source-path->uri "definition/a.clj") 15 2))]
     (lsp/notify! (fixture/did-open-external-path-notification uri (slurp uri)))
 
     (testing "LSP features work on external clojure opened files"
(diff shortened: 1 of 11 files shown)


cli/integration-test/integration/definition_test.clj
```diff
@@ -105,3 +105,26 @@
         (-> (lsp/request! (fixture/hover-external-uri-request uri 7612 5))
             :contents
             (get 1))))))
+
+(deftest definition-external-dependency-zipfile-scheme-escaped-uris
+  (h/delete-project-file "../../.lsp/.cache/")
+  (h/set-escape-uris! true)
```
ericdallo, line 111: I thought about passing a param but would need lots of changes saying that the URI should be encoded, so a dynamic var was less worst 😅

mainej:
Here's my hypothesis about how clojure-lsp works (before this patch) and why the patch fixes things. @ericdallo, do you agree with this thought process?

At startup we give clj-kondo the classpath, but not specific file names. It analyzes the paths, returning analysis with unescaped regular file and jar file names. We convert those file names to URIs, leaving them unescaped. The project internal files are all fine because their URIs are the same whether escaped or not. But the DB's references to external files use unescaped URIs even though the client plans to use escaped URIs. (This is arguably the first part of the bug, but the dep-graph saves us from some problems, as I'll explain in a moment.)

So suppose you're in an internal file, on a usage of an external var and you want to go to its definition. We use the dep graph to go from the var's namespace to the (unescaped) jar file URI. We tell the editor to open that URI, which it does. It sends a didOpen for that external file, giving us an escaped URI.

We then do a little dance with clj-kondo which wants filenames not URIs. We unescape the URI (in shared/uri->filename) and pass that filename over to clj-kondo. clj-kondo returns the analysis. We convert the filename back into a URI (with shared/filename->uri). But we leave it unescaped, which causes the bug: we update the DB, but keep using the unescaped jar file URI. So, the DB is still full of unescaped URIs.

So, now, you're in an external file on a usage of a var defined in another external file and you want to go to that definition. When the client makes this request, it provides the source file's ESCAPED URI. But we don't have any analysis for the escaped URI, only for the unescaped URI. We can't look up the var usage, and therefore can't tell the client where it's defined.

That's why the second navigation is broken. @ericdallo do you feel the integration tests sufficiently cover this two-step navigation (from an internal file, to a jar file, to a second jar file)?

So to summarize, after the change in this PR things are a bit different. At startup and in didOpen we saved escaped jar file URIs. There's a minor change on the first navigation (internal file to jar file). We now tell the client to open an escaped URI. (Does that work? I assume so, but I'll test it to make sure.) On the second navigation (jar file to second jar file) we actually have analysis, so we can provide a destination.

I see why you're hesitant about the way this is being fixed. The fix `(= "zipfile" (:dependency-scheme client-settings))` has two downsides:

* It doesn't fix the bug for any client that escapes URIs this way, unless they also use the zipfile scheme.
* It breaks any other clients that do use zipfile but don't escape.

I guess this is OK as a quick fix. We think that coc.nvim is the only client that escapes this way, and also think it's the only client that uses zipfiles. So as for the first downside, we're only introducing a bug if future clients escape but don't use zipfiles. Similarly, for the second downside, we're only introducing a bug if future clients use the zipfile scheme but don't escape.

For a longer term fix, I suppose we've forced all the clients to include `:dependency-scheme "jar"` in their initialization options, just so that coc.nvim can omit `:dependency-scheme "zipfile"` from its options. Should we do that again, but this time force coc.nvim to include `:uri-scheme "escaped-path"` or something like that? @NoahTheDuke are you amenable to that?

This file name code is so delicate, with so many little hacks. Someday we should take a step back, document all the ways URIs and filenames are used (by clients and by clj-kondo on Windows and UNIX-likes), and come back to it with fresh eyes.

cli/integration-test/integration/definition_test.clj
mainej: Would `with-redefs` work? That'd let you avoid having to remember to set the var back to `false` at the end of the test.

mainej:
Does this need a bump to `clojure-lsp.db/version`? Or do we tell coc.nvim users to delete their cache?

mainej:
I apologize, but I can't test this... something is wrong with my nvim. When I try to open an external definition it always takes me to a blank buffer. @dharrigan would you mind confirming that a build from this branch fixes thing for you?

dharrigan:
Hi,

Unfortunately, I don't think it works. I've included a window recording. I'm opening up the main.clj file, waiting for analysis to complete, then `gd`'ing into validate. Now nothing is shown.

Peek 2022-10-16 06-05

mainej:
@dharrigan your video answers a question buried in my long comment.

> We now tell the client to open an escaped URI. Does that work?

The answer, apparently, is no. coc.nvim gives clojure-lsp escaped URIs, but cannot accept them in return. (I ❤️ all the clients, but hacking around individual clients is never a fun part of working on an API.)

What this means, I think, is that clojure-lsp has to return unescaped URIs to all clients. I see two paths to doing this. First, we could unescape URIs as they're leaving clojure-lsp, in our responses. Alternatively, we could escape URIs as they're entering clojure-lsp, which is what we were doing before by calling `uri->filename` on everything.

I've been pushing for the DB to contain URIs in the same format that the clients send, which would suggest we should take the first option—unescape URIs on the way out of clojure-lsp. But that no longer feels like the best approach. First, we know that unescaping URIs on the way in works—that's how the system worked before. Second, looking through the spec, it's much easier to unescape the client's requests. The server's responses include URIs in many many deeply nested locations. On the contrary, almost all of the client's requests are uniform—there's only one URI, and it's at `[:text-document :uri]`.

@ericdallo I think we can write an integration test to cover these scenarios, but I'm not confident that seeing a green integration test is sufficient. There's too much subtlety in how individual clients work. We need to reproduce this for ourselves, in nvim. I'm going to start from scratch getting my nvim to work. Maybe this time I'll try [@dharrigan's config](https://github.com/dharrigan/vimconfig).

I have time to work on this today, so I'll do that. @ericdallo I'll start from this branch. Mind if I push changes to it?

ericdallo:
Sorry for the delay and thanks for the complete explanation @mainej that's exactly what I found after some debugging.

> That's why the second navigation is broken. @ericdallo do you feel the integration tests sufficiently cover this two-step navigation (from an internal file, to a jar file, to a second jar file)?

I think so, we call hover on the opened jar which is similar to find-definition from a second jar, WDYT?

> This file name code is so delicate, with so many little hacks. Someday we should take a step back, document all the ways URIs and filenames are used (by clients and by clj-kondo on Windows and UNIX-likes), and come back to it with fresh eyes.

For sure, I'd love to see all upstream tools using URIs, starting from clj-kondo, then we wouldn't need to do all this magic probably

> Does this need a bump to clojure-lsp.db/version? Or do we tell coc.nvim users to delete their cache?

Yes, sounds safer to bump db version, I don't see a problem with that.

> I've been pushing for the DB to contain URIs in the same format that the clients send, which would suggest we should take the first option—unescape URIs on the way out of clojure-lsp. But that no longer feels like the best approach. First, we know that unescaping URIs on the way in works—that's how the system worked before. Second, looking through the spec, it's much easier to unescape the client's requests. The server's responses include URIs in many many deeply nested locations. On the contrary, almost all of the client's requests are uniform—there's only one URI, and it's at [:text-document :URI].

Agreed, My first idea was to escape as we did before, but then I thought the same, save in DB as client sends, but on a second thought, this could be dangerous as we have lots of outputs and URIs that come from other upstream like kondo, so it sounds safer to just do the same we did in the past 

> @ericdallo I think we can write an integration test to cover these scenarios, but I'm not confident that seeing a green integration test is sufficient. There's too much subtlety in how individual clients work. We need to reproduce this for ourselves, in nvim. I'm going to start from scratch getting my nvim to work. Maybe this time I'll try @dharrigan's config.

Agreed, I'll find some time this week to do the same.

> I have time to work on this today, so I'll do that. @ericdallo I'll start from this branch. Mind if I push changes to it?

Not at all, feel free to do it

NoahTheDuke:
> coc.nvim gives clojure-lsp escaped URIs, but cannot accept them in return

This seems like a coc.nvim bug unless it's how the LSP spec is written. I can reach out to them and see about fixing or working around it.

mainej:
> I think so, we call hover on the opened jar which is similar to find-definition from a second jar, WDYT?

@ericdallo ... I think it's different. `hover` doesn't return a URI, but `find-definition` does. We need to be checking that URI is in the right format.

> This seems like a coc.nvim bug unless it's how the LSP spec is written.

@NoahTheDuke I couldn't find anything about escaping in the spec. So, predictably since it's unspeced, you get the free-for-all that we're seeing. My guess is that each client and server is just doing whatever their URL lib does, including to some extent clojure-lsp. Anyway, yes, coc.nvim may be doing it wrong, but so are others. For example, I discovered today that Calva escapes the entire jarfile including the second colon `"jar:file%3A///some/path/some.jar%21/some/file.clj"`. This is totally different than other clients, but I can't even say it's wrong. It's invalid to have a colon in a URI scheme, so `jar:file://...` is technically an invalid URI. Such a mess.

mainej:
I eventually got nvim working and was able to reproduce the problem. @dharrigan and/or @snoe will you test this branch? The patch in its current form fixed the problem for me.

cli/src/clojure_lsp/server.clj
```diff
@@ -459,7 +494,7 @@
   ;; about which messages are sent when probably needs to be handled in lsp4clj.
   ;; https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize
   (handler/initialize components
-                      (:root-uri params)
+                      (normalize-uri (:root-uri params))
```
mainej, line 497: I'd like some help thinking about this change. I think that this means that the root-uri never contains `%3a` anymore, which means that `:encode-colons-in-path?` is always `false`, and therefore should be removed.

@LuisThiamNye you introduced `:encode-colons-in-path?` in #439. I don't know if you still follow this project, but if you do, can I recruit your help? Do you remember when you saw colons encoded as `%3a` in the root-uri? What OS/editor/client/repo combination was it? Can you test this branch there?

I think `:encode-colons-in-path?` was for Windows, but I don't have access to a Windows machine. @ikappaki, you've done the most recent Windows maintenance. Do you have any insight or concerns about this change?

Apologies for bringing you both in without much context. As a really quick summary, the intention of this line and the other changes to server.clj is to put every URI we receive from a client into a format that A) will match the URIs we get from clj-kondo, and B) can be used by the editors to open files, especially jar files. By doing this uniformly in server.clj, we don't need as much special logic in other places to convert URIs into particular formats.

@ericdallo I did all my testing with `(normalize-uri (:root-uri params))`, so I know it works on a Mac with emacs-lsp, Calva and coc.nvim. If we decide to revert this line, we have to remember to test all those again.

ericdallo:
> @ericdallo ... I think it's different. hover doesn't return a URI, but find-definition does. We need to be checking that URI is in the right format.

Makes sense, testing a definition for a jar or anything that will return a URI sounds better indeed

cli/src/clojure_lsp/server.clj
ericdallo: 👍

I guess project-uri will never come with colons but I may be wrong.

cli/integration-test/integration/definition_test.clj
ericdallo: I think so, for a moment I thought it was not thread-safe

lib/src/clojure_lsp/shared.clj
```diff
@@ -529,3 +531,13 @@
   expects Clojure style JSON, or when a map needs to be round-tripped from
   clojure-lsp to the client and back without case changes."
   lsp.responses/preserve-kebab-case)
+
+(defn normalize-uri-from-client [uri]
```
ericdallo, line 535: I wonder if we should have this as a lsp4clj flag, it seems to me that all servers could have this encode problem from clients, so a easy flag to enable that or not may be a good idea

cli/src/clojure_lsp/server.clj
LuisThiamNye: > Do you remember when you saw colons encoded as `%3a` in the root-uri?

According to [this comment](https://github.com/clojure-lsp/clojure-lsp/pull/439#issuecomment-851614968), I observed this with Calva on Windows to denote the root drive (`c:` → `c%3A`).

lib/src/clojure_lsp/shared.clj
mainej: Agreed, it's a shared problem. But since lsp4clj doesn't know about the structure of any of the inbound LSP messages, it'd be hard for it to do the conversion itself. This helper function would definitely be a candidate for https://github.com/clojure-lsp/lsp4clj/issues/14 though.This is another thing that might have been easier if we'd modeled the message handlers as routes instead of multimethods.

cli/src/clojure_lsp/server.clj
mainej: Thanks @LuisThiamNye. The unit tests [already cover](https://github.com/clojure-lsp/clojure-lsp/pull/1327/files#diff-8c9f99b87bb2d221ae57cf20e622aac544c6b1b813f227be77c6c7784783b070R234) `c:` being encoded as `c%3a`. I hope that means Calva will work on Windows, but if you or @ikappaki have a chance to confirm, I'd appreciate it.

ikappaki: Hi @mainej,

I had a quick play with this branch and Calva, and navigation from the clojure-lsp project to external libs appears to be working as normal. Do you have a particular test case you want us to try out?

Amongst other things, I've tried the following which I think is relevant since I'm jumping into jars:
1. Open `cli/src/lcojure_lsp/server.clj` and right click on `timbre`, goto definition, which gets me into the timbre-5.2.1.jar
2. Right click on `reduce` keyword, goto definition, it gets me into the clojure `core.clj` jar.

Thanks

ikappaki: Hi @mainej,

I confirm I can jump from `cli/src/lcojure_lsp/server.clj` -> `~\.m2\repository\com\taoensso\timbre\5.2.1\timbre-5.2.1.jar!\taoensso\timbre.cljc` -> `~\.m2\repository\com\taoensso\timbre\5.2.1\timbre-5.2.1.jar!\taoensso\timbre\appenders\core.cljc` with no issue.

mainej: Brilliant! Thanks @ikappaki :pray:

PR #1331, opened 2022-10-21
Allow completing any java* class

### LINKED ISSUE #1329: Cannot Complete Java Classes
**Describe the bug**

I cannot complete Java classes in clojure-lsp in Emacs. Regular clojure (ie `map`, `filter`, etc) completes just fine.

**To Reproduce**

Type `java.io.F` and trigger completion. You will not see any candidates.

**Expected behavior**

If I type `java.io.F` and trigger completion, I should see candidates.

**Screenshots**

Trying to trigger completion of Java classes in Emacs using clojure-lsp:

With cider it works as expected:

 - OS: MacOs
 - Editor: Emacs 29
 - Version: clojure-lsp 2022.10.05-16.48.26-nightly, clj-kondo 2022.10.05

**Additional context**

I have an empty `deps.edn` file:

```clojure
{}
```

I assumed that Java standard library stuff would be picked up by clojure-lsp automatically, as it is with cider. But maybe I have a classpath issue, with clojure-lsp?

Great project btw - amazing stuff.

ericdallo:
Fixes #1329 

@mainej I'd like your review on this if possible.

Since we now have kondo analysis for java-class-definitions, we could stop using the pre-defined symbols.

My first idea was to support completing all known java-class-definitions, this worked but affected performance going from ~50ms -> ~100ms (~1000 -> 27000 completion items) probably because on sort done later (maybe we could consider not sorting java classes?).

Then I thought about only adding "java" classes instead of all known classes, excluding `com*` `org*` etc, this reduced to 5900 completion items but yet 70-80ms.

Then I though avoiding empty completions returning all the 5900 items even if user doesn't want java stuff, we know some editor call completion all the time even without user input, so I added a start-with for `j`, not ideal, but it's safer.

It'd be awesome if we could provide the 27000 items without any performance issue, but not sure if is possible, LMK what you think


ericdallo:
I need to fix tests yet

mainej (approved):
I understand the issue with having too many options. Some thoughts in no particular order:

1. IMO even 1000 is too many options to offer. The implication is that we think that someone is going to find what they're looking for in that list _without typing another character_. That is, they're going to scroll through 1000 options, until they find the 1 symbol that they want. I think if a user is that lost—they can't think of a single letter of a single word related to what they want to do—they should get oriented using some other tool besides clojure-lsp, probably the internet. I would be perfectly fine returning 0 results until at least 1 character is typed, or maybe even 2.
2. The starts-with-j trick makes sense, but I think 5.9K results is too many for the same reason.
3. By filtering for "java", we won't offer completions for classes defined _in_ someone's Clojure project. Those classes are probably the ones they want the most. If possible, I'd split this into two lists, the internal and external class definitions. I'd return all items from the internal list, but use the starts-with-j trick, or something like it, for the external list.
5. An alternative to starts-with-j is to base the match on some deeper part of the class name. That is, `j|` and `java.ne|` offer nothing. It's not until you type `U|` or `java.net.U|`, that we would offer `java.net.URI`. That might create a smaller more useful list of options. This would have to be tested... it may be that the performance is related to traversing the whole list of class definitions, but not to the number of completion items we end up returning.
6. A related idea is to only search class definitions if the search term includes a capital letter.
7. We had some performance problems with returning 1000 items before, which I think @snoe fixed. The fix (or one of many fixes) was to stop querying the db for things defined in clojure.core and cljs.core, and instead hard-code those into `common-symbols`. This wasn't as dynamic, but it was a lot faster. I think we generated the hard-coded list by querying _some_ database for all the core items, and copying them. We could try something similar for Java, using the code you have here. We'd query _some_ database for the `java.*` class definitions, then hard-code them in `common-symbols` as we had before. You'd have to benchmark whether this actually makes a difference... some of the other later db refactorings may mean that looking up all these items isn't a bottleneck anymore.

All that said, I think that what you have currently would be fine as a first draft. It's up to you whether to incorporate some of these ideas now or postpone them to the future.

ericdallo:
You are absolutely right, I didn't think about what you explained on 1, but makes sense, IMO we should check the length of the cursor value before returning results, and yes, I think we can filter user class too, I will take a look again on this

ericdallo:
@mainej I took a different approach, limiting all completion items to 800 (a number we could move to a setting if needed or even decrease a little bit more), from my tests, limiting makes really sense as you mentioned that users could enhance their query for better results if needed.

About java completions, we are not limiting only java classes but all java classes which seems better but we will only filter java class definitions from analysis if the cursor value is not empty since this improved performance of empty completions. Let me know what you think

lib/test/fixtures/java_interop/System.java
```diff
@@ -0,0 +1,5 @@
+package java.lang;
```
ericdallo, line 1: I didn't find a way to pass a code to `clj-kondo` as input for java class definition lint as we have for clojure, maybe this should be improved on clj-kondo side c/c @borkdude

lib/src/clojure_lsp/feature/completion.clj
```diff
@@ -422,20 +422,30 @@
...
+                      (let [class-name* (delay (last (string/split class #"\.")))]
+                        (cond-> []
+
+                          (matches-fn class)
+                          (conj {:label (str class)
+                                 :kind :class
+                                 :priority :java-class-definitions})
+
+                          (and (string/starts-with? class "java.lang")
+                               (matches-fn @class-name*))
```
mainej, line 443: As I read this, this lets you type `Str` and get a completion for `java.lang.String`... Is that correct? I think this would be more valuable if it filtered for "java" not "java.lang". The [java.lang package](https://docs.oracle.com/javase/7/docs/api/java/lang/package-summary.html) defines only 100 classes or so. They're used a lot, but a lot of the action is in other packges—java.util, java.concurrent, java.net, etc.

lib/src/clojure_lsp/feature/completion.clj
```diff
@@ -475,6 +485,12 @@
        (mapv #(dissoc % :priority))
        not-empty))
 
+(defn ^:private limiting-items
+  "Limit the returned items for better performance.
+  If user needs more items one should be more specific in the completion query."
+  [size items]
+  (drop-last (- (count items) size) items))
```
mainej, line 492: This looks a lot like `(take size items)`. Are there performance differences?

mainej (approved):
A few comments/suggestions, but looks good!

lib/test/fixtures/java_interop/System.java
borkdude: Can you explain in more words what you mean here?

borkdude: E.g. this works:

``` clojure
 $ cat /tmp/Foo.java
package dude;

public class Foo {

}

$ clj-kondo --lint /tmp/Foo.java --config '{:output {:analysis {:java-class-definitions true} :format :edn}}'
{:findings [], :summary {:error 0, :warning 0, :info 0, :type :summary, :duration 28, :files 0}, :analysis {:namespace-definitions [], :namespace-usages [], :var-definitions [], :var-usages [], :java-class-definitions [{:class "dude.Foo", :uri "file:/private/tmp/Foo.java", :filename "/private/tmp/Foo.java"}]}}
```

ericdallo: @borkdude I mean using `-`, as the stdin, as we have for clojure code, otherwise we would always have to have the file on disk for tests

lib/src/clojure_lsp/feature/completion.clj
ericdallo: No, the idea here is able to complete java lang classes by its name si cê Clojure allow you use it without importing, like `String`.
So we need to return these items without the package in the name, does that makes sense ?

lib/src/clojure_lsp/feature/completion.clj
ericdallo: Uhh, completely forgot take here :facepalm: I'll change it

lib/test/fixtures/java_interop/System.java
borkdude: OK, issue welcome

lib/src/clojure_lsp/feature/completion.clj
mainej: Ah, I see, yes, that's a nice feature. 👍

PR #1334, opened 2022-10-24
Expand and correct docs about :dependency-scheme setting

mainej:
This corrects some discrepancies between the code, `doc/settings.md` and `doc/all-available-settings.edn` by matching the docs to the code's actual behavior.


docs/settings.md
```diff
@@ -68,42 +68,42 @@ This is useful if you have some rule to apply to clojure-lsp for multiple projec
...
+| `:dependency-scheme`                           | Defines the format that the client and server should use for URLs of dependencies in jars. The two valid values are `"jar"` and `"zipfile"`. Different editors may handle these URLs differently. For example, in nvim when a `"zipfile"` URI is opened, the file will be editable, but `"jar"` URIs will not be. (This can be powerful, but if you edit a jar in your .m2 directory, you may break other projects.) `"jar"` will make urls compatible with java's JarURLConnection. Clients that can't extract the jar natively can make an LSP extension request of `clojure/dependencyContents` with the jar uri and the server will return the jar entry's contents, an approach [used by some Java clients](https://github.com/redhat-developer/vscode-java/blob/a24945453092e1c39267eac9367c759a6c7b0497/src/extension.ts#L290-L298) | `"zipfile"`                                                                                   |
```
ericdallo, line 84: If we decide to not drop zipfile, we could at least change the default to `jar` IMO

ericdallo (approved):

docs/settings.md
mainej: I think very few people use "zipfile" intentionally. But I also know that lots of people have set `:jar` because that's what the docs said, and so have been using "zipfile" accidentally. We could change the default, but we might expose ourselves to situations that have been working with "zipfile" that don't work with "jar". Maybe that's not a bad thing though... we really want "jar" to be robust.

PR #1335, opened 2022-10-25
Fix key to mark code actions preferred

mainej:
This fixes the `Add require` code actions to use the `"isPreferred"` key correctly, so that they are sorted first. Before the migration away from lsp4j we were setting this key correctly, which made editors show the these code actions first. See https://github.com/clojure-lsp/lsp4clj/issues/32 for more context about lsp4j.

Unfortunately, this doesn't fix Emacs lsp-mode. IIRC, it used to sort the `Add require`s first, but even with this change it doesn't.

- [x] Addresses https://github.com/clojure-lsp/lsp4clj/issues/32


ericdallo (approved):
Nice catch

PR #1350, opened 2022-10-29
Bump graal to 22.3.0

ericdallo:


ericdallo:
@ikappaki it seems graal 22.3.0 doesn't work with UPX again for windows :( WDYT? maybe keep it as 22.2.0 or remove UPX integration for windows?

ikappaki:
> @ikappaki it seems graal 22.3.0 doesn't work with UPX again for windows :( WDYT? maybe keep it as 22.2.0 or remove UPX integration for windows?

HI @ericdallo, it appears the hack is not applicable to 22.3 but has been reported upstream at https://github.com/oracle/graal/issues/4340, I suggest we keep the status quo for the moment (I presume we do care about the executable size, 26 vs 115 MB) and see if a workaround emerges.

Thanks

PR #1371, opened 2022-11-12
Improves :kind on symbols, for better iconography

mainej:
When returning symbols from `workspace/symbol` and `textDocument/documentSymbol`, the LSP spec let's you say what "kind" of symbol each is. The clients use the "kind" to pick an icon to display next to each symbol.

Confusingly, for all of the following symbols, clojure-lsp assigned the "kind" as "variable". Therefore these symbols' icons were indistinguishable from symbols defined by `def`.

This patch changes the "kind" as follows.

```clojure
(defn fn-defn-varargs [a & args] a)                 ;; variable -> function (function with var args)
(defmulti interface-defmulti identity)              ;; variable -> interface
(defmethod interface-defmulti :fn-defmethod [x] x)  ;; variable -> function
(defprotocol InterfaceDefprotocol)                  ;; variable -> interface
(definterface InterfaceDefinterface)                ;; variable -> interface
(defrecord ClassDefrecord [])                       ;; variable -> class
(deftype ClassDeftype [])                           ;; variable -> class
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
   - Bump clj-kondo to `2022.11.03-20221105.203751-5`.
 
 - Editor
-  - Show better icons for multimethods, var-arg fns, protocols, records, interfaces and types.
+  - Show better icons for multimethods, var-arg fns, protocols, records, interfaces and types on `workspace/symbol` and `textDocument/documentSymbol`.
 
 ## 2022.11.03-00.14.57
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
   - Only publish progress on initialize if client provided a `workDoneProgress`. #1363
   - Bump clj-kondo to `2022.11.03-20221105.203751-5`.
 
+- Editor
+  - Show better icons for multimethods, var-arg fns, protocols, records, interfaces and types.
```
ericdallo, line 10:
```suggestion
  - Show better icons for multimethods, var-arg fns, protocols, records, interfaces and types on `workspace/symbol` and `textDocument/documentSymbol`.
```

ericdallo (approved):
Looks great!

Issue #210, opened 2020-12-19
Inline-symbol doesn't remove let if symbol was only binding in let


bpringe:
Expected:

* Running inline-symbol on a symbol in a `let`, which is the only binding in the `let`, removes the `let`.

Actual:

* When I run inline-symbol and the symbol is from the only binding in a let, the let is left with an empty binding vector.

  ![inline-symbol-let](https://user-images.githubusercontent.com/12722744/102677503-a0c29900-4157-11eb-9509-3e14d61fb20d.gif)

ericdallo:
This is not quite easy since we need to check some corner cases like 

```clojure
(defn ii []
  (let [x 1]
    (+ 2 x)
    (+ 2 4)))
```
should be:
```clojure
(defn ii []
  (+ 2 1)
  (+ 2 4))
```
`rewrite-clj`, the lib we heavily rely on changing/analyzing Clojure code, doesn't allow replace a form with multiple forms like `(+ 2 1) (+ 2 4)`, is not a correct Clojure form so I'm not sure how to fix that, replacing the `let` by a `do` form would help implement too:
```clojure
(defn ii []
  (do
    (+ 2 1)
    (+ 2 4)))
```

bpringe:
I see. I saw the let removal in this gif, which is what made me expect it: https://github.com/clojure-emacs/clj-refactor.el/blob/master/examples/inline-symbol.gif

I realize that is not rewrite-clj, but I arrived there from the readme link pointing to those examples as a reference for what the refactor commands do.

ericdallo:
Yes, maybe we can check how they do, but we may need rewrite-clj anyway

snoe:
You should be able to do this with rewrite-clj by taking the children of the let and adding them to the defn node.

One day we'll switch to rewrite-cljc and leaning more heavily on its paredit API (and expanding) would make this code so much easier.

ericdallo:
@snoe what is needed to switch to rewrite-cljc ? I can help with that

snoe:
@ericdallo this pr https://github.com/clojure-lsp/clojure-lsp/pull/150 
but on clojurians yesterday I saw this from @lread so this is is some progress that means we can switch to the clj-commons version asap and then just double check things work with the cljc playground merges (which I believe lread already tested clojure-lsp)

> Great news, xsc/rewrite-clj has been transferred to clj-commons/rewrite-clj!
> Much thanks to @xsc and @slipset!
> 
[...]

PR #1381, opened 2022-11-18
Inlining the sole let binding removes the let

mainej:
This fixes https://github.com/clojure-lsp/clojure-lsp/issues/210.

```clojure
(let [|a 1] (inc a))
;; before
(let [] (inc 1))
;; after
(inc 1)
```

If the body of the let contains several forms, they will be wrapped in a `do`.

```clojure
(let [|a 1] (dec a) (inc a))
;; before
(let [] (dec 1) (inc 1))
;; after
(do (dec 1) (inc 1))
```

However, if the let is wrapped in a form that introduces an implicit `do`, the explicit `do` will be omitted.

```clojure
(try (let [|a 1] (dec a) (inc a)))
;; before
(try (let [] (dec 1) (inc 1)))
;; after
(try (dec 1) (inc 1))
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 - Editor
   - Show better icons for multimethods, var-arg fns, protocols, records, interfaces and types on `workspace/symbol` and `textDocument/documentSymbol`.
+  - Inlining the last binding of a let removes the let. #210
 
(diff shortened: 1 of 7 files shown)


ericdallo (approved):
Awesome! loved the `establishes-implicit-do` feature, thank you!

PR #1395, opened 2022-12-01
Add missing ns snippet paren

Invertisment:
This is a single symbol. Nobody will notice that it's fixed. I don't think there is any point to mention it in release log.


ericdallo:
Thank you!

Issue #1403, opened 2022-12-03
Support jump to definition for quoted symbols


mk:
**Is your feature request related to a problem?**

I'd like to be able to jump to the definition of a quoted symbol, e.g.:
```clojure
;; quote
(requiring-resolve 'clojure.set/rename-keys)

;; alias + sharp quote
(alias 's 'clojure.set)
(requiring-resolve `s/rename-keys)

;; alias + normal quote
(alias 's 'clojure.set)
(requiring-resolve 's/rename-keys)
```
(the last should probably try to resolve `'s/rename-keys` first and only if that fails fall back to trying `'clojure.set/rename-keys`

**Additional context**
Besides the requiring-resolve examples above, this would also be great for render-fns in Clerk e.g. https://github.com/nextjournal/clerk/blob/edd0c2fc22bd0207cc8d3762f00f239ce63ee87f/src/nextjournal/clerk/viewer.cljc#L733

We've previously discussed this on slack: https://clojurians.slack.com/archives/CPABC1H61/p1668186899410569

In that slack thread @ericdallo pointed out the current resolution code:

https://github.com/clojure-lsp/clojure-lsp/blob/f82e9e71ddd5780e766a3475515612982ad323a0/lib/src/clojure_lsp/queries.clj#L502-L506

ericdallo:
@mk rechecking this, it seems it'd help a lot if clj-kondo could provide additional __symbol analysis__, otherwise we would need to change some core code to not rely only on kondo analysis but do our custom parsing which is something I'd like to avoid for maintainability reasons.

we used to have issues like that in the past like no keyword analysis, java analysis, and adding in kondo helped a lot of things keep simple, so a flag to analyze symbols as well sounds like something that would help not only 1 or 2 features to work with symbols but probably all clojure-lsp features since most of them rely on kondo analysis.

WDYT @borkdude? Since it would be a new context flag or something, it looks to me good enough to be kondo that provides that info, also it's something that makes sense to me to kondo provide for a project

borkdude:
@ericdallo clj-kondo needs context in which to resolve a symbol. e.g. if a symbol is `a/foo` it needs to know the aliases and imports of that namespace. Given that lsp already has this information from prior calls to clj-kondo, but clj-kondo doesn't have this context, wouldn't it be more evident to do this in lsp? I think you would just have to check the aliases and imports to get the fully qualified name, or am I wrong?

ericdallo:
[...]

PR #1405, opened 2022-12-05
Use CLJS vars as fallback, for clerk render fns

borkdude:

Related issues:

- https://github.com/clojure-lsp/clojure-lsp/issues/1403
- https://github.com/nextjournal/clerk/issues/101

I'll add the changelog entry after review.

### DIFF
--- a/lib/src/clojure_lsp/queries.clj
+++ b/lib/src/clojure_lsp/queries.clj
@@ -250,7 +250,7 @@
       (let [definition (find-definition db (assoc var-usage :lang :clj))]
         (when (:macro definition)
           definition)))
-    ;; fallback to navigate from clojure to clojurescript vars
+    ;; Fallback to navigate from clojure to clojurescript vars, see #1403
     (find-definition db (assoc var-usage :lang :cljs))))
 
 (defmethod find-definition :local-usages
--- a/lib/src/clojure_lsp/queries.clj
+++ b/lib/src/clojure_lsp/queries.clj
@@ -249,7 +249,9 @@
       ;; maybe loaded by :require-macros, in which case, def will be in a clj file.
       (let [definition (find-definition db (assoc var-usage :lang :clj))]
         (when (:macro definition)
-          definition)))))
+          definition)))
+    ;; fallback to navigate from clojure to clojurescript vars
+    (find-definition db (assoc var-usage :lang :cljs))))
 
 (defmethod find-definition :local-usages
   [db {:keys [id uri] :as _local-usage}]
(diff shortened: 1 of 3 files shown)


borkdude:
@ericdallo This makes navigation from clojure to clojurescripts vars in clerk viewer functions work, given the right clj-kondo configuration for `->viewer-fn`, etc. See demo :)

clj- cljs-nav

lib/src/clojure_lsp/queries.clj
```diff
@@ -249,7 +249,9 @@
       ;; maybe loaded by :require-macros, in which case, def will be in a clj file.
       (let [definition (find-definition db (assoc var-usage :lang :clj))]
         (when (:macro definition)
-          definition)))))
+          definition)))
+    ;; fallback to navigate from clojure to clojurescript vars
+    (find-definition db (assoc var-usage :lang :cljs))))
```
ericdallo, line 254: I can't easily understand why this is important, AFAIK a clj code is not possible to call cljs ones right? only cljc -> cljs.
Also, the when above already handles cljs

ericdallo: Maybe create a test that makes a little bit clear this?

borkdude: @mk will explain the details, but try this branch of clerk:

https://github.com/nextjournal/clerk/tree/jump-to-definition

Go to the `notesbooks/jump_definition.clj` notebook and try navigation from inside of the quoted function. You will see that it doesn't work. In this PR we added a fallback for this with the idea of navigating to something is better than nothing.

I used `bb debug-cli` btw and it was very useful, thank you.

ericdallo: Cool :)
I understood the idea, I just have some small concerns/questions:
- Why clerk is the only tool that makes somehow possible to go from clj -> cljs?
- Could this add some confusion since will affect all clojure-lsp users, like people who have both namespaces one for clj and another for cljs and find definition goes to the wrong one? 
- Adding a test may help understand that better, WDYT?

mk: > Why clerk is the only tool that makes somehow possible to go from clj -> cljs?

I don't think Clerk is the only tool. We use the ability to delay evaluation in various projects, e.g. the frontend sends form to the backend to perform effects similar to how Clerk sends forms representing viewer functions to the browser. The separation of `read` and `eval` is a fundamental mechanism in a lisp and I'd love to see lsp support it better.

> Could this add some confusion since will affect all clojure-lsp users, like people who have both namespaces one for clj and another for cljs and find definition goes to the wrong one?

While I can't say this will never happen I would argue that the bug to fix would then be not finding the Clojure definition. As it's now implemented its only using the ClojureScript variant as a fallback.

ericdallo (approved):
Looks good, thank you!

ericdallo:
@borkdude there is a minor lint issue, could you fix so we can merge ?

ericdallo:
Maybe some weird config from [here](https://github.com/clojure-lsp/clojure-lsp/blob/master/.cljfmt.edn)

Issue #1446, opened 2023-01-11
Support navigation of quoted symbols


ericdallo:
https://github.com/clj-kondo/clj-kondo/issues/1902#issuecomment-1378529454

borkdude:
Made a PR here: https://github.com/clojure-lsp/clojure-lsp/pull/1447

This makes the following navigation work:

![viewer-fn-nav](https://user-images.githubusercontent.com/284934/211887915-99a445af-af21-4bff-bd3f-ae90aa00af22.gif)
![exec-fn-nav](https://user-images.githubusercontent.com/284934/211887920-26bc20a5-5102-4f7b-81f1-0db37f278513.gif)

cc @mk

PR #1447, opened 2023-01-11
Symbol analysis

borkdude:
- [x] Fixes https://github.com/clojure-lsp/clojure-lsp/issues/1446

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,11 @@
 ## Unreleased
 
 - General
-  - Bump clj-kondo to `2022.12.11-20221220.093423-5`
+  - Bump clj-kondo to `2022.12.11-20230111.135117-16`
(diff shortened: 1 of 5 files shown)


borkdude:
@ericdallo Please let me know if I should squash the commits of if you are going to squash them when merging

mainej:
This is exciting! Will at address #1246? Is https://github.com/clj-kondo/clj-kondo/issues/1576 also fixed?

borkdude:
Not exactly. The only thing this does is give locations for quoted qualified symbols and an additional `:to` field which contains a namespace to which the qualified symbol might refer to (given the aliases in the same namespace). This helps navigation for symbols like `:exec-fn clojure.core/prn` and `(requiring-resolve 'clojure.core/inc)`, but that's about it.

ericdallo (approved):
Looks good, tested and it works great!
I'll add support for references on master, thank you @borkdude !

Issue #1506, opened 2023-02-22
Enhance hover to search clojuredocs on CLJS 


RCmerci:
**Is your feature request related to a problem? Please describe.**
when writing cljs codes, hovering on fn, and the returned doc doesn't contain clojuredocs.

**Describe the solution you'd like**
when file-type is :cljs, replace(ns-sym "cljs" "clojure")

**Describe alternatives you've considered**
nil

**Additional context**
nil

ericdallo:
Fixed on #1507

PR #1507, opened 2023-02-22
enhance(hover): lookup clojuredocs when using cljs

RCmerci:
- [X] I created an issue to discuss the problem I am trying to solve or an open issue already exists.
- [X] I added a new entry to [CHANGELOG.md](https://github.com/clojure-lsp/clojure-lsp/blob/master/CHANGELOG.md)
- [X] I updated documentation if applicable (`docs` folder)

related issue: #1506

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1132,3 +1132,4 @@ Huge refactor https://github.com/clojure-lsp/clojure-lsp/pull/261 which uses clj
 - LSP 3.16: Add support for `codeAction/resolve` improving performance if client supports it
 - Bump extend lib
 - [CI] Remove auto release, next releases should contain more than one PR/fix
+- Enhance hover to search clojuredocs on CLJS 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
   - Fix `textDocument/hover` issue when doc metadata isn't a string literal.
   - Follow references to other vars in doc metadata for use in `textDocument/hover`.
   - Support `:arglists` meta when finding docs of functions via hover feature. #1504
+  - Enhance hover to search clojuredocs on CLJS 
 
 ## 2023.01.26-11.08.16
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1131,3 +1131,4 @@ Huge refactor https://github.com/clojure-lsp/clojure-lsp/pull/261 which uses clj
 - LSP 3.16: Add support for `codeAction/resolve` improving performance if client supports it
 - Bump extend lib
 - [CI] Remove auto release, next releases should contain more than one PR/fix
+- Enhance hover to search clojuredocs on CLJS
```
ericdallo, line 1134: The changelog entry should be placed at the start of the file 😅

lib/src/clojure_lsp/feature/hover.clj
```diff
@@ -123,7 +123,12 @@
                             (str join-char signatures)))
         markdown? (some #{"markdown"} content-formats)
         doc-line (find-docstring db markdown? uri doc 0)
-        clojuredocs (f.clojuredocs/find-hover-docs-for sym-name sym-ns db*)
+        clojuredocs (or (f.clojuredocs/find-hover-docs-for sym-name sym-ns db*)
```
ericdallo, line 126: Can we add a test for that? otherwise if we change this behavior we wouldn't know we introduced a regression.
A unit test should be enough on hover_test.clj

ericdallo (commented):
Looks good, would be nice to have a test for that but I just noticed we don't have tests for clojuredocs integration yet, I think we can merge after updating the changelog properly

CHANGELOG.md
RCmerci: updated. Misread 2021 as 2023 😅

RCmerci: updated. Misread 2021 as 2023 😅

ericdallo (approved):

ericdallo:

PR #1513, opened 2023-03-10
add index.html for api/, fixes broken link

scottbale:
I noticed a broken link in [lein-clojure-lsp doc](https://github.com/clojure-lsp/lein-clojure-lsp#usage) to https://clojure-lsp.io/api/ - it appears to be broken because there's no index.html file there. 

As a quick fix, this PR adds one which does a redirect to `api/what-is-it/`.  It is just a copy of `api/what-is-it/index.html` but with relative links adjusted and adding the redirect in a `meta` tag with "Refresh".

Alternative ideas
* We could just change the link in the lein-clojure-lsp doc to point to `api/what-is-it/`
* Some sort of landing page at `api/index.html` rather than a redirect
* Could the `api/what-is-it/` URL just be `api/` instead?

### DIFF
--- a/api/index.html
+++ b/api/index.html
@@ -1,0 +1,799 @@
+
+<!doctype html>
+<html lang="en" class="no-js">
+  <head>
+
+      <meta http-equiv="Refresh" content="0; url='what-is-it/'" />
+      <meta charset="utf-8">
+      <meta name="viewport" content="width=device-width,initial-scale=1">
+      
+        <meta name="description" content="Language Server Protocol Implementation for Clojure">
+      
+      
+      
+      
+        <link rel="canonical" href="https://clojure-lsp.io/api/">
+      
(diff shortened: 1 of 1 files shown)


ericdallo:
Hey @scottbale 👋🏻, thanks for noticing that, but I think we can probably take a different approach, all clojure-lsp pages today are generated from markdown files using mkdocs-material and we avoid maintaining raw htmls because of consistency and maintenance, I think we can just fix the link in lein-clojure-lsp to `https://clojure-lsp.io/api/what-is-it/` as I renamed that page recently and forget to check downstream users (my bad for the breaking change). WDYT?

scottbale:
Sounds good! I'll close this PR and open one with that link fix. 👍

scottbale:
I opened clojure-lsp/lein-clojure-lsp#6

PR #1519, opened 2023-03-12
Fix reflection issue in JDK19

borkdude:

### DIFF
--- a/lib/src/clojure_lsp/handlers.clj
+++ b/lib/src/clojure_lsp/handlers.clj
@@ -81,7 +81,7 @@
                                                         (caught-up? db uri desired-version)))
                                               seq)]
               (do
-                (Thread/sleep backoff)
+                (Thread/sleep (long backoff))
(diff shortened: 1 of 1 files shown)


borkdude:
Feedback:

- remove shared
- port reflection test here: https://github.com/borkdude/http-kit/blob/c37f7795a18e88d7eecba0026a7eee77ae51359a/test-native/bb.edn#L15-L18
- add JDK19 to CI

borkdude:
@ericdallo I addressed the feedback. I enabled the `:warn-on-reflection` linter which only warns when JVM interop is performed. I also enabled the `:unsorted-required-namespaces` linter since I only got a warning about this when I did `bb lint`. Having feedback about this earlier is good imo.

borkdude:
Ready for review

ericdallo:
> Having feedback about this earlier is better than discover it when CI fails IMO

I didn't recall of this linter, but makes sense! I'll even bring this to enable at Nubank instead of waiting for clojure-lsp linter complain, thanks!

PR #1534, opened 2023-03-26
Inform client when cancellation is handled

mainej:
When the client cancels a request, the server may or may not observe that notification. That is, it may abort the request, or it may continue processing it. If the server returns a result for a cancelled request, the client [may decide to use it][1]. So, the client needs to know whether the server aborted or handled the request.

Before, when the server aborted a request because of a client cancellation, we were returning an empty response. But to the client, that's ambiguous. It could indicate that the server aborted the request or it could indicate the server is returning an empty result set.

To remove the ambiguity the server now returns a [RequestCancelled][2] error response.

[1]: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/\#implementationConsiderations
[2]: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#errorCodes

### DIFF
--- a/.clj-kondo/funcool/promesa/config.edn
+++ b/.clj-kondo/funcool/promesa/config.edn
@@ -5,4 +5,5 @@
            promesa.core/plet        clojure.core/let
            promesa.core/loop        clojure.core/loop
            promesa.core/recur       clojure.core/recur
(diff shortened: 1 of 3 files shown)


ericdallo (approved):
Thank you!

Issue #1586, opened 2023-06-01
Support JS lib names in completion ns-to-add results and add-require-suggestion command


Deraen:
**Is your feature request related to a problem? Please describe.**

Clojure-lsp can provide require/alias suggestions on completion based on existing requires in the workspace.

For example, if `(:require ["@mui/material/Box$default" :as Box])` is already used, completion on `Box` in another namespace will suggest `Box` alias, with detail field `alias to: @mui/material/Box$default`. But `ns-to-add` data will be `material/Box$default` because(?) the value is coerced to a Clojure symbol at some point?

It is also probable that when the completion item is selected, and the edit is applied (`completionItem/resolve` if I understand correctly), the values are handled as symbols.

Additionally, the same suggestions are available as Code actions, in this case, the code action will have `add-require-suggestion` command the `arguments` has the correct full JS lib require. However, when the action is executed, clojure-lsp will throw error about Invalid symbol.

**Describe the solution you'd like**

If the completion suggestion is created from an existing require form where the namespace is a string, the completion item, and the edit should use the correct (and full) value and keep it as a string.

If I'm using `(:require ["react" :as react])` I'd prefer if `react/` completion also kept using the string as it is common to use strings with all JS libs to separate them from Clojure namespaces.

`add-require-suggestion` command selected from Code actions should work with these JS lib suggestions.

**Describe alternatives you've considered**

-

**Additional context**

Related: https://github.com/clojure-lsp/clojure-lsp/issues/327, but this is just using existing aliases on the project, so there is no need to inspect the JS libs.

ericdallo:
Thanks, I think we never went deep to support cljs requires like that, I suspect we would need some changes in clj-kondo as well, as for kondo analysis `"@mui/material/Box$default" ` is just a string, not a namespace-usage like a clj symbol require.

ericdallo:
Tests like you did in your draft PR is a good start for that, well done

ericdallo:
Fixed by #1587

PR #1587, opened 2023-06-01
Support JS libs in completion and add-require-suggestion

Deraen:
Investigation for #1586

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
   - Fix `suppress-diagnostics` not working on top-level forms when preceded by comment. #1678 
   - Fix add missing import feature on some corner cases for java imports. #1754
   - Fix semantic tokens and other analysis not being applied for project files at root. #1759
+  - Add support for adding missing requires and completion results referring to JS libraries which already have an alias in the project #1587
   
 ## 2024.02.01-11.01.59
 
--- a/lib/src/clojure_lsp/feature/add_missing_libspec.clj
+++ b/lib/src/clojure_lsp/feature/add_missing_libspec.clj
@@ -222,7 +222,9 @@
                            (with-meta class-sym nil)]
 
                           (= :require libspec-type)
-                          (cond-> (with-meta lib-sym nil)
+                          (cond-> (if (string? lib-sym)
+                                    lib-sym
+                                    (with-meta lib-sym nil))
                             (or alias-sym refer-sym) (vector)
                             alias-sym (conj :as (with-meta alias-sym nil))
                             refer-sym (conj :refer [(with-meta refer-sym nil)])))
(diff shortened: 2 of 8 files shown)


ericdallo (commented):
Looks pretty good!

I just noticed the code actions still mention the requires without the quotes, do you think we can make that work too?
image

Deraen:
No problem, just need to check the `js-require` property in code-actions formatting. For these actions the suggestion is already from `find-namespace-suggestions` so it has the property.

Added with a test case now.

Deraen:
Seems like `:refer` completion doesn't work yet:

```
            ["@mui/x-data-grid-pro" :refer [DataGridPro GridActionsCellItem GridToolbar GridLogicOperator]]
            ["react-dropzone-new" :refer [useDropzone]]
```

and either `DataGridPro` or `useDropzone` don't provide autocomplete suggestions of code-actions. But maybe lets make a new issue about that and solve that later?

I guess `refer` code actions are based on knowing the contents of Clojure namespaces, and not the existing refers in the project, so it would require more changes to support this for JS libs. And autocomplete doesn't seem to provide suggestions like these for Clojure either.

Deraen:
While the tests are OK, while using debug-cli binary I noticed I'm getting wrong require forms for non-JS namespaces from add-require code-action.

Deraen:
Okay, I see the problem. Code-action is storing the arguments for `refactor :add-require-suggestion` into arguments option, and that doesn't keep the types. I will probably need to add a new parameter to that call.

ericdallo:
@Deraen I think found a bug testing this, I tried to require `clojure-lsp.logger` require via code action on Emacs and then it was added as: `["clojure-lsp.logger" :as logger]` with quotes

Deraen:
@ericdallo Was that with fix from [19dab26](https://github.com/clojure-lsp/clojure-lsp/pull/1587/commits/19dab26583e58ab1a97439ece53c4c61f6cab15c) because that should have fixed just that? I wasn't able to reproduce the problem e.g. in `lib/src/clojure_lsp/clj_depend.clj` removing the logger require and adding it back using code action.

ericdallo:
I'll double check that, maybe I was in a older commit, can't recall right now, besides that, is this PR ready to review/testing?

Deraen:
Yes, ready for review and testing. I've been using this for over a week now.

ericdallo:
@Deraen I'll test the new version now, but I merged a PR that it seems caused some conflicts. could you resolve the conflicts, please?

Deraen:
@ericdallo Rebased and should be working still based on a quick test.

ericdallo:
@Deraen could you fix the [failing lint task](https://github.com/clojure-lsp/clojure-lsp/actions/runs/7930724582/job/21653561736?pr=1587)?

ericdallo (approved):
Looks great, tested and it works! thanks for the help!

PR #1591, opened 2023-06-06
Fix full qualified namespace

victorpigmeo:

### DIFF
--- a/lib/src/clojure_lsp/kondo.clj
+++ b/lib/src/clojure_lsp/kondo.clj
@@ -81,6 +81,14 @@
          :name-end-row (or (:name-end-row element) (:end-row element))
          :name-end-col (or (:name-end-col element) (:end-col element))))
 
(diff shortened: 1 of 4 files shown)


lib/src/clojure_lsp/kondo.clj
```diff
@@ -129,6 +137,12 @@
                 :name-end-row 0
                 :name-end-col 0))]
 
+    :var-usages
+    [(cond-> element
+       (and (full-qualified-namespace? element)
+            (not (:alias element)))
```
ericdallo, line 143: I'm just a little bit concerned with performance as this function will be called for every var-usages being analyzed, maybe we could check alias first just to filter out some usages.
```suggestion
       (and (not (:alias element))
            (full-qualified-namespace? element))
```

ericdallo (approved):
Looks great, I will just invert the `and` for performance
Thank you!

PR #1623, opened 2023-07-07
Update clj-nix, simplifies nix build

jlesquembre:
I updated `clj-nix` to:
- Add support for Babashka as build tool
- Be more granular about dependencies included in the lock files

The `:nix-flake` alias is no longer necessary. 

Just to build `clojure-lsp`, we don't need the dependencies in the `:debug` alias. I *think* all the other deps.edn files and aliases are needed. If not, you can exclude more dependencies from the lock file with the `--deps-exclude` and/or `--alias-exclude` (you can run `nix run github:jlesquembre/clj-nix#deps-lock -- --help` to see all the options)

I also added a nix shell, which I think is more convenient to run the `deps-lock` command locally and in CI.

### DIFF
--- a/.github/workflows/update_deps-lock.yml
+++ b/.github/workflows/update_deps-lock.yml
@@ -5,6 +5,7 @@ on:
       - master
     paths:
       - "**/deps.edn"
+      - "**/bb.edn"
 
 jobs:
   update-lock:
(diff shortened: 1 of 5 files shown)


ericdallo (approved):
That's awesome! thank you very much again!

PR #1628, opened 2023-07-20
Skip rename of a namespace if it can be found in multiple files

### LINKED ISSUE #1574: File renames changing src files when renaming a test file
**Describe the bug**
When I rename a `test` file, all the imports on my `src` files are changed

**To Reproduce**
1. Clone or open a [project](https://github.com/ericdallo/clojure-sample)
2. Create a file on the `test` directory that has the same `ns` as any `src` file. (ex: `clojure-sample.lsp.call-hierarchy.a`)
3. Rename the file (SPC + f + R on doom-emacs)
4. All the imports of that `ns` will change on `src` files.

**Expected behavior**
Rename only the `test` files if I change a test file.

 - OS: Debian 11
 - Editor emacs (doom-emacs)
 - Version: clojure-lsp 2023.01.26-11.08.16 / clj-kondo 2023.01.20

**Additional context**
It is not a critical bug but could lead to some errors if we don't pay attention.

janosmeszaros:

Fixes  #1574

### DIFF
--- a/lib/src/clojure_lsp/queries.clj
+++ b/lib/src/clojure_lsp/queries.clj
@@ -835,3 +835,11 @@
           (filter #(and (:import %)
                         (safe-equal? class-name (:class %)))))
         (:analysis db)))
+
+(defn find-all-project-namespace-definitions [db namespace]
+  (into []
+        (comp
+          xf-analysis->namespace-definitions
+          (xf-same-name namespace)
+          (medley/distinct-by :uri))
+        (internal-analysis (db-with-ns-analysis db namespace))))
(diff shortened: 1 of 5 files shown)


lib/src/clojure_lsp/queries.clj
```diff
@@ -835,3 +835,11 @@
           (filter #(and (:import %)
                         (safe-equal? class-name (:class %)))))
         (:analysis db)))
+
+(defn find-all-definitions-of-namespace [db namespace]
```
ericdallo, line 839: since this function searchs only in the project and not whole analysis, can't we name it `find-all-project-namespace-definitions`?

janosmeszaros: That's a better name, for sure.
Renamed the fn and added some more tests.

PR #1629, opened 2023-07-20
Update flake.lock, fixes nix darwin build

jlesquembre:
I did a small change to clj-nix, which should fix the build on darwin


stig:
Confirmed that it fixes darwin build for me.

ericdallo (approved):
awesome, thank you very much!

Issue #1637, opened 2023-07-31
Namespaced aliases seems to be ignored


mathieuroblin:
**Describe the bug**
It seems that if an alias defined in `deps.edn` contains a namespace, only the name part of the keyword is picked up in the classpath resolution.

Since there is a bit of setup, please have a look at: https://github.com/mathieuroblin/cljfmt-aliases
This repo contains 3 libs to be loaded differently, two of them are only loaded via an alias and thus are added to the `.lsp/config`.
The project contains two aliases:
1. `:ns/keyword`
2. `:name-keyword`

When starting the LSP (full logs below), it shows that the aliases used for the classpath are: `-A:keyword:name-keyword -Spath`. This is not just a logging issue, the LSP's result shows that all the errors in `lib/3` have been ignored, and I can confirm in various text editors that all the features (like goto) are not working.

*After renaming `:ns/keyword` to `:ns-keyword` everything works as intended.*

**To Reproduce**
Just start the LSP at the root of the provided repo.

**Expected behavior**
The aliases shouldn't be changed by the LSP, the command to resolved the classpath should be: `-A:ns/keyword:name-keyword -Spath`.

<details>
 <summary><b>clojure-lsp output</b></summary>
<pre>
[ 10%] Finding cache               [ERROR] [DB] No cache DB file found
[INFO] [DB] Reading transit analysis cache from /Users/mathieu/Programming/oss/cljlsp-aliases/.lsp/.cache/db.transit.json db took 0ms
[ 15%] Discovering classpath       [INFO] Finding classpath via `/opt/homebrew/bin/clojure -A:keyword:name-keyword -Spath`
[DEBUG] Classpath found, paths:  ["lib2" "lib1" "/Users/mathieu/.m2/repository/org/clojure/clojure/1.11.1/clojure-1.11.1.jar" "/Users/mathieu/.m2/repository/org/clojure/core.specs.alpha/0.2.62/core.specs.alpha-0.2.62.jar" "/Users/mathieu/.m2/repository/org/clojure/spec.alpha/0.3.218/spec.alpha-0.3.218.jar"]
[INFO] [Startup] Using source-paths from classpath: ["/Users/mathieu/Programming/oss/cljlsp-aliases/lib2" "/Users/mathieu/Programming/oss/cljlsp-aliases/lib1"]
[ 20%] Copying kondo configs       [INFO] Copying kondo configs from classpath to project if any...
[...]

PR #1640, opened 2023-08-02
Fix truncation of namespaced keywords

mathieuroblin:
- [X] I created an issue to discuss the problem I am trying to solve or an open issue already exists.
- [X] I updated documentation if applicable (`docs` folder)

Here is a fix to https://github.com/clojure-lsp/clojure-lsp/issues/1637

A few questions remains:
1. I am not sure I provided enough tests, should this be tested at any other level?
2. I tried to stay as faithful to the original implementation as possible. I assumed the orignal use of `name` was to loose the leading `:` so I added an alternative path if the keyword is namespaced but still using the same keyword functions in case there is something else I am not aware of.
3. I never used `lein` before, is there a similar treatment to do? Or the `+` aliases of `lein` follow their own convention?
4. I didn't edit the changelog yet as I am not sure what is the process, I don't think I've seen it in the link you provided. Assuming the commit short message is good enough, should I paste it under `Unreleased` -> `General`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - General
+  - Fix truncation of namespaced keywords #1640
   - Add rewrite-clj node to cursor-info.
   - Fixing semantic-tokens, collons not managed by lsp anymore. #1550
   - Fix `:paths-ignore-regex` setting to consider settings reload.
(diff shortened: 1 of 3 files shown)


mathieuroblin:
I am not 100% sure why the `lint` step is failing, running `bb lint` locally I don't have the errors showing up.

ericdallo:
the lint in CI is failing, I need to fix that somehow

ericdallo:
@mathieuroblin could you add a entry to the CHANGELOG under general section please?

ericdallo (approved):
THank you!

PR #1648, opened 2023-08-13
Use the URI constructor instead of toURI to correctly parse a URL with spaces

Vatyx:
Related: https://github.com/BetterThanTomorrow/calva/issues/2276

When using Calva, I was getting lots of errors coming from the language server. This was happening because it was trying to access a jar file which had a path that had a space in it and it seems that `.toURI` doesn't correctly parse this space. By instead using one of the URI constructors directly, the space is parsed correctly!

This is **literally** my very first piece of real clojure code I've ever written so if the way I've solved this isn't idiomatic or isn't catching some other edge cases please let me know!

### DIFF
--- a/lib/src/clojure_lsp/shared.clj
+++ b/lib/src/clojure_lsp/shared.clj
@@ -206,7 +206,8 @@
   [^String uri]
   (if (string/starts-with? uri "jar:")
     (let [conn (jar-uri-string->jar-url-connection uri)
-          jar-file (uri-obj->filepath ^URI (.toURI ^URL (.getJarFileURL conn)))]
+          url ^URL (.getJarFileURL conn)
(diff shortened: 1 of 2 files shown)


Vatyx:
The windows CL is flaky it seems? https://github.com/clojure-lsp/clojure-lsp/actions/runs/5844975395/job/15848620274?pr=1648

ericdallo:
Thank you for the contribution @Vatyx !
The change looks promising,.but that code already generated lots of bugs in the past and it's a critical code, would you mind create a unit test testing your change?

Vatyx:
Hey @ericdallo, sorry about the delay but finally got around to writing a unit test. What do you think of the one I submitted?

ericdallo (approved):

ericdallo:
Looks good, thank you!

Issue #1653, opened 2023-08-14
`:dependency-scheme` has no effect on Java source files from jars


nbardiuk:
**Describe the bug**
`:dependency-scheme` has no effect on Java source files from jars

**To Reproduce**
1. Given project with `*-sources.jar` in classpath
```clojure
{:deps {org.clojure/clojure$sources {:mvn/version "1.11.1"}
        org.clojure/clojure {:mvn/version "1.11.1"}}}
```
2. And neovim with default clojure-lsp settings (i.e. `:dependency-scheme "zipfile"`)
3. If I try to navigate to the definition of `IPersistentMap` the file path in my editor is `jar:file:///home/user/.m2/repository/org/clojure/clojure/1.11.1/clojure-1.11.1-sources.jar!/clojure/lang/IPersistentMap.java`
4. Although if I try to navigate to the definition of `str` the file path is `zipfile:///home/user/.m2/repository/org/clojure/clojure/1.11.1/clojure-1.11.1.jar::clojure/core.clj`

**Expected behavior**
Both Clojure and Java sources should use the same `zipfile` scheme

**User details:**
 - OS: NixOS
 - Editor neovim
 - Version: 2023.07.01-22.35.41

PR #1655, opened 2023-08-14
Fix `:dependency-scheme` setting for .java files from jars

nbardiuk:

This change fixes the behavior of going to the definition into a java file in a jar using correct `:dependency-scheme` #1653.

Is this a good place to fix the issue?

How can I write a test for this change?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,6 @@
 - General
   - Bump clj-kondo to `2023.09.07`. 
   - Fix move-to-let/expand-let bug for multi-arity functions #1283
-
-- General
   - Fix `:dependency-scheme` setting for .java files from jars #1653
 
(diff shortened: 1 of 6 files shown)


nbardiuk:
@ericdallo, yes I would like to add a test. I will have a look at how to add an integration one.

ericdallo (approved):
Thank you!

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
   - Bump clj-kondo to `2023.09.07`. 
   - Fix move-to-let/expand-let bug for multi-arity functions #1283
 
+- General
+  - Fix `:dependency-scheme` setting for .java files from jars #1653
+
```
ericdallo, line 10:
```suggestion
  - Fix `:dependency-scheme` setting for .java files from jars #1653

```

Issue #1537, opened 2023-03-28
Add Capability:DidChangeConfiguration to allow settings to be changed on the fly


thecontinium:
I am using the [neoconf](https://github.com/folke/neoconf.nvim) plugin in neovim which allows on the fly configuration of lsp servers.

The plugin changes the configuration notifying the server with `"workspace/didChangeConfiguration"` detailed in the specification [here](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeConfiguration).

At the moment clojure-lsp returns with a warning and the configuration settings sent are not changed.

Please see the slack conversation [here](https://clojurians.slack.com/archives/CPABC1H61/p1679990530048619)

Thanks

ericdallo:
We should change [here](https://github.com/clojure-lsp/clojure-lsp/blob/1abd3b51990b11c95e22b27df30adac73eece641/cli/src/clojure_lsp/server.clj#L401) to call `settings` ns and create a function there to update the setting in db.

victorpigmeo:
@ericdallo I'll take a look on this issue 🙂

PR #1658, opened 2023-08-22
Adding functionality to workspace/didChangeConfiguration

victorpigmeo:


victorpigmeo:
Fixes (or at least try to) #1537

docs/settings.md
```diff
@@ -398,3 +398,8 @@ Or to use manual generated stubs:
 ```
 
 clojure-lsp will generate no stubs with that, but analyze that folder and consider it as manual generated stubs.
+
+
+### Changing settings
+
+`clojure-lsp` supports `workspace/didChangeConfiguration` commands to change the server settings while it is running, everything inside the `:settings` map can be changed, but not outside it (the project root for example).
```
ericdallo, line 405: maybe worth mentioning that some settings are read only during startup and won't happen nothing if changed after that

lib/src/clojure_lsp/handlers.clj
```diff
@@ -554,3 +554,9 @@
     components (map :old-uri files)
     :will-rename-files
     (f.file-management/will-rename-files files @db*)))
+
+(defn change-server-settings
```
ericdallo, line 558: WDYT about naming as `did-change-configuration` to follow the name standard?

docs/settings.md
victorpigmeo: nice, do we have a list of these settings?

lib/test/clojure_lsp/feature/change_server_settings_test.clj
```diff
@@ -0,0 +1,120 @@
+(ns clojure-lsp.feature.change-server-settings-test
```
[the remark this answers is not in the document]
victorpigmeo: hmm, makes sence to me, I'll move it

ericdallo (approved):

ericdallo:
@victorpigmeo it seems there is still a failing test because of this PR

victorpigmeo:
yeah, it is failing on windows, sorry I got no time to deal with it yet, but I'll try to fix this evening

ericdallo (approved):
Thank you!

PR #1722, opened 2023-11-10
[install] use dedicated version file for released version

lispyclouds:


install
```diff
@@ -58,7 +58,6 @@ do
             ;;
         *)  # unknown option
             print_help
-            shift
```
lispyclouds, line 61: removed this as `print_help` has an `exit 1` at the end and the `shift` is redundant.

lispyclouds:
not sure if the release script needs to be updated

ericdallo (commented):
@lispyclouds we still need to update that file right? would be nice if we could update it automatically as well in the `release` script

lispyclouds:
yes, hence i wanted to know where and how the `release` file is used/called from.

lispyclouds:
Added it now nonetheless.

ericdallo (approved):

ericdallo:
The release file is manually run when we want a new release, we could do a git tag and leave a GH action do that as well, but it's almost the same, thank you!

Issue #1281, opened 2022-09-29
Hover on local shows information in function call position


snoe:
**Describe the bug**
Hover is showing information for things not under the cursor. Even though parameters and local bindings have little information to show it's better to be told it's a local than showing information for something unrelated that could (in a long function) be completely off screen.

**To Reproduce**
Given the following code and the cursor `|` at these three positions (p1,p2,p3).
```clojure
(def d {})
(defn foo [|a & _]
  (|a)
  (map |a d))
```
At p2 hover shows `a` as expected.

However:
At p1 hover shows the information for `defn`, I would expect `a`
At p3 hover shows the information for `map`, I would expect `a`

**User details (please complete the following information):**
master (f6481e790c9e35105bec6328af82234cbacbdb9d)
and
2022.09.01-15.27.31

ericdallo:
@snoe that was actually intended as `a` is a local, when coding is useful to see the hover docs of the function being called otherwise the current symbol, for example:
```clojure
(foo |a b c)
```

It's useful to see foo docs to check arity and how many args the function has, unless a, b or c are another function call or something, do you have a more concrete example of why that would be better?

mainej:
I’ve never quite understood the difference between `hover` and `signatureHelp`. Is that relevant here?

ericdallo:
signatureHelp was intended to help with that, but most editors and users rely on hover to see function docs/arity/examples, emacs for example, shows the current function on eldoc while typing.

snoe:
@ericdallo it feels really inconsistent and confusing and afaict goes against the purpose of hover.
1. inconsistent:`(map |a d)` if a is a local or var, you get different behavior, so hover to get signature only works sometimes and from a user perspective I think it would be hard to know whether it will or not.
2. crosspurpose: I often lose track of what's a var or a local, hover should be able to tell me that and deducing that "oh I saw the call hover, so that must be a local" is janky.
3. confusing: github image is upload isn't working, so I can't show real world but it's fairly common that hover on a local and I get docs for something that's completely offscreen. In this instance imagine the `if` is above the fold, and you hover on z, it feels very wrong to get a hover for the if.
```clojure
(if x
  {:a a
   :b b
   :c c
   :d d
   ... ...
   :z |z}
   ...)
```

ericdallo:
The idea for that was: why we would want to show locals docs if it's a local? We always try to show the current function being called docs

ericdallo:
[...]

PR #1727, opened 2023-11-24
Show caller for hover documentation

TheBlob42:
Fix for #1281 

I totally agree with @snoe on how the hover functionality should work, but I also see the value in the proposed solution at the end of the linked issue. This is my attempt to implement it. Since it is my first time diving into the clojure-lsp code there might be conventions or other things that I have not done "correctly". Please feel free to point everything out that you think could be solved in a better way (also when there are naming conventions etc.).

I would check the CHANGELOG.md and documentation once we have agreed on the code changes in general

    - already existed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 - General
   - Fix deep-merge of client settings with project settings.
   - Fix `max-line-length` on clean-ns feature not respecting some lines when contains a lot of refers. #1729
+  
+- Editor
+  - Fix edn tree visualization for keys with same name in the same level. #1558
 
 - API/CLI
   - Drastically improves performance of `format` task matching `cljfmt`, avoiding analyzing the project. #1723
--- a/lib/src/clojure_lsp/feature/document_symbol.clj
+++ b/lib/src/clojure_lsp/feature/document_symbol.clj
@@ -59,7 +59,7 @@
         (rest coll)
         (cons (first coll) (lazy-seq (remove-first item-to-remove (rest coll))))))))
 
-(defn ^:private edn->element-tree [m keyword-elements symbol-elements]
+(defn ^:private edn->element-tree [m keyword-elements* symbol-elements*]
   ;; TODO use tail recur for better performance
   (when (coll? m)
     (->> m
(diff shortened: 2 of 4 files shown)


TheBlob42:
I have another idea regarding the correct syntax highlighting of the calling-line:
Could we wrap it inside an annotated  code block? At least for the markdown case

image

We would add another newline since we can't inline a code block with syntax highlighting (to my knowledge), but we would get the correct highlighting for the function and arguments. The additional code block lines should be handled by the editor (Neovim is working on it, I guess VSCode already does this). What do you think @ericdallo ?

ericdallo:
@TheBlob42 nice, the markdown idea is what I was expecting, but we should keep it at the first line, some editors like emacs has a hover preview feature which shows the first line of hover, so in this case would just show `calling:` which is not useful, maybe we should move `calling:` to the same line inside the code block or as a comment in the end of the first line or so, WDYT?

image

TheBlob42:
I did not know about this Emacs feature (quite cool actually). Therefore your proposal of putting everything into the same line makes a lot of sense. I personally would prefer to put the `calling: ` element up front, to make it clear that this documentation is not for the element directly under the cursor. Putting it at the end of the line, would increase the "risk" for it to be overlooked or in cases of long function definitions even cut off.

I would love your input on how to insert it into the hover text. We could simply insert it (it should then be highlighted like a symbol), as a string (for some different highlighting) or my personal favorite with the `#_`  [macro](https://clojure.org/reference/reader#_dispatch) which should be highlighted as a comment and is the closest thing to an inline comment in Clojure. Or we use something else. WDYT?

image

image

image

ericdallo:
I liked the `#_` one, I hope it will be clear to users, it's already better to what we have ATM which is nothing, so I think we can go with that option

TheBlob42:
Sorry my rebasing auto closed this PR and I can't reopen it. I opened a new PR instead here #1735

TheBlob42:
Reopen this PR

TheBlob42:
Sorry my rebasing auto closed this PR and I don't understand how to reopen it :joy:
Please check the new PR with the latest state here #1735

PR #1735, opened 2023-12-11
Show caller for hover documentation

TheBlob42:
Sorry my rebasing efforts auto closed the original PR #1727 
For the full discussion check the original PR. @ericdallo please have another look

### DIFF
--- a/lib/src/clojure_lsp/feature/hover.clj
+++ b/lib/src/clojure_lsp/feature/hover.clj
@@ -119,10 +119,32 @@
                (string/join ", ")
                (format "(%s)"))))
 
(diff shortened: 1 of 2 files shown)


TheBlob42:
Is there a way to re-run the checks or is this restricted to only the maintainers?
Seems like there are sometimes some network issues that block the checks from succeeding :thinking:

ericdallo (approved):

ericdallo:
> Is there a way to re-run the checks or is this restricted to only the maintainers?

I guess so, but couldn't find a setting to enable for outsiders

Anyway, looks good, thank you for the help!

Issue #1754, opened 2024-01-15
Java imports code action breaks namespace form


armed:
**Describe the bug**
In some cases import java package code action breaks `:import` section of a ns form. I believe it happens when a new package's path matches to one which already imported. E.g. `java.awt.event` and `java.awt.<something>`.

**To Reproduce**
1. assume following ns declaration:
```clojure
(ns my.namespace
  (:import [java.awt.event ActionEvent]))
  
;; ActionEvent is used somewhere in the code
```
2. Add a new java class usage `Robot.`
```clojure
(def robot (Robot.))
```
3. Execute code action which suggests to add `java.awt.Robot` to imports declaration
4. Form becomes broken:
```clojure
(ns my.namespace
  (:import [[java.awt Robot event] ActionEvent]))
```

**Expected behavior**
A form should look like this:
```clojure
(ns my.namespace
  (:import [java.awt.event ActionEvent]
           [java.awt Robot]))
```

Issue #1316, opened 2022-10-09
Add missing requires from `comment` should not add to ns


mainej:
**Is your feature request related to a problem? Please describe.**

Currently `add-missing-require` always adds the missing namespace to the `ns`'s `:require` form. But when in an RCF, a top-level `(comment ...)` form, this isn't what some users want. They feel that this clutters their `ns` with namespaces that aren't really used. They'd prefer that the missing namespace be put in a `(require ...)` form in the RCF.

**Describe the solution you'd like**

I'd like one of two things. I think the best would be a configuration option that modified `add-missing-require`'s behavior. Bonus points for adding the missing namespace to an existing `(require ...)`. Alternatively, there could be a new command.

**Describe alternatives you've considered**
It's possible to add `(require ...)` by hand.

**Additional context**
See discussion in [Slack](https://clojurians.slack.com/archives/CPABC1H61/p1664877429715259)

ericdallo:
I see 2 good possible options:
- Add support as a flag, we can discuss the default value.
- Always ask when adding a require inside a RCF if user wants to add on comment or top-level ns

mainej:
Yeah, either of those would work. Prompting the user is a nice idea. This is a rare enough use-case that a prompt wouldn't be too intrusive, IMO.

blnote:
Hi, I'm working on this a little: right now I've introduced a flag `add-require-inside-comments` and  I'm showing a prompt whenever the code-action `add-missing-libspec` is called within a comment form and this flag is not set (neither true or false). 
However, `code-completion` adds lib-specs too and asking might be difficult in this case, since the edit seems to be done before the user commits to a completion (?)...
so maybe I'd just stick to the setting of the flag in this case and default to the old behavior if it is not present ? ...just writing down thoughts :)

ericdallo:
Fixed by #1767

PR #1767, opened 2024-02-04
Added setting to allow requires/imports to be added within the current comment form during code action and completion. [#1316]

blnote:

Added setting to allow requires and imports to be added within the current comment form during code action and completion. [#1316](https://github.com/clojure-lsp/clojure-lsp/issues/1316)

 `:completion {:add-libs-to-rcf [true|false|nil]}`

- `false`: everything works as before
- `nil` or missing: when a code action for adding a missing lib is executed within a comment form, the user will be asked where the form shall be included
- `true`:  `imports` and `requires` from within a comment form are always added to the top of the form.
 
### Discussion

I like the enhancement, because it makes it possible to eval the require without scrolling to the top of the namespace. However,  this is just a suggestion :)
I started with a simpler refactoring separately form the existing code, but ended up duplicating the pattern-matching logic of `add-missing-libspec/add-to-namespace*`, so I extracted the shared part into two separate functions. (did not change the logic, but lots of added lines because of it) Certainly open for discussion.

### Notes on code changes
- added `edit/inside-rcf?` to test if we are in comments
- `add-missing-libspec/add-to-namespace*` was already quite long and the comment-case differed in a few points (e.g. modifying the comment form while the cursor is inside caused conflicts with completion when the whole edited form is returned) so I decided to create a separate function `add-to-rcf*`. which of the two function is chosen depends on whether the zloc-argument points to a comment form.
- since cursor position had been lost in completions with delayed resolving, I modified `completion.clj`: the position of the comment form is extracted in `completion-item-with-unresolved-alias-edit` and passed along as `rcf-pos`, the comment form is then recovered in `completion-item-with-alias-edit` for resolving


lib/src/clojure_lsp/feature/add_missing_libspec.clj
```diff
@@ -175,61 +224,112 @@
...
+    (when (and
+            (not (= false? c-setting))
+            rcf-zloc
+            (or c-setting
+                (and
+                  producer
+                  (= "Yes"
+                     (producer/show-message-request
+                       producer
+                       "Add require inside this comment form? (to avoid this prompt, set { :completion {:add-libs-inside-rcf true|false }.)"
```
ericdallo, line 257: I'd leave this as:
```suggestion
                       "Add require inside this comment form?"
```

And if anyone complains we can point to the docs

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 - General
   - Bump clj-kondo to `2023.12.16-20240202.114914-15`.
 
+- Editor
+  - Added setting to allow requires and imports to be added within the current comment form during code action and completion: `:completion {:add-libs-to-rcf [true|false|nil]}` #1316
```
ericdallo, line 9: Please remember to update settings.md + docs/all-available-settings.md with the default value

lib/src/clojure_lsp/feature/add_missing_libspec.clj
blnote: thanks for your quick response - sure, that's fine with me, I'll update code & docs (forgot to push them yesterday...)
I wrote the prompt before adding the functionality for imports, maybe we should ask "Add lib inside this comment form?" instead?

blnote: Nevermind, I can just parametrize it and ask for require and import respectively.

blnote:
While testing I thought I created a problem, but turned out it was issue [1754](https://github.com/clojure-lsp/clojure-lsp/issues/1754). The problem seems to be in `add_missing_libspec/find-when-starts-with-sym` .
A quick fix could be to not only look if the existing libs **starts with** the lib-to-add, but also check for **equality**  whenever the existing lib-import is in a list/vector.
this fixes the test-case and all tests run through, but I'm not sure... could be that to make it more robust, one would have to look at **all** occurrences of strings in ns of which the lib-to-add is a prefix and choose a suitable one. Don't know if you worked on this already?

ericdallo:
@blnote thanks, looks good, I'll do some manual tests, could you do a one last solve conflict please?

Also, I intend to fix #1754 soon

ericdallo (approved):

ericdallo (approved):
Thank you!

PR #1806, opened 2024-04-18
fix: uri->source-paths when one source-path is the prefix of another

### LINKED ISSUE #1805: rename ns picks wrong source-path
**Describe the bug**

Performing a rename of namespace `foo.bar` (living in `src/cljs/foo/bar.clj`) to `foo.quux` results in file `src/clj/foo/quux.clj`. The source-path changes from `src/cljs` to `src/clj`.

**To Reproduce**

- `M-x lsp-rename`
- `foo.bar` to `foo.quux`

**Expected behavior**

I expect the renamed ns to stay within the same source-path.

 - OS: MacOs
 - Editor: Emacs
 - Version: 2024.03.31-19.15.11-nightly

**Additional context**

We think the bug is in `filename->source-paths`. It uses `string/starts-with?` which gives a false positive on `src/clj` because `src/cljs/...` indeed starts with `src/clj`!

https://github.com/clojure-lsp/clojure-lsp/blob/a922507de86257b4dc0976fd05049fed5a08c158/lib/src/clojure_lsp/shared.clj#L254

To test this theory we changed the project config to have `src/aljs` and `src/clj`. After that the rename uses the right source-path.

We're willing to make a PR.

opqdonut:
fixes #1805


opqdonut:
Sorry for the churn! Had to iterate a bit in order to not break existing use cases & to get tests to pass on all platforms. Ready for review now.

ericdallo (approved):
Looks good! thank you for your contribution!

PR #1825, opened 2024-05-22
Fix call to ns-groups when :filename-pattern in use

SevereOverfl0w:

### DIFF
--- a/lib/src/clojure_lsp/feature/diagnostics.clj
+++ b/lib/src/clojure_lsp/feature/diagnostics.clj
@@ -43,7 +43,7 @@
 (defn ^:private unused-public-var->finding [element kondo-config]
   (let [keyword-def? (identical? :keyword-definitions (:bucket element))
         kondo-config (if (:ns element)
-                       (kondo-config-for-ns kondo-config (:ns element) (:filename element))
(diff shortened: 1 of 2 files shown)


SevereOverfl0w:
I don't understand what & why the macos integration test is timing out, is it known to be flaky?  I've pushed up a commit using a dynamic var & memoization (as I saw elsewhere in the file) to see if it will speed up the build sufficient to get it through.

lib/src/clojure_lsp/feature/diagnostics.clj
```diff
@@ -16,6 +16,8 @@
 
 (set! *warn-on-reflection* true)
 
+(def ^:dynamic *uri->filename* nil)
```
ericdallo, line 19: I'm not sure it's a good idea to start using dynamic vars in the project like that, I think passing as args would be best, otherwise, we would start introducing states between namespaces and that would be dangerous

SevereOverfl0w: I'll remove this commit, it was an attempt at blind performance optimization before I realized that the mac integration test is failing for everyone!

SevereOverfl0w: Dropped the commit

ericdallo (approved):
Thank you!

PR #1864, opened 2024-08-10
Fix move-form problem with cljc files.

danjohansson:
Some edits where made twice causing the second edit to fail.

When usages of the moved form where detected in cljc files 2 usages, one for each language, where found for the same position. The solution filters out one of them to avoid edit the file twice.

Added two tests, that are copies of the previous two but with cljc files instead. There is some duplication there now because of this. I can clean that up further if you wish?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- General
+  - Fix move-form double edit problem in cljc files.
+
 ## 2024.08.05-18.16.00
 
 - General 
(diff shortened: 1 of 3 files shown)


ericdallo (approved):
Looks good, I think the test is ok for now, thank you!

ericdallo:
I intend to work to improve that feature soon for other cases as well

danjohansson:
Ok thanks for letting me know. :+1: 
I have one more issue I have been trying to fix today, that I just made a PR for. 
Hope that is ok.

Issue #1448, opened 2023-01-12
Better error message when client doesn't support Creation WorkspaceEdit


bo-tato:
**Describe the bug**
"Create test" code action fails if I haven't already created the file `test/<namespace>/<filename>_test.clj`
And (I'm using neovim) I just get a unclear error message:
```
Error executing vim.schedule lua callback: /usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:409: method "workspace/applyEdit":
...
```
I'm not sure if this is an issue in neovim's lsp handling, or in clojure-lsp, looking at the lsp log it seems that clojure-lsp tells it to do "workspace/applyEdit" on the _test.clj file, which doesn't exist so vim errors. 

**To Reproduce**
1. Open any clojure file in neovim with lsp setup
2. Run "code actions"
3. Pick option "create test for..."

**Expected behavior**
Maybe clojure-lsp or neovim should be creating the file if it doesn't exist, or at least returning a clear error message about what file is missing that we need to create

<details>
 <summary><b>Log - client <-> server</b></summary>
<pre>
[DEBUG][2023-01-12 11:39:37] .../lua/vim/lsp.lua:1381	"LSP[clojure_lsp]"	"client.request"	1	"workspace/executeCommand"	{  arguments = { "file:///aoc/protohackers/src/protohackers/challenge2.clj", 9, 7 },  command = "create-test"}	<function 1>	1
[DEBUG][2023-01-12 11:39:37] .../vim/lsp/rpc.lua:285	"rpc.send"	{  id = 96,  jsonrpc = "2.0",  method = "workspace/executeCommand",  params = {    arguments = { "file:///aoc/protohackers/src/protohackers/challenge2.clj", 9, 7 },    command = "create-test"  }}
[DEBUG][2023-01-12 11:39:37] .../vim/lsp/rpc.lua:388	"rpc.receive"	{  id = 96,  jsonrpc = "2.0"}
[DEBUG][2023-01-12 11:39:37] .../vim/lsp/rpc.lua:388	"rpc.receive"	{  id = 8,  jsonrpc = "2.0",  method = "workspace/applyEdit"}
[DEBUG][2023-01-12 11:39:37] .../vim/lsp/rpc.lua:403	"server_request: callback result"	{  status = true}
[...]

PR #1907, opened 2024-10-31
Fix client capability check that prevents test file creation in Neovim

zhming0:
Related to #1448, this PR fixes a problem where `clojure-lsp` incorrectly check client capabilities that prevent test file creation in Neovim.

According to LSP Spec, for `WorkspaceEdit` ([link](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspaceEdit)): 

> If a client neither supports `documentChanges` nor `workspace.workspaceEdit.resourceOperations` then only plain `TextEdit`s using the `changes` property are supported.

However, Clojure LSP was only checking against `workspace.workspaceEdit.documentChanges`. So if client only set `workspace.workspaceEdit.resourceOperations` (like Neovim LSP client by default),  when creating a new test file, `documentChanges` gets incorrectly passed into `changes` -> LSP client then report `clojure-lsp` violate LSP spec.

I can confirm after the fix, create new test file through code action works. 

--- 

More details, not quite matter to this PR.

I feel the official spec looks pretty confusing, but hopefully I understand it correctly: 

* When `documentChanges = true` and `resourceOperations = nil`, client expects `documentChanges: TextDocumentEdit[]`. 
* When `documentChanges = true` or `resourceOperations` is [action], client expects `documentChanges: TextDocumentEdit[] | action`. 
* When `documentChange = false` and `resourceOperations = false`, `documentChanges` is not permitted.


ericdallo:
Thank you!

Issue #1920, opened 2024-11-26
I would like simple keyword completion to be based on all keywords in the project.


mitchelkuijpers:
**Is your feature request related to a problem? Please describe.**
 I am coming from cursive where I can basically type `:` and then it already starts autocompleting all simple `keywords` from the project + libraries. I got really used this behavior (this prevents a lot of typos).

**Describe the solution you'd like**
I would simple keywords (not aliased or namespaced) autocompletion for the whole project + libraries.

**Describe alternatives you've considered**
I currently fixed this by in emacs by combining cursive autocompletion with clojure-lsp completion with `cape-capf-super`.

**Additional context**
Small snippet from slack:

> **Mitchel Kuijpers**
Just to make sure I understand correctly if I have a file somewhere in my project with the keyword :foo should I expect to get autocompletion for :foo  if I am in a different file in the same project? Because I just tried nvim with LSP and I notice the same behavior I cannot easily get autocompletion for all keywords in the project (it only works if they are in the same file).
With cider I do get all keywords so I currently combine the autcompletion with clojure-lsp but that gives me a lot of double results when autocompleting. I am coming from cursive where I can basically type : and then it already starts autcompleting all keywords from the project + libraries.

> **Mitchel Kuijpers**
Am I understanding this correctly 
[@ericdallo](https://clojurians.slack.com/team/UKFSJSM38)
?

> **Eric Dallo**
Yes, I can't recall why in the past we decided that we would only provide simple keywords (not aliased or namespaced) for the same namespace ([ref](https://github.com/clojure-lsp/clojure-lsp/blob/1e0a34ec65f6a8d92492a15c436c2a43325f06aa/lib/src/clojure_lsp/feature/completion.clj#L472-L473)), that's how it works currently, but I can see how convenient should be provide all, feel free to create a issue about it.
About the cljs completion, I think it's a bug, feel free to create another one about it

mitchelkuijpers:
Btw I would be happy to pick this up, should I just take a stab at it and create a PR?

ericdallo:
@mitchelkuijpers I believe this was for performance reasons, but I can't see where this would cause performance reasons taking a quick look, feel free to try open a PR providing that with tests.
[...]

PR #1921, opened 2024-11-26
Remove restriction on simple keyword completion

mitchelkuijpers:
Removed filter that would only complete simple keywords in the current namespace.

- [x] I created an issue to discuss the problem I am trying to solve or an open issue already exists. #1920 


ericdallo (approved):
Looks good, thanks, will keep an eye if that causes any performance problem or something we didn't mention anywhere.

PR #1934, opened 2025-01-11
Add textEdit to completion items

### LINKED ISSUE #1933: Completions don't respect Clojure word boundaries in Zed
**Describe the bug**
I've been experimenting with Clojure in [Zed](https://zed.dev). Mostly it works pretty well out of the box. Completions don't seem to respect Clojure word boundaries, however. See screenshots below

**Screenshots**

https://github.com/user-attachments/assets/251d7d74-40dc-4ad0-a974-a95eb02f179f

https://github.com/user-attachments/assets/ccede837-25c3-4e5c-b583-b1aa57843be9

**Additional context**

https://github.com/zed-extensions/clojure/issues/2

paulbutcher:

Fixes #1933 and https://github.com/zed-extensions/clojure/issues/2

The bug arises from the fact that clojure-lsp is returning `CompletionItem`s without a text edit entry. As [the specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem) says:

> The `insertText` is subject to interpretation by the client side.
	 * Some tools might not take the string literally. For example
	 * VS Code when code complete is requested in this example
	 * `con<cursor position>` and a completion item with an `insertText` of
	 * `console` is provided it will only insert `sole`. Therefore it is
	 * recommended to use `textEdit` instead since it avoids additional client
	 * side interpretation.

Here's the behaviour in Zed after this change:

https://github.com/user-attachments/assets/2383a64f-c5fa-4130-af0e-a4399e01e32f

https://github.com/user-attachments/assets/7a618787-9916-4d03-aed3-4c3f5b1421bd

I've tested in VSCode and (as far as I can see) everything is still working as expected. I believe it should do so, as this change conforms to the Language Server Protocol specification.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 - Editor
   - Change simple keyword completion to return all known keywords. #1920
+  - Return textEdit to CompletionItems to fix completion in Zed #1933
 
 ## 2024.11.08-17.49.29
 
--- a/lib/src/clojure_lsp/feature/completion.clj
+++ b/lib/src/clojure_lsp/feature/completion.clj
@@ -599,6 +599,21 @@
                        [(:score %1) (:label %2) (:detail %2)]))
        not-empty))
 
+(defn ^:private completion-range [row col cursor-value cursor-element]
+  (or
+    (shared/->range cursor-element)
+    (shared/->range {:row row
+                     :col (- col (count (str cursor-value)))
+                     :end-row row
+                     :end-col col})))
+
+(defn ^:private add-text-edit [range items]
+  (map (fn [item]
+         (merge {:text-edit {:new-text (:label item)
+                             :range range}}
+                item))
+       items))
+
 (defn completion [uri row col db]
   (let [root-zloc (parser/safe-zloc-of-file db uri)
         ;; (dec col) because we're completing what's behind the cursor
@@ -710,7 +725,8 @@
              sorting-and-distincting-items
              ;; Limit the returned items for better performance.
              ;; If user needs more items one should be more specific in the completion query.
-             (take 600))))))
+             (take 600)
+             (add-text-edit (completion-range row col cursor-value cursor-element)))))))
 
 ;;;; Resolve Completion Item (completionItem/resolve)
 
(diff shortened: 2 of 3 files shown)


paulbutcher:
I can see that the Windows and MacOS tests are failing, but I'm not sure why. The failures don't seem to be relevant to any of the changes I've made in this PR so I'm marking it as "Ready for Review". If I'm missing something, please let me know.

Scutua (approved):

ericdallo:
@paulbutcher thanks for the contribution, the changes looks good and I think I never noticed this problem because all editors had a good enough implementation for completing on the client side.
One concern though is, completion is a really performance important feature, and I noticed that with this PR we increased in 40-50% the json size of completion since we now are returning a new field with multiple nested fields, I'm wondering if we should return that only in the `completionItem/resolve` request which should be triggered when user hover over a item or select it, WDYT?

this way for 100 items, we would not include this field for each item, but only to the selected one

paulbutcher:
That's a good point. I did try to use `itemDefaults` (this is exactly the reason they were added to the spec), but they're not supported by Zed (which is the primary motivation for the change).

I'll see if I can move this into `completionItem/resolve` 👍

paulbutcher:
@ericdallo I've looked and I'm not sure it will help? As I understand it, `completionItem/resolve` just gets given the `CompletionItem` returned by `textDocument/completion`. Right now that doesn't include the information it would need to create the `textEdit` (specifically `textDocument` and `position`). I could add these to the `data` parameter of each `CompletionItem` but this would increase the size of each by approximately the same proportion as adding `textEdit` directly.

Am I missing something?

NoahTheDuke:
> but they're not supported by Zed (which is the primary motivation for the change).

Isn't this a bug with Zed and not with clojure-lsp?

ericdallo:
@paulbutcher I believe we could add something like `"range": "5:3,5:10"` to specify the range in a smaller way, and translate it in the resolve request, but not sure that worth it, it'd be definitly smaller in size in the completion json, WDYT?

ericdallo:
Yep, that's the golden question hehe
There is a issue to add performance tests to clojure-lsp which would help know how bad it is, for now, I'd be satisfied with a few manual tests, like in zed or vscode measure the time the request took vs master, and if you notice any visual delay

paulbutcher:
Fair enough. I'm happy to be guided by you (you have much more context than I do!). Let me know if you would like me to investigate shifting this to `completionItem/resolve` or whether you would prefer to stick with it as it currently stands.

ericdallo:
@paulbutcher I did some manual tests in emacs and vscode and didn't notice any difference, I think it's pretty ok to merge :)

PR #1947, opened 2025-01-14
Linked edit fixes

### LINKED ISSUE #1943: Restrict linked edits to namespace aliases only
As discussed in #1936

paulbutcher:

Fixes #1943

This is a work in progress: it makes things better than they currently are, but definitely still isn't a complete solution.

This PR:

- Restricts linked edits to only work for namespace aliases (as per #1936).
- Fixes the [error reported by VSCode/Calva](https://github.com/BetterThanTomorrow/calva/issues/1374#issuecomment-2586621458) by returning `nil` instead of an error when there are no applicable linked edits.
- Fixes the [bug in VSCode/Calva](https://github.com/BetterThanTomorrow/calva/issues/1374#issuecomment-2587256089) in which the entire symbol (not just the alias portion) is modified

However it still doesn't work entirely reliably (the timing-related issue described [here](https://github.com/BetterThanTomorrow/calva/issues/1374#issuecomment-2587256089) is still evident). I'm also not really happy with the fact that I've had to cobble together `alias-range` instead of using `->range` (as far as I can see there's no other way to dig the alias portion out of a symbol?).

I'll see if I can work out what's going on with the timing issue, but I'm still very unsure whether the relatively little value added by linked edits for Clojure is worth the trouble? 🤷

### DIFF
--- a/.github/workflows/bump_downstreams.yml
+++ b/.github/workflows/bump_downstreams.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Prepare java
         uses: actions/setup-java@v1
         with:
-          java-version: 1.8
+          java-version: 11
 
       - name: Install Clojure
         uses: DeLaGuardo/setup-clojure@master
@@ -40,7 +40,7 @@ jobs:
       - name: Prepare java
         uses: actions/setup-java@v1
         with:
-          java-version: 1.8
+          java-version: 11
 
       - name: Install Clojure
         uses: DeLaGuardo/setup-clojure@master
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,12 +51,12 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04]
-        jdk: [8, 11, 15, 17, 19]
+        jdk: [11, 15, 17, 19]
         include:
           - os: windows-latest
-            jdk: 8
+            jdk: 11
           - os: macos-latest
-            jdk: 8
+            jdk: 11
     steps:
       - uses: actions/checkout@v2.2.0
         with:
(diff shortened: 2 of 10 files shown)


ericdallo:
> (as far as I can see there's no other way to dig the alias portion out of a symbol?).

Unfortunately, yes

> I'll see if I can work out what's going on with the timing issue, but I'm still very unsure whether the relatively little value added by linked edits for Clojure is worth the trouble? 🤷

I think the problem is the analysis we lost when we change the buffer, when we edit anything, clojure-lsp sends to clj-kondo the whole text to get new analysis, changing its internal db (atom), I believe there is some issue while the analysis was not saved to the db, but that usually means the request are in some weird order

paulbutcher:
> I think the problem is the analysis we lost when we change the buffer, when we edit anything, clojure-lsp sends to clj-kondo the whole text to get new analysis, changing its internal db (atom), I believe there is some issue while the analysis was not saved to the db, but that usually means the request are in some weird order

That makes sense: I can see that `find-element-under-cursor` is sometimes returning `nil` if it's called "soon" after an edit. What can I do about this?

paulbutcher:
@ericdallo I've pushed changes which should fix the integration tests.

In the process, I discovered that `assert-submap` is sensitive to the order of the maps, which surprised me (and feels a bit fragile?). But anyway, I think this should now be good to go.

ericdallo:
> That makes sense: I can see that find-element-under-cursor is sometimes returning nil if it's called "soon" after an edit. What can I do about this?

@paulbutcher I think I have the fix for this, we need to wrap in handlers.clj the linkedEdit call with `process-after-all-changes`, this may fix the concurrence issue, could you try?

ericdallo (approved):
Looks good, thank you!

PR #1948, opened 2025-01-15
Add process-after-changes to linked edits

paulbutcher:

@ericdallo as discussed.

This makes things *much* better in both VSCode and Zed, but:

* I still see errors in VSCode if I type quickly (nowhere near as many as before this change, but it still happens pretty frequently with my "normal" typing speed.
* It seems to be entirely reliable in Zed (no matter how fast I type). There is a Zed-specific error with hyphens `-` which I've already [reported](https://github.com/zed-industries/zed/issues/23063).

### DIFF
--- a/lib/src/clojure_lsp/handlers.clj
+++ b/lib/src/clojure_lsp/handlers.clj
@@ -468,8 +468,9 @@
       (f.call-hierarchy/outgoing uri row col components))))
 
 (defn linked-editing-ranges
-  [{:keys [db*]} {:keys [text-document position]}]
-  (shared/logging-task
+  [{:keys [db*] :as components} {:keys [text-document position]}]
+  (process-after-changes
+    components (:uri text-document)
     :linked-editing-range
     (let [db @db*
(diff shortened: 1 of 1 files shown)


paulbutcher:
@ericdallo I'm not sure I understand why unit tests are failing? They don't seem to be related to the changes in this PR?

ericdallo:
I'm working on fixing the unit-tests right now

ericdallo (approved):
LGTM, not sure why on vscode still happens, especially weird if happens only there

paulbutcher:
@ericdallo I don't know why I didn't think to try this before, but the linked editing issue with fast typing isn't Clojure specific (I can reproduce exactly the same problem when using linked edits in HTML as well). So I think we can chalk this up as a VSCode bug.

(I was looking at it because I was worried that it suggested we had a concurrency issue, which might be relevant to the completion issues I'm currently looking at, but this seems to be a red herring).

paulbutcher:
For completeness: https://github.com/microsoft/vscode/issues/180156

PR #1955, opened 2025-01-24
Add :hover :hide-signature-call setting

### LINKED ISSUE #1954: Disable #_calling: in hover
As implemented after discussion in #1281, hovering a local shows a line indicating the surrounding calling context:

```clojure
(defn fo|o-bar [x]
  (+ x x))
```

Requesting hover will show:

```
#_calling: clojure.core/defn [name doc-string? attr-map? [params*] prepost-map? body] [name doc-string? attr-map? ([params*] prepost-map? body) + attr-map?]
──────────────────────────────────────────────────────────────────────────────
user/foo-bar
[x]
```

I don't want or need the first line of information. It is distracting and noisy. If I needed to see the signature of the surrounding call, I'd move to it and look at it, or I'd trigger signatureHelp.

Add a config option to not show this.

1) Do nothing.
2) change it somehow?

**Additional context**
I use neovim so I can get hover without editing text, which maybe gives me a different perspective than in other editors where there's no difference. But for me it's very distinct: when I am moving around and reading code, I want only the documentation for a given variable. When I'm editing code, I want the signatureHelp. There's no need for overlap.

NoahTheDuke:
Defaults to false. If true, doesn't include `#_calling: user/foo [x y z]` at the start of hover text.

Closes #1954


ericdallo (approved):
Looks good, please fix lint with `bb lint-fix`

PR #1960, opened 2025-01-31
Add textEdit to completion items (second attempt)

paulbutcher:

This is a second attempt at #1934. The problems caused by that PR resulted from the fact that `cursor-element` might be stale (if running with `:fast-but-stale`). So instead, we derive it from `cursor-loc` (which is never stale).

Another difference from the previous attempt is that we only add `textEdit` if we've been able to parse the file. This means that we fall back on the client's behaviour if the file is syntactically incorrect. Not ideal, but better than giving it incorrect `textEdit` data.


ericdallo (approved):
The changes look good, let's wait for user test feedback

Issue #2017, opened 2025-04-14
Code action suggests incorrect :refer for local unresolved function instead of creating it


afucher:
When calling a local function that is defined later in the same namespace, clojure-lsp correctly reports it as unresolved. However, the first code action suggested is to add a `:refer` to the current namespace, as if the function was external. This suggestion is incorrect and breaks the ns form by transforming it into `(ns [baz :refer [func-1]])`. The correct suggestion — creating the missing function — is available, but it appears second in the list.

**To Reproduce**
1. Create a file with the following code:

```clojure
(ns baz)

(defn func-2 []
  (func-1))

(defn func-1 [] ())
```

1. Place the cursor on func-1 in the body of func-2.

1. Trigger the code actions (lightbulb or equivalent).

1. Notice that the first suggestion is to add a :refer to baz.

**Expected behavior**
The first suggestion should be to create a new function func-1, since the unresolved symbol is a local function defined later in the file. Referencing the current namespace with :refer is both semantically incorrect and syntactically invalid.

**User details (please complete the following information):**

OS: MacOS
Editor VSCode (Calva), Intellij (clojure-lsp-intellij)
Version: 2025.03.27-20.21.36

Additional context
The suggestion to :refer the current namespace does not make sense in this context, and could lead to confusion or broken code, especially for beginners. Ideally, clojure-lsp would prioritize creating the local function in these scenarios.

PR #2018, opened 2025-04-16
Fix code suggestion for unresolved-symbol

acamargo:
https://github.com/clojure-lsp/clojure-lsp/issues/2017

### DIFF
--- a/lib/test/clojure_lsp/feature/code_actions_test.clj
+++ b/lib/test/clojure_lsp/feature/code_actions_test.clj
@@ -35,6 +35,33 @@
                                  {:workspace {:workspace-edit true}}
                                  (h/db))))))
 
+;; {:title "Add require '[baz :refer [func-1]]'",
+;;  :kind :quick-fix,
+;;  :is-preferred true,
+;;  :command
+;;  {:title "Add require suggestion",
+;;   :command "add-require-suggestion",
+;;   :arguments ["file:///a.clj" 2 3 "baz" nil "func-1" nil]}}
+
+(deftest avoid-self-refer-suggestion
+  (h/load-code-and-locs (h/code "(ns baz)"
+                                "(defn func-2 []"
+                                "  (func-1))"
+                                "(defn func-1 [] ())"))
+  (testing "Given a unresolved-var because the function is declared after its use
+            When I ask for code actions
+            Then I don't get an 'Add require' suggestion"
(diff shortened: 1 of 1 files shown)


deps.edn
```diff
@@ -5,7 +5,8 @@
                               clojure-lsp/cli {:local/root "cli"}
                               io.github.clojure/tools.build {:git/url "https://github.com/clojure/tools.build.git"
                                                              :tag "v0.10.7"
-                                                             :sha "573711e"}}}
+                                                             :sha "573711e"}
+                              nubank/matcher-combinators {:mvn/version "3.9.1"}}}
```
acamargo, line 9: I'm not sure if this is the best place to declare this dependency.

ericdallo: no, this is a dev alias, it should be included in a test alias, and probably on lib/deps.edn as it's a code thing

lib/src/clojure_lsp/feature/add_missing_libspec.clj
```diff
@@ -584,6 +590,7 @@
                             [{:ns (name common-refer)
                               :refer cursor-name-str}]
                             (->> (q/find-all-var-definitions db)
+                                 (remove #(uri-nses (:ns %)))
                                  (filter #(= cursor-name-str (str (:name %))))
                                  (map (fn [element]
                                         {:ns (str (:ns element))
```
ericdallo, line 596: We could receive a transudcer function in `find-all-var-definitions`, but for now, this is already a good quick improvement:
```suggestion
(into []
      (comp
        (remove #(uri-nses (:ns %)))
        (filter #(= cursor-name-str (str (:name %))))
        (map (fn [element]
               {:ns (str (:ns element))})))
      (q/find-all-var-definitions db))
```

lib/test/clojure_lsp/feature/code_actions_test.clj
```diff
@@ -7,7 +7,9 @@
    [clojure-lsp.test-helper :as h]
    [clojure.java.io :as io]
    [clojure.string :as string]
-   [clojure.test :refer [deftest is testing]]))
+   [clojure.test :refer [deftest is testing]]
+   [matcher-combinators.matchers :as m]
+   [matcher-combinators.test :refer [match?]]))
```
ericdallo, line 12: lint

Issue #1986, opened 2025-03-10
Navigate from declared var to definition


borkdude:
Given the following code, it would make sense if you could navigate from the declared var to the actual definition using find-definition:

``` clojure
(declare foo ;; navigate from here
)

(defn foo ;; to here
 [])

(foo)
```

PR #2025, opened 2025-04-18
Navigation from `declare` to actual definition

KarolAdamiec:
Enhance navigation by extending `find-definition` to work with declarations too. Check for existence of actual definition as a guard, because without it LSP returns `nil` and some editors have "cascades" for navigation handlers (i.e. Doom Emacs) and they try to find symbol using more naive search methods. #1986 

### DIFF
--- a/lib/src/clojure_lsp/queries.clj
+++ b/lib/src/clojure_lsp/queries.clj
@@ -320,13 +320,12 @@
 
 (defmethod find-definition :var-definitions
   [db {:keys [imported-ns defined-by name ns] :as var-definition}]
-  (let [xf-declared-definition (comp xf-analysis->var-definitions
-                                     (xf-same-fqn ns name)
-                                     (filter #(not= 'clojure.core/declare (:defined-by %)))
-                                     (xf-same-lang var-definition))
-        actual-definition (find-last-order-by-project-analysis
-                            xf-declared-definition
-                            (db-with-ns-analysis db ns))]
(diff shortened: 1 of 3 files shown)


lib/src/clojure_lsp/queries.clj
```diff
@@ -319,12 +319,28 @@
...
-                               :to imported-ns))
-    var-definition))
+  [db {:keys [imported-ns defined-by name ns] :as var-definition}]
+  (let [xf-declared-definition (comp xf-analysis->var-definitions
+                                     (xf-same-fqn ns name)
+                                     (filter #(not= 'clojure.core/declare (:defined-by %)))
+                                     (xf-same-lang var-definition))
+        actual-definition (find-last-order-by-project-analysis
+                            xf-declared-definition
+                            (db-with-ns-analysis db ns))]
```
ericdallo, line 329: with this change we are doing this find for every single usage of `find-definition` feature, which is the most used clojure-lsp feature, where we only really want to do when the definied-by is a declare, I'm wondering if we could wrap this with a `delay` and only use it [here](https://github.com/clojure-lsp/clojure-lsp/pull/2025/files#diff-953310fb5e54c1b32a0f931450ceb70ffc4e0130927391c85964242ebbc61c7cR339) after the definied-by check, WDYT?

ericdallo (approved):
Looks good! thank you!

PR #2042, opened 2025-05-07
Custom linters support

ericdallo:

### DIFF
--- a/docs/all-available-settings.edn
+++ b/docs/all-available-settings.edn
@@ -13,7 +13,8 @@
  :linters {:clj-kondo {:level :on
                        :report-duplicates true
                        :ns-exclude-regex ""}
-           :clj-depend {:level :info}} ;; Only if any clj-depend config is found
+           :clj-depend {:level :info} ;; Only if any clj-depend config is found
+           :custom {}}
  :clean {:automatically-after-ns-refactor true
          :ns-inner-blocks-indentation :next-line
          :ns-import-classes-indentation :next-line
--- a/lib/deps.edn
+++ b/lib/deps.edn
@@ -19,6 +19,7 @@
         org.benf/cfr {:mvn/version "0.152"}
         io.forward/semver {:mvn/version "0.1.0"}
         babashka/fs {:mvn/version "0.5.22"}
+        org.babashka/sci {:mvn/version "0.8.41"}
         com.github.clojure-lsp/lsp4clj {:mvn/version "1.13.1"
                                         #_#_:local/root "../../lsp4clj"
                                         :exclusions [funcool/promesa]}}
--- a/lib/src/clojure_lsp/feature/custom_linters.clj
+++ b/lib/src/clojure_lsp/feature/custom_linters.clj
@@ -1,6 +1,5 @@
 (ns clojure-lsp.feature.custom-linters
   (:require
-   [babashka.fs :as fs]
    [clojure-lsp.logger :as logger]
    [clojure-lsp.settings :as settings]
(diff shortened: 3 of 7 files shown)


docs/all-available-settings.edn
```diff
@@ -15,7 +15,8 @@
                        :ns-exclude-regex ""}
            :clj-depend {:level :info} ;; Only if any clj-depend config is found
            :clojure-lsp/unused-public-var {:level :info}
-           :cloujure-lsp/different-aliases {:level :off}}
+           :cloujure-lsp/different-aliases {:level :off}
```
acamargo, line 18:
```suggestion
           :clojure-lsp/different-aliases {:level :off}
```

lib/src/clojure_lsp/feature/diagnostics/custom.clj
```diff
@@ -0,0 +1,127 @@
...
+                                     'java.io.StringReader java.io.StringReader
+                                     'clojure.lang.LineNumberingPushbackReader clojure.lang.LineNumberingPushbackReader
+                                     ;; enable assert
+                                     'java.lang.AssertionError java.lang.AssertionError}
+                           :imports {'Exception 'java.io.Exception
+                                     'System java.lang.System
+                                     'File java.io.File}
+                           :load-fn (fn [{:keys [namespace]}]
+                                      (let [package (namespace-munge (name namespace))
+                                            path (str "clojure-lsp.exports/linters/" (string/replace package "." "/") ".clj")
```
acamargo, line 61: Is this going to work on windows?

ericdallo: Yes, as this paths comes from the jar which is universally separated by `/`, I tested that wondering the same thing

lib/src/clojure_lsp/feature/diagnostics/custom.clj
```diff
@@ -0,0 +1,127 @@
...
+
+(def ^:private required-fields #{:uri :range :severity :message :source :code})
+
+(defn ^:private missing-required-fields [diagnostic]
+  (seq (remove (set (keys diagnostic)) required-fields)))
+
+(defn ^:private custom-diagnostic->lsp [diagnostic]
+  (-> diagnostic
+      (dissoc :uri)
+      (update :severity #(case %
```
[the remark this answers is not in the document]
ericdallo: Not sure I got the advantage, for the end user, the ideal is to be keywords, especially for avoid breaking changes, and we already have this mapping around clojure-lsp.

Also, it's a number in LSP spec but because enums are numbers is spec, if you check the name of them is they same keyword name

acamargo (approved):

lib/src/clojure_lsp/feature/diagnostics/custom.clj
```diff
@@ -0,0 +1,129 @@
...
+
+(def ^:private required-fields #{:uri :range :level :message :source :code})
+
+(defn ^:private missing-required-fields [diagnostic]
+  (seq (remove (set (keys diagnostic)) required-fields)))
+
+(defn ^:private custom-diagnostic->lsp [{:keys [level range] :as diagnostic}]
+  (-> diagnostic
+      (dissoc :uri :level)
+      (assoc :range {:start {:line (dec (:row range))
```
acamargo, line 39: If a user forgets or mistypes the keywords `:row`, `:col`, `:end-row`, or `:end-col` in `:range`, these `dec`s will throw a NPE.

Not sure if we should provide a default value or validate them in `missing-required-fields`

PR #2059, opened 2025-05-22
Add support for rewrite nodes in custom linter

### LINKED ISSUE #2058: [custom-lint] Support analyze code of nodes defined by specific functions
We now support custom linters as per #2043, which adds a powerful support for clojure-lsp diagnostics, but clojure-lsp db/queries can be quite overwhelming and complex.

Discussing with @acamargo, it's common to want to lint specific function bodies of functions defined by a specific function, like clojure.core/def or schema.core/defn etc, we should offer a API function that helps with that, something like `(clojure-lsp.custom-linters-api/find-var-usages-defined-by #{:schema.core/defn clojure.core/defn} db)`.

We need to understand better if we should return analysis (var-usages or locals or everything?), nodes (rewrite-clj nodes).

### Example 

Example of a custom-linter with that need:

Say we have the following code:

```clojure
(defn foo []
  (str :bar "123")) ;; we want a linter: don't use :bar, use :baz
```

During custom-lint, we want to warn the user that one should not use `:bar` but `:baz`.

With kondo hooks, we could add a `:analyze-call` of `clojure.core/defn` which we would have the function body node and walk 
through the children looking for `:bar`.

With clojure-lsp custom-linter, we have the uri and db, so we could look for `:keywords` bucket in that uri, find it's related var-definition and register the linter, is there any helper that would help with that?

ericdallo:
Fixes #2058 


lib/src/clojure_lsp/custom_linters_api.clj
```diff
@@ -54,6 +56,13 @@
   [db ns-sym name-sym]
   (q/find-element-from-sym db ns-sym name-sym))
 
+(defn find-node-from-sym
+  "Find the rewrite-clj node from a fully qualified symbol."
```
acamargo, line 60: I think the function is returning a zipper instead of a node. They're different things.

ericdallo: The idea is return a zipper indeed, maybe improve the doc? I like the current name

ericdallo: Improved doc

Issue #1747, opened 2024-01-05
Add ability to safely analyze without executing project.clj


technomancy:
**Is your feature request related to a problem? Please describe.**
It should be possible to safely open an untrusted file without running code that's part of the project.

Imagine you are looking at a codebase, and you're not sure if you can trust the code. You open up one of its files in your editor to read it and see if it's safe or not. Since your editor has `clojure-lsp` configured, this starts up analysis. On the bright side, `clojure-lsp` does not execute any of the project code in the `src/` directory. However, it does run `lein classpath`, which executes `project.clj`. If the project is in fact unsafe, your machine has now been compromised.

**Describe the solution you'd like**
When `clojure-lsp` detects a Leiningen project, it should try running `lein static-classpath` first instead of `lein classpath` since the former is safe to run in any project. Since this task is relatively new, it should probably fall back to `lein classpath` if the former fails.

https://codeberg.org/leiningen/leiningen/commit/62175459ac3cde9faf0f812ec202ef25a8e9cb11

I'm open to suggestions for improvements to the implementation of `static-classpath`.

**Additional context**
It should be possible to configure it so it never performs the unsafe fallback. It should probably also be possible to configure it so it always prefers `lein classpath` since there are projects that need plugins in order to calculate the classpath.

However, I don't have a clear idea of how such configuration should be accomplished. If the configuration is read from the project repo itself, it would defeat the purpose, since an attacker could just set it to unsafe mode as part of the untrusted repository.

ericdallo:
Thanks for the report @technomancy, clojure-lsp has a general waterfall settings checking from project to home dir, but maybe we could check only the home one for this one somehow.
Also we don't have ATM a way to fallback for classpath lookup commands, we would need to support that, unless we think in a way the fallback is not needed.

PR #2114, opened 2025-09-01
Improve wording in Emacs config section

technomancy:
Improve the wording of the Emacs section of client config.

Clarify that lsp-mode is not required to use clojure-lsp with Emacs.

I don't know much about use-package or lsp-mode so I couldn't really
do much with the rest of the config section, but it could probably
still be clarified more.

In particular, if clojure-lsp is already supported by lsp-mode, why
would you need to add additional config yourself on top of that
existing support?

In addition, it seems the suggested config will automatically activate
clojure-lsp every time you open a clojure-mode file; this should not be
suggested without a warning given that clojure-lsp is not safe to use
on untrusted projects:

https://github.com/clojure-lsp/clojure-lsp/issues/1747

Probably such a warning belongs somewhere more general than just the
Emacs section, but the Emacs section should at least link to that
warning before suggesting any auto-activation.

### DIFF
--- a/docs/clients.md
+++ b/docs/clients.md
@@ -21,7 +21,10 @@ exec /usr/local/bin/clojure-lsp
 
 ## Emacs
 
-[lsp-mode](https://emacs-lsp.github.io/lsp-mode) has built in support for `clojure-lsp`. With `use-package`, add the following to your emacs config:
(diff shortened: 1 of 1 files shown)


ericdallo (approved):
thanks, this is old docs indeed, we need to revamp this section

Issue #2118, opened 2025-09-10
extract-method extracts only one sexpression


blueskyonmars:
**Is your feature request related to a problem? Please describe.**

I'm coming from Java/Intellij and I'm used to selecting multiple lines in a method to refactor.  I tried to do this with extract-function (in Calva) and only the first line was extracted.  The extract-function API only take an edit position (zloc) not a range, so this is understandable.  

**Describe the solution you'd like**

But I would have liked for all the expressions that I had selected to be included in the new function.

**Describe alternatives you've considered**

The best alternative is to copy and paste the additional lines.  Of course, sometimes this involves additional parameters, which might be annoying.

**Additional context**
There's a conversation here: https://clojurians.slack.com/archives/CPABC1H61/p1757517962830459 which Slack's AI summarizes as

[@JR](https://clojurians.slack.com/team/U02PB3ZMAHH) is having an issue with the extract-method functionality in Calva, a Clojure development environment. [@ericdallo](https://clojurians.slack.com/team/UKFSJSM38) explains that the current implementation only extracts a single expression, and suggests potential improvements to support extracting multiple expressions.

blueskyonmars:
Since I reported this, I'll take a look at how to address it.  I'm no Clojure or LSP expert, but I'll give it a shot.

PR #2170, opened 2025-12-09
added an extract function code action function with an option for selection

blueskyonmars:
This adds selection, potentially of multiple expressions, to the extract-function refactoring.  The existing refactoring based on the caret position should be unaffected and still work.  It puts the location at the end of the argument array so clients like Calva that modify the array don't need to change.   There is a minor change to the error handling of the existing functionality, in that an error is returned when there is nothing to extract instead of an empty refactoring.  (I believe that error case shouldn't happen outside of testing, at least via Calva.)

I didn't see anything to update for docs.

This addresses https://github.com/clojure-lsp/clojure-lsp/issues/2118

### DIFF
--- a/cli/deps.edn
+++ b/cli/deps.edn
@@ -24,7 +24,7 @@
            :build {:extra-paths ["../lib/resources"]
                    :extra-deps {babashka/fs {:mvn/version "0.5.30"}
                                 babashka/process {:mvn/version "0.6.25"}}
-                   :replace-deps {io.github.clojure/tools.build {:tag "v0.10.11" :sha "c6c670a"}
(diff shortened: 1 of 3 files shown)


ericdallo (commented):
Looks good! just small nitpicks

I need to test manually too

lib/test/clojure_lsp/refactor/transform_test.clj
```diff
@@ -532,17 +532,574 @@
...
+         zloc-end     (h/load-code-and-zloc code-end file-uri)]
+     (transform/extract-function start-row start-col
+                                 end-row end-col
+                                 zloc-start
+                                 zloc-end
                                  file-uri
                                  new-fn-name
                                  (h/db)))))
 
 (deftest extract-function-test
```
ericdallo: Nice tests!

Issue #2175, opened 2025-12-17
Extract-function from a thread-first/thread-last should include the threading parameter in the new signature


blueskyonmars:
I noticed that when I use extract-function from within a thread (-> or ->>), the new function doesn't include the threading variable in the new function's parameter list.  

It would be great if the new function includes not just the threading variable in the parameter list, but also includes the extracted body in the appropriate thread.  This is especially useful with a multi-expression extraction.

For now, I'm adding this manually.

This was discussed on the slack here, along with a video of a prototype.
https://clojurians.slack.com/archives/CPABC1H61/p1765998398276249

PR #2203, opened 2026-01-08
extract-function extracts threaded expressions correctly

blueskyonmars:
This addresses issue #2175, to include a way to extract threaded expressions into a new function.


ericdallo (approved):
It looks and works great!

Issue #2225, opened 2026-02-22
add performance smoke tests for Code Actions


blueskyonmars:
**Is your feature request related to a problem? Please describe.**

The issue here is that changes to Code Actions could slow down the editor experience.  It would be good to know when this happens via a test.

**Describe the solution you'd like**
To start, it would be good to have an end to end time measurement.

**Describe alternatives you've considered**
There are many options to measuring performance:
- measure end to end performance
- measure performance as a unit test that directly invokes the function calls
- measure times with different codebases
- measure time
- measure memory use
- etc

I looked into measuring the function calls directly, but there were two problems.  1) The database is put together by the unit test environment (eg h/db) and is quite small when compared with the size of even a relatively small program like clojure-lsp.  2) The function call times for these small tests are quite fast, which makes it hard to measure degradation precisely, particularly across machines.

Given this, I think using clojure-lsp as a target project and running clojure-lsp similar to the way the integration tests do would provide a real-world example.  The downside is that to debug any performance issues, one would need to use clj-async-profiler to find the offending method.

The downside of using clojure-lsp as a target project is that it can change, affecting the performance being measured.

So this would be more of a performance smoke test, but having a smoke test may well be good enough.

**Additional context**

This is somewhat related to https://github.com/clojure-lsp/clojure-lsp/issues/909, but is smaller in scope.

PR #2226, opened 2026-02-22
end to end performance tests

blueskyonmars:
This addresses issue #2225 - add performance smoke tests for Code Actions

### DIFF
--- a/cli/integration-test/performance/code_action_test.clj
+++ b/cli/integration-test/performance/code_action_test.clj
@@ -6,7 +6,6 @@
    [integration.lsp :as lsp]))
 
 ;; read files from the actual clojure-lsp codebase.  Note that this relies on the directory structure
-;; TODO: would it be better to use "git rev-parse --show-toplevel"?
 (def ^:private sample-file-name "../../../../../lib/src/clojure_lsp/feature/code_actions.clj")
 (def ^:private sample-file-with-many-lines "../../../../../lib/src/clojure_lsp/refactor/transform.clj")
 (def ^:private sample-file-with-many-symbols "../../../../../lib/src/clojure_lsp/queries.clj")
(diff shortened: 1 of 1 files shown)


cli/integration-test/performance/code_action_test.clj
```diff
@@ -0,0 +1,91 @@
...
+   [integration.fixture :as fixture]
+   [integration.lsp :as lsp]))
+
+;; read files from the actual clojure-lsp codebase.  Note that this relies on the directory structure
+;; TODO: would it be better to use "git rev-parse --show-toplevel"?
+(def ^:private sample-file-name "../../../../../lib/src/clojure_lsp/feature/code_actions.clj")
+(def ^:private sample-file-with-many-lines "../../../../../lib/src/clojure_lsp/refactor/transform.clj")
+(def ^:private sample-file-with-many-symbols "../../../../../lib/src/clojure_lsp/queries.clj")
+(def ^:private sample-file-with-high-fanout "../../../../../lib/src/clojure_lsp/handlers.clj")
+(def ^:private sample-file-with-large-structures "../../../../../lib/src/clojure_lsp/common_symbols.clj")
```
ericdallo: I think that's fine if we add performance tests to the CI, so one would notice it needs to update here.
Even so, I don't remember the last time I needed to change those ns names, probably many years

cli/integration-test/performance/code_action_test.clj
```diff
@@ -0,0 +1,91 @@
...
+  "find the mean runtime "
+  [body-fn]
+  (doall (for [_ (range execution-count)
+               :let [start-time (System/nanoTime)]]
+           (do
+             (body-fn)
+             (let [end-time (System/nanoTime)]
+               (nano->ms end-time start-time))))))
+
+;; TODO: also check P90 and median?
```
ericdallo: p90 would be nice in the future

Issue #1827, opened 2024-05-23
Add inline-function to commands


zikajk:
**Is your feature request related to a problem? Please describe.**
Occasionally, you may come across code that has a lot of nested functions and the naming is not good. 
In this case you may to use `xref / peek`, but it only allows you to jump one level. 
And If one jumps between a lot of namespaces, it is hard to remember everything. 

Intellij offers the possibility to replace a call to function by its definition, so even if a function has several nestings, it is possible to replace it in place and thus understand it more easily. 

**Describe the solution you'd like**
Being able to replace function call by its definition - similar to how `inline-function` works.

ericdallo:
@zikajk I tried to understand but not sure I got it, could you provide a code example like "when cursor is here and I call the code action, the code would be like that"

zikajk:
> @zikajk I tried to understand but not sure I got it, could you provide a code example like "when cursor is here and I call the code action, the code would be like that"

Sure. It is total opposite of "extract function". 
1. When cursor is on _(foo 1)_ and _foo_ is defined as _(defn foo [n] (println n))_ 
2.The code would be like _(println 1)_

ericdallo:
ah got it, sounds valid

PR #2240, opened 2026-03-09
implementation of inline-function

blueskyonmars:
This is an implementation of inline function.  It does come with some limitations mentioned in the comments.  #1827


ericdallo (approved):
💯

PR #2262, opened 2026-04-15
Preserve existing formatting in add-missing-libspec

### LINKED ISSUE #2261: "Add missing require" code action reformats the whole `:require` block
**Describe the bug**

When I trigger the "add missing require" code action on an `ns` form where the first required namespace sits on the same line as `:require`, the whole `:require` block gets re-indented instead of just the new entry being added. The diff ends up touching every line, which makes code review noisy for a one-line logical change.

**To Reproduce**

1. Start with an `ns` form like this, where `:require` keeps the first libspec on the same line and the rest are aligned under it:

   ```clojure
   (ns my.ns
     (:require [ns.alpha :as a]
               [ns.beta :as b]
               [ns.delta :as d]
               [ns.epsilon :as e]))
   ```

2. Somewhere in the file, use a symbol from a namespace that isn't required yet, e.g. `ns.gamma/foo`.
3. Trigger the code action to add the missing require.
4. `:ns-inner-blocks-indentation` and `:keep-require-at-start?` are not set in `.lsp/config.edn`, just the defaults.

What I see:

```clojure
(ns my.ns
  (:require
   [ns.alpha :as a]
   [ns.beta :as b]
   [ns.delta :as d]
   [ns.epsilon :as e]
   [ns.gamma :as g]))
```

Every existing line of the `:require` block is changed: `:require` now has a newline after it, and every libspec is indented under it with 2 spaces instead of aligned with the first one.

**Expected behavior**

Only the newly added entry should appear in the diff; the existing alignment should be preserved:

```clojure
(ns my.ns
  (:require [ns.alpha :as a]
            [ns.beta :as b]
            [ns.delta :as d]
            [ns.epsilon :as e]
            [ns.gamma :as g]))
```

**Screenshots**

N/A

<details>
<pre>
ADD JSON HERE
</pre>
</details>

<details>
<pre>
ADD HERE
</pre>
</details>

 - OS: <fill in>
 - Editor: Calva (VS Code)

**Additional context**

Git diffs for a single added require touch the whole block, hiding the actual change under formatting noise, and any team convention that keeps the first libspec on the same line as `:require` gets silently rewritten every time someone uses this code action.

Workaround: adding this to `.lsp/config.edn` avoids the reformat:

```clojure
{:clean {:ns-inner-blocks-indentation :keep}}
```

But I'd expect the default behavior to preserve whatever style the file already uses.

jramosg:

Introduce a new function to detect when the first child of :require/:import
is on the same line as the keyword. This allows the auto-clean feature to
preserve user-defined formatting styles instead of re-flowing the entire
block during namespace refactoring.

Fixes #2261


docs/settings.md
```diff
@@ -368,6 +368,8 @@ Whether to call clean-ns on the namespace after applying any refactor to it like
 
 How to indent ns children forms like require,import.
 
+When this setting is not set and `:keep-require-at-start?` is also unset, the automatic clean-up that runs after `add-missing-libspec` / `add-require-suggestion` / `add-missing-import` will behave as if `:keep` were configured when the ns is already in `:same-line` style, so the refactor only adds the new entry instead of re-flowing the whole `:require`/`:import` block. When no libspec is on the same line as `:require`/`:import`, the default remains `:next-line`.
```
ericdallo: `keep-require-at-start?` is deprecated for a long time in favor of `:ns-inner-blocks-indentation`

ericdallo:
Sorry for the delay @jramosg, I'm tyring to understand what are the best changes for clojure-lsp, commented in the issue [here](https://github.com/clojure-lsp/clojure-lsp/issues/2261#issuecomment-4304919205)

lib/src/clojure_lsp/feature/add_missing_libspec.clj
```diff
@@ -244,7 +244,7 @@
                                         (z/up)
                                         (cond->
                                          (or (not add-form-type?)
-                                             (fast= :next-line ns-inner-blocks-indentation)) (z/append-child* (n/newlines 1)))
+                                             (not (fast= :same-line ns-inner-blocks-indentation))) (z/append-child* (n/newlines 1)))
```
jramosg, line 247: With empty ns we cannot use `keep` so it creates multi-line block on empty `ns` without explicit setting.
This aligns with the clojure style guide, which says `:next-line` approach is better @ericdallo

jramosg:
@ericdallo the default behavior change to `:keep` is ready to review. I had to update some tests and I added 3 new ones to test that
- preserves existing :same-line formatting without explicit setting
- preserves existing :next-line formatting without explicit setting
- creates multi-line block on empty ns without explicit setting

Let me know your opinion and if you need anything else. Thanks!!

ericdallo (approved):
Looks good!

PR #2278, opened 2026-04-27
Support find-definition for fully qualified vars even when the namespace is not explicitly required

### LINKED ISSUE #2028: Support navigation of fqs not required
currently for:

```clojure
(ns some.foo)
(def my-var)
```

```clojure
(ns other.bar)

some.foo/my-var ;; find-definition won't work
```

clojure-lsp recognizes there is a missing require for `some.foo`, but won't allow find-definition to work unless that is required.

We could support the find-definition to work if the alias is the full namespace, like this case.

jramosg:

- Create `resolved-unknown-var-usage-namespace` to resolve unknown namespace usages for variables. When it is resolved, it adds `:to` keyword to `var-usage`
- Add unit tests:
    + namespace is resolved (alias is the full namespace)
    + namespace is not resolved
- Add integration tests for resolved namespace

Fixes #2028

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Support find-definition for fully qualified vars even when the namespace is not explicitly required. #2028
 - Fix `create-test` code action appending a duplicate `deftest` when one with the matching name already exists, now navigating to the existing deftest instead. #2274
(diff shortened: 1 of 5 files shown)


ericdallo (approved):
I wondered about the performance of find-definition, but I think it's good, it should only happen for unknown ns, thank you!

PR #2281, opened 2026-04-27
unit tests for command actions

blueskyonmars:
This addresses issue 2279 which complains that there isn't unit test coverage for clojure-lsp.feature.command/run-command.


ericdallo (approved):
Thanks! we can reuse this for other commands in the future or some automatic way to check that

PR #2282, opened 2026-04-30
Add Performance integration tests for server initialization

### LINKED ISSUE #2259: measure initialize (startup) time

It would be good to have a performance measurement of initialization, or startup, time.  This suggestion was from Eric in https://github.com/clojure-lsp/clojure-lsp/issues/2241:

>  Another one: startup (initialize), which calls kondo to analyze whole classpath, considering both cached (.lsp/.cache) and not cached

I think we can piggyback off #2241 and include the p50, p90, and p95 timings for clojure-lsp initialization.  This would use clojure-lsp itself as non-trivial target project.

I'm thinking that we could startup without the .lsp/.cache files first, and then try again with the cached files to measure the cached and uncached performance.  Not sure about how to remove the files, and if it's safe to do so, during performance testing.

Manual timing

**Additional context**

n/a

Jimmysit0:

Introduces the feature to measure and gate the initialization phase. Fixes [#2259](https://github.com/clojure-lsp/clojure-lsp/issues/2259). Also Integrated P50 latency assertions to ensure future changes do not regress startup performance. 

Performance on my machine: 
image

In order to run:
1) Build the native binary `bb native-cli`
2) `bb performance-test ./clojure-lsp`


ericdallo:
@Jimmysit0 I added performance-tests to CI and I'm seeing: 
```
FAIL in (initialization-performance) (/home/runner/work/clojure-lsp/clojure-lsp/cli/integration-test/performance/initialization_test.clj:)
measure Initialization (Warm Start)
expected: (< p50 max-p50)
  actual: (not (< 2858 2000))
```

Jimmysit0:
> @Jimmysit0 I added performance-tests to CI and I'm seeing:
> 
> ```
> FAIL in (initialization-performance) (/home/runner/work/clojure-lsp/clojure-lsp/cli/integration-test/performance/initialization_test.clj:)
> measure Initialization (Warm Start)
> expected: (< p50 max-p50)
>   actual: (not (< 2858 2000))
> ```

@ericdallo It seems that the latency assertions I set were too low. I tried to set them based on what I saw on my machine but seems like the edge wasn't enough. I modified `p50-cold-max-ms` and `p50-warm-max-ms`. It should work now on every machine as I increased the edge significantly but still catch poor startup performance.

ericdallo:
Thank you!!

Issue #1734, opened 2023-12-07
Unable to require suggested namespace when `ns` isn't present


raghavio:
**Describe the bug**
I'm trying to do a require via VSCode's quick fix suggestion box. It's failing with below error in output:
```
Request workspace/executeCommand failed.
  Message: Internal error
  Code: -32603 
[object Object]
```

**This is failing only when I don't have a namespace declartion in my file at the top using the `ns` macro.**

**To Reproduce**
1. Create a new clj file.
2. Write any function from another namespace. eg: `str/split-lines`
3. VSCode will show a pop up for quick fix to require the namespace.

**Expected behavior**
I expect it to add the require namespace using the `require` function but it's failing.

<details>
 <summary><b>Log - clojure-lsp</b></summary>
<pre>
2023-12-06T19:22:28.758Z  INFO [clojure-lsp.handlers:475] - :code-actions 87ms
2023-12-06T19:22:28.877Z  INFO [clojure-lsp.handlers:475] - :code-actions 3ms
2023-12-06T19:22:29.070Z  INFO [clojure-lsp.handlers:502] - :semantic-tokens-full 1ms
2023-12-06T19:22:29.739Z  INFO [clojure-lsp.handlers:385] - :clojuredocs-raw 0ms
2023-12-06T19:22:29.739Z  INFO [clojure-lsp.handlers:385] - :clojuredocs-raw 0ms
2023-12-06T19:22:29.937Z  INFO [clojure-lsp.handlers:475] - :code-actions 1ms
2023-12-06T19:22:30.099Z  INFO [clojure-lsp.handlers:302] - :document-highlight 0ms
2023-12-06T19:22:30.362Z  INFO [clojure-lsp.handlers:475] - :code-actions 1ms
2023-12-06T19:22:32.580Z  ERROR [clojure-lsp.server:55] - Error receiving message: Internal error (-32603)
{:id 6808, :method "workspace/executeCommand"}
com.oracle.svm.core.posix.thread.PosixPlatformThreads.pthreadStartRoutine  PosixPlatformThreads.java:  211
            com.oracle.svm.core.thread.PlatformThreads.threadStartRoutine       PlatformThreads.java:  832
                                                     java.lang.Thread.run                Thread.java: 1583
                                                 java.lang.Thread.runWith                Thread.java: 1596
                       java.util.concurrent.ThreadPoolExecutor$Worker.run    ThreadPoolExecutor.java:  642
[...]

PR #2286, opened 2026-05-06
while adding :require, add a ns if missing

blueskyonmars:
This addresses #1734, it adds a namespace form, if needed, when a :require (or :import) is added via a code action


ericdallo:
thank you!!

PR #2288, opened 2026-05-08
Add defaultLibrary semantic token modifier for clojure.core/cljs.core functions and macros

### LINKED ISSUE #2287: Add defaultLibrary semantic token modifier for clojure.core/cljs.core fns and macros
This adds support for the LSP `defaultLibrary` semantic token modifier on `clojure.core` and `cljs.core` functions and macros, allowing editors to highlight them in a distinct color — similar to how Cursive highlights core symbols in IntelliJ.

See screenshot:

Image

hjrnunes:
This adds support for the LSP defaultLibrary semantic token modifier on `clojure.core` and `cljs.core` functions and macros, allowing editors to highlight them in a distinct color, similar to how Cursive highlights core symbols in IntelliJ.

#### Config

The feature is opt-in via the `:semantic-tokens` `:default-library-for-core?` setting (defaults to `false`):
```clojure
;; .lsp/config.edn
{:semantic-tokens {:default-library-for-core? true}}
```
#### Editor theming

VS Code users can customize colors in settings.json:
```json
{
  "editor.semanticTokenColorCustomizations": {
    "rules": {
      "function.defaultLibrary": "#e06c75",
      "macro.defaultLibrary": "#c678dd"
    }
  }
}
```
#### Observations

- The defaultLibrary modifier is already declared by clojure-lsp (used for dynamic vars like *foo*) so we extend it to var usages that resolve to `clojure.core` or `cljs.core` via clj-kondo's `:to` field.
- The existing dynamic var behavior (`*foo*` → `:variable` with `:defaultLibrary`) should be unchanged.
- fixes gh-2287


ericdallo (approved):
Looks good! 
will give it a try soon

ericdallo:
@hjrnunes check how it looks in my emacs before and after:

imageimage

1. It would be nice to color macros vs functions differently, like before, I know this is a modifer, so maybe it's a client configuration, but would be nice to explore what we could do to make it work by default, maybe it's changing something on client side after this PR
2. I wonder if we could discuss how we avoid a big change for people visually, I mean, maybe the default color should be the same as before and we actually change non core functions to be another color, maybe a slightly different one from core ones.

PR #2294, opened 2026-05-27
Fix cyclic dependency check for as-alias

### LINKED ISSUE #2108: cyclic-dependencies is over-zealous #2 about :as-alias
**Describe the bug**
A ns that is required `:as-alias` should not contribute to a cyclic dependency since it does not load the ns (indeed, there doesn't even need to be a real ns behind such a require.

**To Reproduce**

1. Create two nses, `a` and `b`; have `a` require `b`; have `b` require `a` with `:as-alias`
2. This is flagged (incorrectly) as a cyclic dependency

**Expected behavior**
`:as-alias` should not contribute to a cyclic dependency.

 - Version: [2025.08.15-15.37.37](https://github.com/clojure-lsp/clojure-lsp/releases/tag/2025.08.15-15.37.37)

**Additional context**

jramosg:
Aliases created with :as-alias do not load the target namespace, so they should not count as dependency edges when detecting namespace cycles.

Add same-line and multiline regression coverage for :as-alias requires.

Fixes: #2108

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Added Performance integration tests for server initialization: Measuring Cold Start and Warm Start, ensuring that future changes don't regress the startup time of the LSP server.
+- Fix `cyclic-dependencies` linter falsely reporting cycles for `:as-alias` requires. #2108
 ## 2026.05.05-12.58.26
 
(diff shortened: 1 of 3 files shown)


ericdallo (approved):
Nice catch, thank you!

Issue #2326, opened 2026-06-18
Speed up startup diagnostics publishing on large projects


ericdallo:
Part of #2313.

On large projects most of `initialize` goes to publishing diagnostics, not analyzing them. Measured on a warm artemisia restart (3651 internal files, 54407 diagnostics):

| phase | time |
|---|---|
| clj-kondo analysis (warm, skipped) | 0ms |
| built-in linters (unused-public-var etc.) | 2.1s |
| custom linters | 3.6s |
| publish-all-diagnostics | 59.0s |
| `:lsp/initialize` total | 73.1s |

Publishing is 81% of startup. Building the diagnostics is cheap (~35ms); the cost is the serial notification drain. Every URI's `publishDiagnostics` goes through the diagnostics channel's 25ms per-`:uri` debounce. The debounce coalesces rapid edits to the same file on didChange, but at startup every URI is distinct and published once, so it adds roughly 16-25ms of pure latency per URI (3651 x ~16ms ~= 59s) while blocking `initialize`.

The debounced channel was introduced in #1153 to avoid overflowing the channel at startup, so a fix needs to keep that property while not paying the per-URI debounce on a one-shot bulk publish. Options to explore: route the startup bulk publish straight to the producer (or a non-debounced bulk path), and/or move the publish off the `initialize` critical path so the editor becomes responsive immediately.

As a first step this adds an `:internal/publish-all-diagnostics` timing marker so the cost is visible and measurable.

ericdallo:
Implemented in `2374e58f3db58187597f9241859138ac2b8339d8`.

Two changes: publish startup diagnostics directly to the producer (bypassing the per-uri debounce channel from #1153, which it can't overflow since there is no channel), and run the publish in a background thread so it no longer blocks `initialize`.

Measured on warm artemisia restarts (3651 internal files, 54407 diagnostics):

| stage | `:lsp/initialize` | publish |
|---|---|---|
| before | 73.1s | 59.0s (blocking) |
| direct publish | 40.6s | 32.3s (blocking) |
| direct + off critical path | 8.5s | 39.7s (background) |

Time-to-interactive dropped ~88% (73.1s to 8.5s); diagnostics stream in afterward. The remaining background cost is the per-notification floor (3651 distinct-uri `publishDiagnostics` plus editor processing), which is no longer on the critical path.

Issue #2316, opened 2026-06-12
Skip re-analysis of unchanged source paths on startup


ericdallo:
Part of #2313

Source-paths are fully re-analyzed with clj-kondo on every startup, even when the db cache is valid and almost nothing changed since the last session. External classpath analysis already skips unchanged paths via mtime checksums, source files don't get the same treatment.

On clojure-lsp repo itself, `:internal/project-paths-analyzed-by-clj-kondo` averages 5.4s of a ~13s warm initialize, the single biggest chunk.

Proposal:
- keep per-file checksums for source files too (`shared/generate-and-update-analysis-checksums` already does this for external batches)
- on startup, analyze only new/changed files and reuse the cached analysis for the rest
- invalidate as today when kondo config/settings/classpath hashes change

The tricky part is invalidation correctness: files changed while the server was down, branch switches, deleted files, etc., so this needs good test coverage around the mtime/hash detection.

While at it, two smaller warm startup items worth auditing (here or in follow-ups):
- `:internal/copy-kondo-configs` averages 1.1s, could be skipped when the classpath hash didn't change
- `:internal/refresh-clojuredocs-cache` averages 1.2s, should be fully async if it isn't already

Expected impact: warm initialize here from ~13s to ~8s. Scales with project source size, so monorepos with thousands of source files should save tens of seconds on every startup. External-only and cold startups unaffected.

Will post before/after warm initialize timings here.

ericdallo:
Two findings while measuring:

- the cache today only stores external analysis: it's written right after the external classpath scan, before source-paths analysis finishes (verified on 3 different project caches, none has internal uris or keyword/local buckets). So this issue also needs to start persisting source analysis, checksums alone aren't enough.
- on artemisia (huge private Nubank monorepo, 3,655 source files), `:internal/project-paths-analyzed-by-clj-kondo` took 18.2s of a 90s cold initialize, and it runs on every startup.

ericdallo:
Implemented in 95ea20a2fe7b44de70cb55a80233d2d1f21f6629

Source paths were fully re-analyzed with clj-kondo on every startup. Now the internal analysis is persisted in the db cache (with the dep-graph, documents, clj-kondo findings and a per-source-file mtime checksum), and a warm startup only re-runs kondo on the source files whose checksum changed, evicts files deleted while the server was down, and reuses the cached analysis for the rest. Needs a cache version bump (15 -> 16).

[...]

PR #2327, opened 2026-06-18
[debug] Diagnose CI integration test hang (#2326)

ericdallo:
Not for merge. Debugging vehicle to capture the CI-only integration test hang.

CI integration tests have been hanging since #2316 at `integration.definition-test/definition-external-dependency-jar-scheme` (opening external `clojure/core.clj` from the jar and hovering). It does not reproduce locally (debug jvm, prod jvm, or cold caches), and no in-test timeout fires, so the cause is invisible.

This adds:
- println markers in that test to pinpoint where it stalls (after definition / slurp / didOpen / hover)
- an `if: failure()` step on the jvm and graalvm integration jobs to upload the sample-test server log (`*.out`) as an artifact

Once CI runs, the markers + uploaded server log should reveal the actual failure, then this becomes the fix (or is closed in favor of a clean fix PR).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,6 +113,14 @@ jobs:
       - name: Run integration tests
         run: bb integration-test
 
+      - name: Upload integration server logs
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: integration-server-log-jvm-${{ matrix.os }}-${{ matrix.jdk }}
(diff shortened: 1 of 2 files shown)


ericdallo:
Replaced by #2329, which keeps only the useful server-log artifact upload and drops the debug markers. The java-interop failures this was meant to diagnose are fixed in 92b5012b (#2328).

PR #2411, opened 2026-07-21
Bump clj-kondo to 2026.05.26-20260720.212655-48 and Fix clj kondo warnings

jramosg:
- [] I created an issue to discuss the problem I am trying to solve or an open issue already exists.
- [] I added a new entry to [CHANGELOG.md](https://github.com/clojure-lsp/clojure-lsp/blob/master/CHANGELOG.md)

### DIFF
--- a/lib/src/clojure_lsp/feature/completion.clj
+++ b/lib/src/clojure_lsp/feature/completion.clj
@@ -245,7 +245,7 @@
     (cond-> {:label (element->label element cursor-alias priority)
              :priority (generic-priority->specific-priority element priority)}
       deprecated (assoc :tags [1])
-      kind (assoc :kind kind)
+      :always (assoc :kind kind)
       detail (assoc :detail detail)
       :always (completion-item-with-unresolved-documentation
                 {:name (-> element :name str)
--- a/lib/test/clojure_lsp/feature/diagnostics_test.clj
+++ b/lib/test/clojure_lsp/feature/diagnostics_test.clj
@@ -218,6 +218,7 @@
                            (foo 1 2))
                       (baz :p :q :r)
                       bar)
+                  #_{:clj-kondo/ignore [:constant-condition]}
                   (cond-> 0
                     int? (bar :a :b)
                     false (foo)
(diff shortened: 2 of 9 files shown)


lib/src/clojure_lsp/feature/completion.clj
```diff
@@ -245,7 +245,7 @@
     (cond-> {:label (element->label element cursor-alias priority)
              :priority (generic-priority->specific-priority element priority)}
       deprecated (assoc :tags [1])
-      kind (assoc :kind kind)
```
jramosg, line 248: `kind` comes from

```clojure
(defn ^:private element->completion-item-kind [{:keys [bucket arglist-strs macro]}]
  (cond
    (#{:namespace-definitions
       :namespace-usages} bucket)
    :module

    (#{:namespace-alias} bucket)
    :property

    (#{:keyword-usages :keyword-definitions} bucket)
    :keyword

    (and (#{:var-definitions} bucket)
         (or arglist-strs macro))
    :function

    (#{:var-definitions :locals} bucket)
    :variable

    (#{:var-usages} bucket)
    :reference

    (#{:local-usages} bucket)
    :value

    (#{:java-class-usages} bucket)
    :class

    :else
    :reference))
```

so it's always truthy

lib/test/clojure_lsp/feature/diagnostics_test.clj
```diff
@@ -235,7 +235,9 @@
                   "user/bar is called with 1 arg but expects 0"
                   "Expected: function, received: positive integer."
                   "user/bar is called with 1 arg but expects 0"
+                  "Condition always true"
```
jramosg, line 238:
```clojure
(cond-> 0
  int? (bar :a :b)
  false (foo)
  :else (baz 3))
````

`int?` and `:else` are always true, so I added the two new warnings. other alternatives:
- use `:clj-kondo/ignore`
- change them to `:always`

ericdallo:
I endup fixing in https://github.com/clojure-lsp/clojure-lsp/pull/2410 because Ihad other fixes there, but thank you for the input!

Issue #1810, opened 2024-04-22
Add "Replace full qualified ns with alias" feature


mrkam2:
An idea of an LSP/Calva editor action: when full namespace name is used in the code, introduce an alias and use it for all such places (or use an existing one). Example:
(ns a
  (:require [clojure.walk])

(clojure.walk/walk identity identity {:a 1})
(clojure.walk/walk identity identity {:b 2})
have an action to convert this to:
(ns a
  (:require [clojure.walk :as walk]))

(walk/walk identity identity {:a 1})
(walk/walk identity identity {:b 2})
cc @pez

Could also work if original code had (:require [clojure.walk :as walk]).

Slack discussion: https://clojurians.slack.com/archives/CPABC1H61/p1710203639508699.

U-C-S:
This would be nice to have. 

Will work on a implementation for this soon, though might need some help

blueskyonmars:
I have a prototype for this, if U-C-S isn't working on it, I can turn it into a PR.

PR #2417, opened 2026-07-22
Add code action to replace fully-qualified namespace with alias, and add :as to the :require form #1810

blueskyonmars:
This allows the user to convert a fully qualified namespace to an alias inline, potentially changing the libspec to add an :as.


blueskyonmars:
Included code-actions test in [d857d39](https://github.com/clojure-lsp/clojure-lsp/commit/d857d39fd3f9198e6df00d23269687126d1e2454)

ericdallo (changes requested):
Funcionally looks good, but I think we should use a different name for the command

`use-alias-suggestion` kinda couples the command with a previous diagnostic which is not the standard we do for other commands, I believe something like `swap-namespace-with-alias` or something would fit better, WDYT?

blueskyonmars:
Yes, "swap namespace with alias" describes this better.  I'll make the changes and re-record the .gif

Issue #2429, opened 2026-07-28
missing tests for inline-function and move-to-for-let code actions


blueskyonmars:
**Is your feature request related to a problem? Please describe.**

The inline-function and move-to-for-let actions are missing unit tests.

**Describe the solution you'd like**
Add unit tests for code-actions for these actions.

**Describe alternatives you've considered**
crossing fingers

**Additional context**
Found by code coverage tool

PR #2430, opened 2026-07-28
additional code coverage for inline-function and move-to-for-let code actions

blueskyonmars:
Just two additional code-actions tests that I missed earlier, for #2429


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Fix unreachable "Classpath not found" error branch on stubs generation.
 - Fix db cache write failure when a `#_{:clj-kondo/ignore [...]}` hint precedes java interop code, leaking non-serializable data into the analysis. #2380
 - Fix completion error in deps.edn/project.clj when the map is transiently unbalanced while typing a new key. #2384
+- Additional unit tests for code-actions for inline-function and move-to-for-let #2429
```
ericdallo: we don't really need to include changelog entries for tests or things that don't generate value for final user

blueskyonmars: Good point.  I'll take it out

PR #2458, opened 2026-08-27
Show built-in documentation when hovering Clojure special forms.

### LINKED ISSUE #2457: Hover on Clojure special forms does not show documentation
**Describe the bug**

`textDocument/hover` does not return documentation when hovering over the
Clojure special form `def`.

The response only contains the bare `def` symbol and the source location.
Hovering over `defn` returns its full documentation as expected.

This is reproducible through the raw LSP request and is not specific to Emacs.

This report concerns hovering over the `def` form itself, not hovering over
the user-defined var name.

**To Reproduce**

1. Open a Clojure file containing:

   ```clojure
   (ns example)
   (def foo 1)
   ```

2. Send `textDocument/hover` at line `1`, character `1` (zero-based), pointing
   at the `def` token.
3. Observe that the response contains only `def` and the file location.

For comparison, sending the same request for a `defn` token returns complete
signature and documentation text.

**Expected behavior**

The hover response should include the built-in documentation for the special
form, for example:

````markdown
```clojure
(def symbol doc-string? init?)
```

Special Form

Creates and interns a global var with the name of symbol in the current
namespace (*ns*) or locates such a var if it already exists. If init is
supplied, it is evaluated, and the root binding of the var is set to the
resulting value.

Please see http://clojure.org/special_forms#def
````

**Actual behavior**

The response is effectively:

````markdown
```clojure
def
```

----

*[/tmp/example.clj](file:///tmp/example.clj)*
````

<details>
<pre>
Request:

{
  "jsonrpc": "2.0",
  "id": 2,
  "method": "textDocument/hover",
  "params": {
    "textDocument": {
      "uri": "file:///tmp/example.clj"
    },
    "position": {
      "line": 1,
      "character": 1
    }
  }
}

Response:

{
  "jsonrpc": "2.0",
  "id": 2,
  "result": {
    "range": {
      "start": {
        "line": 1,
        "character": 1
      },
      "end": 1,
      "character": 4
    },
    "contents": {
      "kind": "markdown",
      "value": "```clojure\\ndef\\n```\\n\\n----\\n\\n*[/tmp/example.clj](file:///tmp/example.clj)*"
    }
  }
}
</pre>
</details>

<details>
<pre>
No relevant errors reported.
</pre>
</details>

- OS: Arch Linux
- Editor: Emacs 32.0.50
- Client: Eglot
- Version:

  ```text
  clojure-lsp 2026.07.06-14.34.19
  clj-kondo 2026.05.26-SNAPSHOT
  ```

**Additional context**

Clojure itself provides documentation for this form:

```clojure
(require '[clojure.repl :refer [doc]])
(doc def)
```

The issue appears to be that `def` is analyzed as a special form without a
namespace, while the existing hover ClojureDocs lookup only runs for symbols
with a namespace.

I have a proposed fix and can submit a pull request after the intended
special-form documentation source is confirmed.

0WD0:
Closes #2457.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Update development page with code coverage, performance, and debugging information #2431
 - Log more fine-grained timings for (mostly) initialization tasks to make solving user's performance issues easier #2433
 - Fix API/CLI rename corrupting files when a line contains multiple occurrences of the renamed symbol. #2450
+- Show built-in documentation when hovering Clojure special forms. #2457
 
 ## 2026.07.06-14.34.19
 
--- a/lib/src/clojure_lsp/feature/hover.clj
+++ b/lib/src/clojure_lsp/feature/hover.clj
@@ -2,6 +2,7 @@
   (:require
    [clojure-lsp.feature.clojuredocs :as f.clojuredocs]
    [clojure-lsp.feature.file-management :as f.file-management]
+   [clojure-lsp.feature.special-forms :as f.special-forms]
    [clojure-lsp.parser :as parser]
    [clojure-lsp.queries :as q]
    [clojure-lsp.refactor.edit :as edit]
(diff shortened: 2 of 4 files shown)


NoahTheDuke:
oh this is very nice

ericdallo (approved):
Clever, thanks!

